Issue #4, opened 2021-04-19
Integration tests rely on actual configuration variable


avnerenv0:
`examples/003_configuration_variable/main.tf` relies on the organization having 
- A configuration variable called `AWS_DEFAULT_REGION` in the organization scope.
- A configuration variable called `AWS_DEFAULT_REGION` in the project scope, in the `Default Organization Project` project.

This should _at least_ be documented but preferably fixed so the integration tests can run on a "clean" organization and not depend on prior manual setup, besides the API Key.

liranfarage89:
- The code in `002_project/main.tf` assumes that there is a default project with the name Default Organization Project. If there is no such project, the test fails.

```
│ Error: Could not find a project with name: Default Organization Project
│ 
│   with data.env0_project.default,
│   on main.tf line 1, in data "env0_project" "default":
│    1: data "env0_project" "default" {
│ 
╵

```

Issue #3, opened 2021-04-19
Integration tests fail to run on Mac/OSX


avnerenv0:
Getting an error

```
Provider terraform-registry.env0.com/env0/env0 v6.6.6 does not have a package
available for your current platform, darwin_amd64.
```

this can be fixed by changing `tests/harness.go` in `buildFakeTerraformRegistry` making the `registry_dir` end with `darwin_amd64` instead of `linux_amd64`, but we should have this resolved automatically.

PR #23, opened 2021-05-23
Chore: restructure packages

### LINKED ISSUE #10: Remove `.vscode` folder from the repo
We need to add gitignore to that folder.

yaronya:
1. Renamed packages and folders.
2. Splitted `client` package into `client` + `http` in favor of making the `client` layer testable.
3. Use `resty` package instead of native go http module.
4. Removed `.vscode` folder and added it to `.gitignore` (Fixes #10)


avnerenv0 (approved):
@yaronya it looks good to my novice eyes
I would just make sure that the current integration tests pass - they might be not much but they are all we have. Please notice https://github.com/env0/terraform-provider-env0/issues/3 and https://github.com/env0/terraform-provider-env0/issues/4

yaronya:
Integration tests were successful until failing on #24 (which doesn't seem like a regression introduced in this PR).
I've opened a task to fix this bug ASAP as a part of the current effort.

PR #25, opened 2021-05-24
Fix: organizationId is missing when creating projects

### LINKED ISSUE #13: Creating a new project returns an error
It seems like we are not sending the Organization id to the API, so we get an UnAuthorized error from the API.
We need to add it to the project resource. 

https://github.com/env0/terraform-provider-env0/blob/main/env0tfprovider/resource_project.go#L35

roni-frantchi:
Resolves #13

### DIFF
--- a/client/project.go
+++ b/client/project.go
@@ -26,7 +26,7 @@ func (self *ApiClient) ProjectCreate(name string) (Project, error) {
 	var result Project
 	organizationId, err := self.organizationId()
 	if err != nil {
-		return Project{}, nil
+		return Project{}, err
 	}
 
 	request := map[string]interface{}{"name": name, "organizationId": organizationId}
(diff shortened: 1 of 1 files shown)


client/project.go
```diff
@@ -24,8 +24,13 @@ func (self *ApiClient) Project(id string) (Project, error) {
 
 func (self *ApiClient) ProjectCreate(name string) (Project, error) {
 	var result Project
-	request := map[string]interface{}{"name": name}
-	err := self.http.Post("/projects", request, &result)
+	organizationId, err := self.organizationId()
+	if err != nil {
+		return Project{}, nil
```
avnerenv0, line 29: This shouldn't be - 
```suggestion
		return Project{}, err
```
?

roni-frantchi: Yes it should thanks!

avnerenv0 (approved):

roni-frantchi:
[TF Provider - Can't create project](https://trello.com/c/it7aD4Rz/3281-tf-provider-cant-create-project)

PR #27, opened 2021-05-24
Add github installation id

### LINKED ISSUE #19: Support github integrated templates
1. Add optional `github_installation_id` attribute to template resource + data schemas.
2. Send it on Create/Update operations.
3. Attach to object on Read. (both resource/data)

eranelbaz:
Resolves https://github.com/env0/terraform-provider-env0/issues/19

### DIFF
--- a/env0/data_template.go
+++ b/env0/data_template.go
@@ -85,7 +85,7 @@ func dataTemplate() *schema.Resource {
 			},
 			"github_installation_id": {
 				Type:        schema.TypeInt,
-				Description: "The installation id env0 application on the repository",
+				Description: "The env0 application installation id on the relevant github repository",
 				Optional:    true,
 			},
 		},
--- a/env0/data_template.go
+++ b/env0/data_template.go
@@ -83,6 +83,12 @@ func dataTemplate() *schema.Resource {
 				Description: "if specified, will only retry (on destroy) if error matches specified regex",
 				Computed:    true,
 			},
+			"github_installation_id": {
+				Type:        schema.TypeInt,
+				Description: "The installation id env0 application on the repository",
+				Computed:    true,
+				Optional:    true,
+			},
 		},
 	}
 }
@@ -138,6 +144,10 @@ func dataTemplateRead(ctx context.Context, d *schema.ResourceData, meta interfac
 	} else {
 		d.Set("retries_on_destroy", 0)
 		d.Set("retry_on_destroy_only_when_matches_regex", "")
+	}
+
+	if template.GithubInstallationId != 0 {
+		d.Set("github_installation_id", template.GithubInstallationId)
 	}
 
 	//TODO: sshkeys
--- a/env0/data_template.go
+++ b/env0/data_template.go
@@ -86,7 +86,6 @@ func dataTemplate() *schema.Resource {
 			"github_installation_id": {
 				Type:        schema.TypeInt,
 				Description: "The installation id env0 application on the repository",
-				Computed:    true,
 				Optional:    true,
 			},
 		},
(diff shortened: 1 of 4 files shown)


eranelbaz:
[TF Provider - GH installation id attribute for template resource and data types](https://trello.com/c/ar5QK83L/3279-tf-provider-gh-installation-id-attribute-for-template-resource-and-data-types)

env0/data_template.go
```diff
@@ -83,6 +83,12 @@ func dataTemplate() *schema.Resource {
 				Description: "if specified, will only retry (on destroy) if error matches specified regex",
 				Computed:    true,
 			},
+			"github_installation_id": {
+				Type:        schema.TypeInt,
+				Description: "The installation id env0 application on the repository",
+				Computed:    true,
```
yaronya, line 89: Why [`Computed`](https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#computed)?

env0/data_template.go
```diff
@@ -140,6 +146,10 @@ func dataTemplateRead(ctx context.Context, d *schema.ResourceData, meta interfac
 		d.Set("retry_on_destroy_only_when_matches_regex", "")
 	}
 
+	if template.GithubInstallationId != 0 {
```
yaronya, line 149: Shouldn't this be:
```suggestion
	if template.GithubInstallationId != nil {
```
?

env0/resource_template.go
```diff
@@ -88,6 +88,11 @@ func resourceTemplate() *schema.Resource {
 				Description: "if specified, will only retry (on destroy) if error matches specified regex",
 				Optional:    true,
 			},
+			"github_installation_id": {
+				Type:        schema.TypeInt,
+				Description: "description for the template",
```
yaronya, line 93: Please fix the description for this attribute

env0/data_template.go
```diff
@@ -83,6 +83,12 @@ func dataTemplate() *schema.Resource {
 				Description: "if specified, will only retry (on destroy) if error matches specified regex",
 				Computed:    true,
 			},
+			"github_installation_id": {
+				Type:        schema.TypeInt,
+				Description: "The installation id env0 application on the repository",
```
yaronya, line 88: Please fix description (also include the word github if possible)

yaronya (changes requested):
Nicely done, left some comments

env0/data_template.go
eranelbaz: can't use `nil` here since this is number

env0/data_template.go
eranelbaz: b8eb7e24e86f888217bb444ede4084e39046e515 you right

env0/resource_template.go
eranelbaz: b8eb7e24e86f888217bb444ede4084e39046e515

env0/data_template.go
eranelbaz: b8eb7e24e86f888217bb444ede4084e39046e515

env0/data_template.go
yaronya: @eranelbaz so that means default value in case template doesn't have GH installation id would be `0`? 🤔

eranelbaz: The default value of `int` in go is `0`,
so if we don't have `GithubInstallationId` on the tempalte its value should be `0` IIRC

avnerenv0: no that won't work

avnerenv0: At least we need to check it because if we create a template with `githubInstallation: 0` it passes validation but later on makes a lot of trouble for the org

yaronya (approved):

PR #28, opened 2021-05-24
Fix: ssh_key_names should be ssh_keys

### LINKED ISSUE #17: Assigning an SSH Key to a Template is not working
This is only implemented with `ssh_key_names` right now, but the template also needs the ID of the ssh key

roni-frantchi:
Resolves #17 

image


roni-frantchi:
[TF Provider - Can't add ssh key to template](https://trello.com/c/KeiRFqCX/3282-tf-provider-cant-add-ssh-key-to-template)

avnerenv0 (approved):
Nice!
Wonder if you should also add something about this in the 'docs' (README.md) or we can wait for that later?

roni-frantchi:
Updated readme @avnerenv0 https://github.com/env0/terraform-provider-env0/blob/fix-template-ssh-keys/README.md#resources

PR #29, opened 2021-05-24
Fix: Adjust Configuration api

### LINKED ISSUE #24: Can't create/update configuration variables
Create/Update configuration variables fails with `body is not an array`, as we aren't really sending the request as an array and API schema has changed recently.

avnerenv0:
Configuration API is now Array-only
Adjusted that

closes #24

### DIFF
--- a/client/configuration_variable.go
+++ b/client/configuration_variable.go
@@ -39,7 +39,7 @@ func (self *ApiClient) ConfigurationVariableCreate(name string, value string, is
 	if err != nil {
 		return ConfigurationVariable{}, err
 	}
-	var result ConfigurationVariable
+	var result []ConfigurationVariable
 	request := map[string]interface{}{
 		"name":           name,
 		"value":          value,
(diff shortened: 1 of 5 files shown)


avnerenv0:
[TF Provider - can't create/update configuration variables](https://trello.com/c/fvSSWpiE/3309-tf-provider-cant-create-update-configuration-variables)

eranelbaz (commented):
Question, i see you make a lot of `[0]` 
can't we multiple CRUD using terraform?

avnerenv0:
> Question, i see you make a lot of `[0]`
> can't we multiple CRUD using terraform?

We can do anything we are writing the code :)
But in this case it makes sense to have a "Configuration Property" resource that is singular, and it's also how TF providers are usually designed. The TF resources try to be 'independant' so we can't make one API call for multiple resource instances.

Our API is getting/giving arrays, so to play nice with a singular resource I had to `[0]`

roni-frantchi (approved):
> Our API is getting/giving arrays, so to play nice with a singular resource I had to [0]

Consider checking if >1 and log/warn/throw

avnerenv0:
> > Our API is getting/giving arrays, so to play nice with a singular resource I had to [0]
> 
> Consider checking if >1 and log/warn/throw

Since it's create/update and I'm also the one making the request (unlike find-by-name cases), I'm ok with leaving as is

Issue #16, opened 2021-04-21
Import should be implemented for all resource types


avnerenv0:
```
terraform import "env0_configuration_variable.test1" b94f954c-0b06-4fb4-8cb5-ed5161c99480
env0_configuration_variable.test1: Importing from ID "b94f954c-0b06-4fb4-8cb5-ed5161c99480"...

Error: Not implemented
```

:(

PR #43, opened 2021-05-25
Feat import configuration variable

eranelbaz:
PR 3/4 for https://github.com/env0/terraform-provider-env0/issues/16
Enable `terraform import env0_configuration_variable`

Note that after the import terraform run read in order to refresh the state, so in read we query for the variable again, but this time we need to set more details about the variable as the scope and the scope id,
for example if the scope in the input is `PROJECT` we need to change it to `project_id` and set the property value to be the scope id

Input example
```
terraform import env0_configuration_variable.tested1 '{  "Scope": "TEMPLATE",  "ScopeId": "e4025694-8370-4ab8-954a-af6cd4167d78",  "name": "fasdf"}'   
```

### DIFF
--- a/client/model.go
+++ b/client/model.go
@@ -38,6 +38,7 @@ type ConfigurationVariableSchema struct {
 }
 
 type ConfigurationVariable struct {
+	ScopeId        string                      `json:"scopId"`
 	Value          string                      `json:"value"`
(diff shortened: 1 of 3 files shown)


roni-frantchi (approved):
looking good

PR #55, opened 2021-05-27
Fix: harness test failures

### LINKED ISSUE #53: Integration tests fail - getConfigurationVariable - interface conversion
```
Stack trace from the terraform-provider-env0 plugin:

panic: interface conversion: interface {} is string, not *client.ApiClient

goroutine 30 [running]:
github.com/env0/terraform-provider-env0/env0.getConfigurationVariable(0x18b2c8d, 0x6, 0x0, 0x0, 0x0, 0x0, 0xc000120b80, 0x12, 0x0, 0x0, ...)
	/Users/avnersorek/env0/terraform-provider-env0/env0/data_configuration_variable.go:147 +0x8f3
github.com/env0/terraform-provider-env0/env0.dataConfigurationVariableRead(0x19ae020, 0xc000300240, 0xc000338600, 0x1882cc0, 0xc000683d80, 0xc0006b4760, 0xc00048f950, 0x100ff98)
	/Users/avnersorek/env0/terraform-provider-env0/env0/data_configuration_variable.go:103 +0x2af
```

yaronya:
Fixes #53 

### Solution
1. Fixed wrong parameter passed to `getConfigurationVariable`, `getSshKeyById`, `getSshKeyByName`.
2. Fixed wrong logic of `getTemplateByName` and `getSshKeyByName`.

### DIFF
--- a/env0/data_sshkey.go
+++ b/env0/data_sshkey.go
@@ -34,7 +34,7 @@ func dataSshKeyRead(ctx context.Context, d *schema.ResourceData, meta interface{
 	var sshKey client.SshKey
 	var err diag.Diagnostics
 	if nameSpecified {
-		sshKey, err = getSshKeyByName(meta, name)
+		sshKey, err = getSshKeyByName(name, meta)
 		if err != nil {
 			return err
 		}
--- a/env0/data_sshkey.go
+++ b/env0/data_sshkey.go
@@ -60,22 +60,25 @@ func getSshKeyByName(name interface{}, meta interface{}) (client.SshKey, diag.Di
 	apiClient := meta.(*client.ApiClient)
 
 	sshKeys, err := apiClient.SshKeys()
-	var sshKey client.SshKey
 	if err != nil {
 		return client.SshKey{}, diag.Errorf("Could not query ssh keys: %v", err)
 	}
-	if len(sshKeys) > 1 {
-		return client.SshKey{}, diag.Errorf("Found multiple SSH Keys for name: %s. Use ID instead or make sure SSH Keys names are unique %v", name, sshKeys)
-	}
+
+	var sshKeysByName []client.SshKey
 	for _, candidate := range sshKeys {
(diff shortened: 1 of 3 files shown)


env0/data_sshkey.go
```diff
@@ -60,22 +60,25 @@ func getSshKeyByName(name interface{}, meta interface{}) (client.SshKey, diag.Di
...
 	sshKeys, err := apiClient.SshKeys()
-	var sshKey client.SshKey
 	if err != nil {
 		return client.SshKey{}, diag.Errorf("Could not query ssh keys: %v", err)
 	}
-	if len(sshKeys) > 1 {
-		return client.SshKey{}, diag.Errorf("Found multiple SSH Keys for name: %s. Use ID instead or make sure SSH Keys names are unique %v", name, sshKeys)
-	}
+
+	var sshKeysByName []client.SshKey
```
avnerenv0, line 67: @yaronya you are the Go master - is there a normal way we could DRY this code ? 
Something like `getOnlyOneByName` ? 
We do the same thing for projects, ssh keys, templates, maybe more things ...

avnerenv0 (approved):
Looks good but if we can make this code better that might be better

env0/data_sshkey.go
yaronya: @avnerenv0 Wouldn't say master 😿 
I gave that an effort but found myself struggling with the idea of polymorphism in Go which exists only for `interface`s and not for `struct`s (in this case we need a "parent struct" with a `Name` member).

Since I really wanna finish the UTs today - I let that one go. Do you think it's worth an issue?

avnerenv0: <div><br/>via <a href="https://giphy.com/bfv/">Braxton Family Values </a> on <a href="https://giphy.com/gifs/bfv-sisters-braxton-family-values-we-tv-3o7TKNRz1oDplVigj6">GIPHY</a></div>

PR #59, opened 2021-05-30
Chore: Add Configuration Variables Client Unit Tests

avnerenv0:
resolves #29 

### Solution
Adds test for the Configuration Variables Client

### DIFF
--- a/client/configuration_variable_test.go
+++ b/client/configuration_variable_test.go
@@ -1,0 +1,68 @@
+package client_test
+
+import (
+	. "github.com/env0/terraform-provider-env0/client"
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Configuration Variable", func() {
+	mockConfigurationVariable := ConfigurationVariable{
+		Id:             "idX",
+		Name:           "configName",
+		Value:          "configValue",
+		OrganizationId: organizationId,
+		IsSensitive:    true,
+		Scope:          "PROJECT",
+		Type:           0,
+		ScopeId:        "project-123",
+		UserId:         "user|123",
+	}
+
+	Describe("ConfigurationVariableCreate", func() {
+		var createdConfigurationVariable ConfigurationVariable
+
+		BeforeEach(func() {
+			mockOrganizationIdCall(organizationId)
+
+			expectedCreateRequest := []map[string]interface{}{{
+				"name":           mockConfigurationVariable.Name,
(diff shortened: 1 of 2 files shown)


avnerenv0:
[UT for client/configuration_variable.go](https://trello.com/c/YsbXl01Z/3354-ut-for-client-configurationvariablego)

client/configuration_variable_test.go
```diff
@@ -0,0 +1,179 @@
...
+				ConfigurationVariableType(mockConfigurationVariable.Type),
+				nil,
+			)
+		})
+
+		It("Should get organization id", func() {
+			organizationIdCall.Times(1)
+		})
+
+		It("Should send POST request with params", func() {
```
eyalatzmon, line 61: it seems like that this It only check that httpCall has been called one time, maybe we should add assert for the expected params (`toHaveBeenCalledWith`). 
if we stay with httpCall.Times(1) maybe it will be more clear to remove "with params" from the It description

eyalatzmon: oh I missed the `mockHttpClient.EXPECT()` in line 40.
you can ignore the comment above

avnerenv0: yeah I don't like it either but that's how gomock works ...
@yaronya any way we could move the assertion from the `Before` to the `It`?

yaronya: That's how `gomock` works and how mocking in golang works in general, we can't compare it to JS 🤷 

The mock is created per function arguments so this has to be done **prior** to function call. For the sake of nicer BDD, I've used the `Times()` function to assert it on the `It` clause so in case it fails, the test output log would include it.

client/configuration_variable_test.go
```diff
@@ -0,0 +1,179 @@
...
+	Describe("ConfigurationVariableUpdate", func() {
+		var updatedConfigurationVariable ConfigurationVariable
+
+		BeforeEach(func() {
+			mockOrganizationIdCall(organizationId)
+
+			newName := "new-" + mockConfigurationVariable.Name
+			newValue := "new-" + mockConfigurationVariable.Value
+
+			expectedUpdateRequest := []map[string]interface{}{{
```
eyalatzmon, line 90: there are a common fields with `expectedCreateRequest` maybe we can dry by implement common expectedRequest

eyalatzmon (approved):
Looks good to me, left one comment

client/configuration_variable_test.go
avnerenv0: I couldn't find a nice way to copy/clone the map object in Go ...
So leaving as is for now ...

PR #60, opened 2021-05-30
Chore: sshkey client tests

razbensimon:
no unit tests for sshkey client

### Solution
add tests for sshkey client 
remove organization_id from public API. it is not necessary because of the organization cache.

### DIFF
--- a/client/model.go
+++ b/client/model.go
@@ -34,8 +34,8 @@ type Project struct {
 }
 
 type UpdateProjectPayload struct {
-	Name        string `json:"name,omitempty"`
-	Description string `json:"description,omitempty"`
+	Name           string `json:"name,omitempty"`
+	Description    string `json:"description,omitempty"`
 }
 
 type ConfigurationVariableSchema struct {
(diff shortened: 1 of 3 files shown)


razbensimon:
[UT for client/sshkey.go](https://trello.com/c/gwmZwGdC/3356-ut-for-client-sshkeygo)

client/model.go
```diff
@@ -149,7 +149,11 @@ type SshKey struct {
 }
 
 type SshKeyCreatePayload struct {
-	Name           string `json:"name"`
+	Name  string `json:"name"`
+	Value string `json:"value"`
+}
+
+type SshKeyCreatePayloadExtended struct {
+	SshKeyCreatePayload
```
roni-frantchi, line 157: Why the "extended"?  
Can we make it more similar to the other types?

razbensimon: I will be glad for a better name. 
We just need separated modes for Public API used by our users, and an Internal model for our internal REST api.

roni-frantchi: Why bother separating the two?  
How this entity different than say, Project?

razbensimon: We should do it for all the models contains organizationId. as discussed with @yaronya.
we want to expose fields that can be used by the user. it is weird to expose `organizationId` and do an unnecessary validation on our side - and throw an exception to the user. just do not expose it.

razbensimon: 45896c3

roni-frantchi (approved):

PR #62, opened 2021-05-31
Chore: provider unit tests (using ginkgo)

yaronya:
WIP

Blocked by https://github.com/onsi/ginkgo/issues/813

### Solution
WIP


yaronya:
Couldn't make this work out-of-the-box (used the workaround proposed [here](https://github.com/onsi/ginkgo/issues/813#issuecomment-851685310) but that didn't work). Seems like the terraform testing framework was not designed to run in a wrapping goroutine like used by `ginkgo` so this feels quite ROI negative ATM.

PR #63, opened 2021-05-31
Fix: Template creation model to omit empty github installation id

avnerenv0:
Creating a new template without specifying `github_installation_id` results in `githubInstallationId:0` which is not allowed.

### Solution
Make sure this property is also omitted when transforming to JSON.

### DIFF
--- a/client/model_test.go
+++ b/client/model_test.go
@@ -5,17 +5,25 @@ import (
 
 	. "github.com/env0/terraform-provider-env0/client"
 	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/types"
 )
 
 var _ = Describe("Models", func() {
 	Describe("TemplateCreatePayload", func() {
-		It("Should omit Github Installation Id when it's not there", func() {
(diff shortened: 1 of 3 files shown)


avnerenv0:
[Provider Integration tests creating a template with githubInstallationId=0](https://trello.com/c/aeClUs3D/3393-provider-integration-tests-creating-a-template-with-githubinstallationid0)

client/model_test.go
```diff
@@ -0,0 +1,21 @@
...
+	"encoding/json"
+
+	. "github.com/env0/terraform-provider-env0/client"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Models", func() {
+	Describe("TemplateCreatePayload", func() {
+		It("Should omit Github Installation Id when it's not there", func() {
```
eranelbaz, line 13: what about specify the id and check it exists in the payload

eranelbaz (commented):
small test case

client/model_test.go
avnerenv0: when you're right you're right
https://github.com/env0/terraform-provider-env0/pull/63/commits/4f8e41a00788f47f1ee22b5197e17666b8806019

roni-frantchi (approved):

PR #76, opened 2021-06-01
chore: add tests for orgamization client

razbensimon:

### Solution
add missing tests

### DIFF
--- a/client/organization_test.go
+++ b/client/organization_test.go
@@ -1,0 +1,67 @@
+package client_test
+
+import (
+	"errors"
+	. "github.com/env0/terraform-provider-env0/client"
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Organization", func() {
+	mockOrganization := Organization{
+		Id:   organizationId,
+		Name: "env0 🦄",
+	}
+
+	Describe("Organization", func() {
+		var organization Organization
+		var err error
+
+		Describe("Success", func() {
+			BeforeEach(func() {
+				organizationsResult := []Organization{mockOrganization}
+				httpCall = mockHttpClient.EXPECT().
+					Get("/organizations", nil, gomock.Any()).
+					Do(func(path string, request interface{}, response *[]Organization) {
+						*response = organizationsResult
+					})
+
+				organization, err = apiClient.Organization()
+			})
+
+			It("Should send GET request once", func() {
+				httpCall.Times(1)
+			})
+
+			It("Should return ssh keys", func() {
+				Expect(organization).Should(Equal(mockOrganization))
+				Expect(err).Should(BeNil())
+			})
+		})
+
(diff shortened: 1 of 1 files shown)


razbensimon:
[UT for client/organization.go](https://trello.com/c/GbZCo1sf/3355-ut-for-client-organizationgo)

client/organization_test.go
```diff
@@ -0,0 +1,67 @@
...
+					})
+
+				organization, err = apiClient.Organization()
+			})
+
+			It("Should send GET request once", func() {
+				httpCall.Times(1)
+			})
+
+			It("Should return ssh keys", func() {
```
eranelbaz, line 37:
```suggestion
			It("Should return organization", func() {
```

client/organization_test.go
```diff
@@ -0,0 +1,67 @@
...
+				expectedErr := errors.New("some error")
+				httpCall = mockHttpClient.EXPECT().
+					Get("/organizations", nil, gomock.Any()).
+					Return(expectedErr)
+
+				_, err = apiClient.Organization()
+				Expect(expectedErr).Should(Equal(err))
+			})
+
+			It("On error from server", func() {
```
eranelbaz, line 54: change on of the headers

client/organization_test.go
```diff
@@ -0,0 +1,67 @@
...
+				organizationsResult := []Organization{mockOrganization, mockOrganization}
+				httpCall = mockHttpClient.EXPECT().
+					Get("/organizations", nil, gomock.Any()).
+					Do(func(path string, request interface{}, response *[]Organization) {
+						*response = organizationsResult
+					})
+
+				_, err = apiClient.Organization()
+				Expect(err).ShouldNot(BeNil())
```
eranelbaz, line 63: maybe also assert the error?
because the error here is relevant so the user will understand what happened

eranelbaz (commented):
Couple of small comments
and what about the `organizationId()` function?

client/organization_test.go
razbensimon: opsss

client/organization_test.go
razbensimon: @eranelbaz didn't understand you

razbensimon:
> and what about the `organizationId()` function?

We should not test private methods. also, this logic already tested in other test files.
I also asked it in the office, and we discussed this point already 👍 .

client/organization_test.go
razbensimon: c3f746b

client/organization_test.go
razbensimon: c3f746b

client/organization_test.go
eranelbaz: you have twice `On error from server`
so change on of them so if the test will fail we will understand faster whats the problem

razbensimon: d18a4cf

eranelbaz (approved):
Looking really good

PR #94, opened 2021-06-02
chore: add tests for sshkey (resource + data)

razbensimon:

### Solution
- added tests for sshkey
-  fix quotes bug in helper tests


env0/resource_sshkey_test.go
```diff
@@ -0,0 +1,41 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: resourceConfigCreate(resourceType, resourceName, map[string]string{
+					"name":  sshKey.Name,
+					"value": sshKey.Value,
+				}),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceFullName, "id", sshKey.Id),
+					resource.TestCheckResourceAttr(resourceFullName, "name", sshKey.Name),
+					resource.TestCheckResourceAttr(resourceFullName, "value", sshKey.Value),
```
yaronya, line 30: @razbensimon So the decrypted ssh key exists on the returned object? 🤔

env0/data_sshkey_test.go
```diff
@@ -0,0 +1,67 @@
+package env0
+
+import (
+	"github.com/env0/terraform-provider-env0/client"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"testing"
+)
+
+func TestUnitSshKeyDataSourceById(t *testing.T) {
```
yaronya, line 9: It really feels like we can DRY those two test functions. I think it's worth an effort 🙏

env0/data_sshkey_test.go
```diff
@@ -0,0 +1,67 @@
...
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceFullName, "id", sshKey.Id),
+					resource.TestCheckResourceAttr(resourceFullName, "name", sshKey.Name),
+				),
+			},
+		},
+	}
+
+	runUnitTest(t, testCase, func(mock *client.MockApiClientInterface) {
+		mock.EXPECT().SshKeys().AnyTimes().Return([]client.SshKey{sshKey}, nil)
```
yaronya, line 35: Can we try to be more specific with `.Times(1)`?
This is also relevant for all other tests

env0/resource_sshkey_test.go
```diff
@@ -0,0 +1,41 @@
...
+	resourceName := "test"
+	resourceFullName := resourceAccessor(resourceType, resourceName)
+	sshKey := client.SshKey{
+		Id:    "id0",
+		Name:  "name0",
+		Value: "Key🔑",
+	}
+
+	testCase := resource.TestCase{
+		ProviderFactories: testUnitProviders,
```
yaronya, line 20: FYI `ProviderFactories` factories isn't needed here anymore because `runUnitTests` overrides it anyhow. Relevant for other tests in this PR as well. 
(I know you copied it from my test probably, I removed it in one of my recent merged PRs)

yaronya (commented):
Nice work 👍 
Left you some comments ✍️

env0/resource_sshkey_test.go
razbensimon: this is the `resource`,  not the `data` tests

env0/resource_sshkey_test.go
razbensimon: fa38ab5

env0/data_sshkey_test.go
razbensimon: 20f7ef7
worked only for resource tests file. data must have `AnyTimes` from some reason

yaronya: Can you tell maybe how many times it runs? (you can check out using debugger). 
I'm asking because it's important to understand this tho.If it's only two - we can add a setting that prevents the second time from running so we'd be able to assert the number of calls easily.

env0/resource_sshkey_test.go
yaronya: Yeah but still, I get that it should be a part of the `SshKeyCreatePayload` but I'm not sure it should be a part of the ssh key schema because nobody uses it and it's a secret value. WDYT?

env0/resource_sshkey_test.go
```diff
@@ -0,0 +1,40 @@
...
+					resource.TestCheckResourceAttr(resourceFullName, "id", sshKey.Id),
+					resource.TestCheckResourceAttr(resourceFullName, "name", sshKey.Name),
+					resource.TestCheckResourceAttr(resourceFullName, "value", sshKey.Value),
+				),
+			},
+		},
+	}
+
+	runUnitTest(t, testCase, func(mock *client.MockApiClientInterface) {
+		mock.EXPECT().SshKeyCreate(client.SshKeyCreatePayload{Name: sshKey.Name, Value: sshKey.Value}).Times(1).Return(sshKey, nil).Return(sshKey, nil)
```
yaronya, line 36: @razbensimon You got a redundant `Return()` call here 🙊

env0/data_sshkey_test.go
razbensimon: a1a0f56

env0/resource_sshkey_test.go
razbensimon: 414a3cd

env0/data_sshkey_test.go
razbensimon: where do you see it on the debugger? what line

yaronya: 🤙

env0/resource_sshkey_test.go
razbensimon: put sensitive for now, and lets talk with the team

razbensimon: 7887c41

env0/data_sshkey_test.go
razbensimon: a3bca40

yaronya (approved):

PR #103, opened 2021-06-03
Chore helpers tf with other then string

eranelbaz:
We don't have a way to create tf configuration other then string to string


RLRabinowitz (approved):
My PR also holds these changes, so nice one

PR #106, opened 2021-06-03
Fix: template path + description attributes removal

### LINKED ISSUE #85: Template Resource - removing "path" doesn't remove it
1. Create a template resource with `path = "shoes"`
2. Apply
3. Remove the `path` and apply again
4. The plan says `      - path                   = "shoes" -> null`
5. Nothing really changes

Expected behaviour would be to update `path` with an empty string when it's not defined

yaronya:
Fixes #85


RLRabinowitz (approved):
💪🏻

PR #108, opened 2021-06-04
Chore: go2hcl

razbensimon:
HCL generation needed more complex values support.

### Solution
- Move `test_helpers.go` logic to a new package called `go2hcl`
- Support int, bool, string, arrays and maps!
- Tests added for HCL generation

## Usage
```go
	hcl := ResourceConfigCreate("env0_x", "myresource", map[string]interface{}{
		"aaa": "test",
		"bbb": []int{0, 1, 2},
		"ccc": map[string]interface{}{
			"field1": 123,
			"field2": true,
			"field3": "hello",
		},
	})

```

## Output:
```hcl
resource "env0_x" "myresource" {
	aaa = "test"
	bbb = [0, 1, 2]
	ccc {
		field1 = 123
		field2 = true
		field3 = "hello"
	}
}
```

It was tiring!


yaronya:
@razbensimon That's a cool implementation but yet it could become a burden for us to maintain considering there are existing "official" solutions for generating HCL syntax such as hashicorp's [hclwrite](github.com/hashicorp/hcl2/hclwrite) package.

eranelbaz (commented):
why not leaving everyone use `test_helpers` and ut will use the `go2hcl`?
less refactor of existing files, no?

razbensimon:
@yaronya I understand you consern but from quick look on `hclwrite` it is not a magic solution. Assuming we rewrote it by that lib, my functions still be exists and just have different implementation - which we still need to "maintain". So no different.
We can try to do it  but in my opinion it is not in priority.
If no, we stays with the current helpers file, witch is the same as this pr but just without the arrays & object support. Which means we will have tests with different structures because their need of this support.

With those choices, let's discuss it more if needed. Generally I think we need to merge it because what I mentioned.

razbensimon:
[Support complex HCL generation](https://trello.com/c/yQzONKmD/3436-support-complex-hcl-generation)

razbensimon:
@yaronya @RLRabinowitz So?

@roni-frantchi @eranelbaz WDYT?

go2hcl/go2hcl.go
```diff
@@ -0,0 +1,115 @@
...
+	switch reflectedType.Kind() {
+	case reflect.String:
+		return fmt.Sprintf("\"%v\"", reflect.ValueOf(value).Interface()), nil
+	case reflect.Bool:
+		return fmt.Sprintf("%v", reflect.ValueOf(value).Interface()), nil
+	case reflect.Int, reflect.Int64, reflect.Int32, reflect.Int16, reflect.Int8,
+		reflect.Uint, reflect.Uint64, reflect.Uint32, reflect.Uint16, reflect.Uint8:
+		return fmt.Sprintf("%v", reflect.ValueOf(value).Interface()), nil
+	case reflect.Float32, reflect.Float64:
+		return fmt.Sprintf("%v", reflect.ValueOf(value).Interface()), nil
```
eranelbaz, line 83: same string same case?

eranelbaz (commented):
WDYT

razbensimon:
Closed after discussion, we don't want to maintain this code.

PR #109, opened 2021-06-05
Fix: template ssh keys read

### LINKED ISSUE #100: `resource "env0_template"` does not detect drifts in `ssh_keys`
1. Create `resource "env0_template"` without `ssh_keys` and apply it
2. Go to env0, attach an SSH key to the template, and save
3. run terraform plan, no changes are detected

yaronya:
We aren't setting template ssh keys upon reading, so there's not drift detection for this attribute.
Fixes #100 

### Solution
Set `ssh_keys` attribute on read + convert `[]TemplateSshKey` `[]map[string]string` 

@RLRabinowitz this would probably affect your tests PR #105

### DIFF
--- a/env0/resource_template.go
+++ b/env0/resource_template.go
@@ -199,6 +199,13 @@ func resourceTemplateRead(ctx context.Context, d *schema.ResourceData, meta inte
 	d.Set("revision", template.Revision)
 	d.Set("type", template.Type)
 	d.Set("terraform_version", template.TerraformVersion)
+
+	var rawSshKeys []map[string]string
+	for _, sshKey := range template.SshKeys {
+		rawSshKeys = append(rawSshKeys, map[string]string{"id": sshKey.Id, "name": sshKey.Name})
+	}
+	d.Set("ssh_keys", rawSshKeys)
+
 	if template.Retry.OnDeploy != nil {
 		d.Set("retries_on_deploy", template.Retry.OnDeploy.Times)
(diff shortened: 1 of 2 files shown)


env0/resource_template.go
```diff
@@ -199,6 +199,13 @@ func resourceTemplateRead(ctx context.Context, d *schema.ResourceData, meta inte
 	d.Set("revision", template.Revision)
 	d.Set("type", template.Type)
 	d.Set("terraform_version", template.TerraformVersion)
+
+	var rawSshKeys []map[string]string
+	for _, sshKey := range template.SshKeys {
+		rawSshKeys = append(rawSshKeys, map[string]string{"id": sshKey.Id, "name": sshKey.Name})
```
eranelbaz, line 205: why we need this loop?
looks like `template.SshKeys` already contains what we need no?

eranelbaz (commented):
small thing that looks off for me

env0/resource_template.go
yaronya: That's the whole fix basically.
Terraform schema expects a "raw" type (as it's defined with `TypeMap` for this attribute) so we must convert it from a custom type such as `TemplateSshKey` to the expected `map[string]string`

eranelbaz: so can't we change it from "raw" map to out TempalteSshKey?

yaronya: Nope, schema doesn't support non raw types.

eranelbaz (approved):
I hate it, but 🤷

RLRabinowitz:
> I hate it, but 🤷

That's Golang for you

PR #110, opened 2021-06-05
Fix: data project description attribute

### LINKED ISSUE #88: "description" shouldn't exist on the `data "env0_project"`
Currently, `description` is an optional field of `data "env0_project"`. If you create a `data` using this `description`, it is completely ignored.

We should either:
1. Remove it
2. Make it computed (and compute it on read)

yaronya:
Fixes #88


eranelbaz (approved):
easy peasy

PR #111, opened 2021-06-05
Chore: add User-Agent header

### LINKED ISSUE #22: API calls from the provider should come with a custom user-agent
something like `terraform-provider-env0`
if possible should include the provider version, tf version, maybe the architecture or something like that

yaronya:
☝️ 
Fixes #22 

### Solution
1. Add `User-Agent` request header to all API calls as `terraform-provider-env0` incl. SDK + provider versions (using terraform's `UserAgent` functionality)
2. Add some debugging ability (inspired by [this scaffolding repository](https://github.com/hashicorp/terraform-provider-scaffolding))
3. Make sure `go generate` also properly formats the terraform `examples` files

P.S - seems like testing request headers via `httpmock` isn't trivial, so 🤷

### DIFF
--- a/client/http/client.go
+++ b/client/http/client.go
@@ -23,19 +23,24 @@ type HttpClient struct {
 	client    *resty.Client
 }
 
-func NewHttpClient(apiKey string, apiSecret string, apiEndpoint string, restClient *resty.Client) (*HttpClient, error) {
+type HttpClientConfig struct {
+	ApiKey      string
+	ApiSecret   string
+	ApiEndpoint string
+	UserAgent   string
+	RestClient  *resty.Client
+}
+
+func NewHttpClient(config HttpClientConfig) (*HttpClient, error) {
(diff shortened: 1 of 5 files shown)


avnerenv0:
[Set provider user agent from go to env0 terraform provider](https://trello.com/c/9VaKKENJ/3402-set-provider-user-agent-from-go-to-env0-terraform-provider)

avnerenv0 (commented):
@yaronya there is an example of how they get the version in https://github.com/hashicorp/terraform-provider-scaffolding/blob/main/internal/provider/provider.go

Any chance we could also do it?
Would be very helpful as we're bound to get some alerts from outdated provider versions

eranelbaz:
@yaronya I agree with @avnerenv0, the version is relevant to understand if we get alerts from outdated providers or not

avnerenv0 (approved):

PR #117, opened 2021-06-12
Fix: template retry attributes

### LINKED ISSUE #83: Template Resource - Retry Regexes don't work
The retry regexes just do nothing

### LINKED ISSUE #84: Template Resource - removing retries_on_deploy/destroy doesn't disable retry
1. Apply a template with `retries_on_deploy`
2. Remove `retries_on_deploy`
3. Apply - the plan says `retries_on_deploy     = 4 -> null`
4. The retries don't actually change

yaronya:
Fixes #83 
Fixes #84

### DIFF
--- a/client/model.go
+++ b/client/model.go
@@ -76,12 +76,12 @@ const (
 
 type TemplateRetryOn struct {
 	Times      int    `json:"times,omitempty"`
-	ErrorRegex string `json:"errorRegex,omitempty"`
+	ErrorRegex string `json:"errorRegex"`
 }
 
 type TemplateRetry struct {
-	OnDeploy  TemplateRetryOn `json:"onDeploy,omitempty"`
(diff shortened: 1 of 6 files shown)


client/model.go
```diff
@@ -76,12 +76,12 @@ const (
 
 type TemplateRetryOn struct {
 	Times      int    `json:"times,omitempty"`
-	ErrorRegex string `json:"errorRegex,omitempty"`
+	ErrorRegex string `json:"errorRegex"`
```
eranelbaz, line 79: why remove the `omitempty`?
if the user didn't put any what we want to do?

eranelbaz (approved):
it looks good,
just why we remove the omit empty?

client/model.go
yaronya: We must remove `omitempty` to enable proper delete of this field

Issue #90, opened 2021-06-02
Add pre-commit hooks


razbensimon:
We need some run this as pre-commit git hook on this repo
1. `go fmt` (`husky` & `prettier` style)

Considering those too (maybe with a different hook? commit can be slow)
2. generate mocks
3. generate docs

yaronya:
I think 2 + 3 should be done in CI/Release rather than locally.
See also #57

yaronya:
Will be done in #262

PR #118, opened 2021-06-12
Run go fmt on CI

yaronya:
Fail on CI if `go fmt` has any output to ensure non-formatted code doesn't get merged.
Targets #90

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
       - name: Go fmt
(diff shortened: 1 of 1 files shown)


eranelbaz (dismissed):

eranelbaz:
@yaronya did you check that the current code base passing the `go fmt`?

eranelbaz:
@yaronya also looks like the `test` check not reporting...

yaronya:
@eranelbaz tested it, but had some specific fuckup with github actions. Fixed now.

eranelbaz (approved):

PR #121, opened 2021-06-25
Use env on shebang line in build.sh

ivankovnatsky:


avnerenv0 (approved):
Thanks @ivankovnatsky

PR #129, opened 2021-08-08
Add support for gitlab templates

### LINKED ISSUE #128: Add support for GitLab templates

eranelbaz:
resolves #128

Allow the user to create gitlab templates
### Solution
Add terraform attributes `gitlab_token_id` and `gitlab_project_id`
calculate `isGitlab` using those attributes


client/model.go
```diff
@@ -108,6 +108,7 @@ type TemplateCreatePayload struct {
 	TokenName            string           `json:"tokenName"`
 	TokenId              string           `json:"tokenId"`
```
razbensimon, line 109: add omit to TokenId also here

env0/data_template.go
```diff
@@ -79,6 +79,16 @@ func dataTemplate() *schema.Resource {
 				Description: "The env0 application installation id on the relevant github repository",
 				Optional:    true,
 			},
+			"token_id": {
+				Type:        schema.TypeString,
+				Description: "The env0 application token id on the relevant gitlab account",
```
razbensimon, line 84: the token_id is used also for Other vcs, so change the description accordingly

env0/data_template.go
```diff
@@ -131,6 +141,11 @@ func dataTemplateRead(ctx context.Context, d *schema.ResourceData, meta interfac
 		d.Set("github_installation_id", template.GithubInstallationId)
 	}
 
+	if template.TokenId != "" {
+		d.Set("token_id", template.TokenId)
+		d.Set("gitlab_project_id", template.GitlabProjectId)
+	}
```
razbensimon, line 147: here also, we dont need to set GitlabProjectId if it is just Other VCS.
so probably need to do something like this:

```go
if template.TokenId != "" {
	d.Set("token_id", template.TokenId)
        if template.GitlabProjectId != 0 {
        	d.Set("gitlab_project_id", template.GitlabProjectId)
        }
}
```

env0/resource_template.go
```diff
@@ -99,6 +99,16 @@ func resourceTemplate() *schema.Resource {
 				Description: "The env0 application installation id on the relevant github repository",
 				Optional:    true,
 			},
+			"token_id": {
+				Type:        schema.TypeString,
+				Description: "The env0 application token id on the relevant gitlab account",
```
razbensimon, line 104: here again

env0/resource_template.go
```diff
@@ -120,6 +130,19 @@ func templateCreatePayloadFromParameters(d *schema.ResourceData) (client.Templat
 	if githubInstallationId, ok := d.GetOk("github_installation_id"); ok {
 		result.GithubInstallationId = githubInstallationId.(int)
 	}
+	if tokenId, ok := d.GetOk("token_id"); ok {
+		result.TokenId = tokenId.(string)
+	}
+	if gitlabProjectId, ok := d.GetOk("gitlab_project_id"); ok {
+		result.GitlabProjectId = gitlabProjectId.(int)
+	}
+
```
razbensimon, line 139: does it work? i remember we do not save this on db

razbensimon (dismissed):
Nice work. just need to use this version and QA that it is working

env0/data_template.go
eranelbaz: WDYM?
Where we can use that for Other vcs?
can you suggest different description?

env0/resource_template.go
eranelbaz: Here again

env0/resource_template.go
eranelbaz: 5a4bc914

env0/data_template.go
eranelbaz: decc5707

client/model.go
eranelbaz: fc896435

eranelbaz:
7b8fc61a fixed CI

env0/resource_template.go
razbensimon: I didnt thought you need to delete it, just tell me if it works or not - or if we SHOULD make it work?

env0/data_template.go
razbensimon: why outside the first `if`?
we should not Set it if there is no tokenID first

env0/data_template.go
razbensimon: something like:
`the token id used for private git repos or for integration with GitLab. More info: https://docs.env0.com/docs/templates#manage-git-connectivity`

Your call

env0/resource_template.go
eranelbaz: I don't think we should
I saw on our FE we don't get it so i guess its not returned from BE

env0/data_template.go
eranelbaz: it doesn't really matter in my taste,
WDYT?

env0/data_template.go
eranelbaz: 1e8ebd59

env0/data_template.go
eranelbaz: 52220487
as we talked added to resource

razbensimon (approved):
nice!

PR #139, opened 2021-08-31
Add team resource

RLRabinowitz:
Feature Request: Support team resource

### Solution
Support team resource with two attributes: name and description.
Add tests

### DIFF
--- a/env0/resource_team.go
+++ b/env0/resource_team.go
@@ -1,0 +1,151 @@
+package env0
+
+import (
+	"context"
+	"errors"
+	"github.com/env0/terraform-provider-env0/client"
+	"github.com/google/uuid"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"log"
+)
+
+func resourceTeam() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: resourceTeamCreate,
+		ReadContext:   resourceTeamRead,
+		UpdateContext: resourceTeamUpdate,
+		DeleteContext: resourceTeamDelete,
+
+		Importer: &schema.ResourceImporter{StateContext: resourceTeamImport},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:        schema.TypeString,
+				Description: "The name of the team",
+				Required:    true,
+			},
+			"description": {
+				Type:        schema.TypeString,
+				Description: "Description for the team",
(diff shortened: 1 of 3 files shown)


env0/resource_team.go
```diff
@@ -0,0 +1,151 @@
...
+	err := apiClient.TeamDelete(d.Id())
+	if err != nil {
+		return diag.Errorf("could not delete team: %v", err)
+	}
+	return nil
+}
+
+func resourceTeamImport(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	id := d.Id()
+	var getErr diag.Diagnostics
```
eranelbaz, line 100: why you need this instead of returning `nil, err`?
i mean if `uuidErr` is nil then we don't need to make another `if` why not just return? same for the `else`

eranelbaz (commented):
what about tests for multiple teams with the same name?
api returns nothing for the id?
all the errors we should add `t.Run` for them

RLRabinowitz:
> what about tests for multiple teams with the same name?

You mean in import? We currently don't have tests for imports (doesn't seem like the current framework supports it)

> api returns nothing for the id?

This is something we never addressed in our code. We could throw an error in such a case. Not sure it's that important though

> all the errors we should add t.Run for them

The thing is that the tests aren't really unit tests. So testing those isn't nice (you need to mock exact amount of previous calls)

env0/resource_team.go
RLRabinowitz: I don't understand what you mean. 🤙🏻

eranelbaz:
@RLRabinowitz 

I mean you are throwing errors such as `could not update team:` so lets assert we get them when the relevant error accour,
those are stuff we do look at template resource, configuration variable resource cloud credentials project assignment and more

> The thing is that the tests aren't really unit tests. So testing those isn't nice (you need to mock exact amount of previous calls)

I pretty sure we do that, we have stuff like `gomock.InOrder` to handle those

RLRabinowitz:
@eranelbaz I know we do that sometimes. I'm just not sure it's worth the effort

RLRabinowitz:
Added tests for all but `delete`, because it's impossible with the flow (the error is outside of a step).
That just shows how bad this framework is for unit tests

eranelbaz (approved):
<div>

RLRabinowitz:
[TF provider - Support team resource](https://trello.com/c/MxQpNSEQ/4032-tf-provider-support-team-resource)

PR #140, opened 2021-08-31
Feat: Assign teams a project resource

liranfarage89:
### Issue & Steps to Reproduce / Feature Request
- a new resource: team-project-assignments
image

### Solution
- Use the rest CRUD API that was created for that: #[4781](https://github.com/env0/env0/pull/4781) #[4797](https://github.com/env0/env0/pull/4797)

PR #143, opened 2021-09-05
Fix: Validate project_role

liranfarage89:
While QAing #140 , @RLRabinowitz notice that on PR-Plan we don't validate the role.
### Solution
validate it.

### DIFF
--- a/env0/resource_team_project_assignment.go
+++ b/env0/resource_team_project_assignment.go
@@ -2,7 +2,8 @@ package env0
 
 import (
 	"context"
-	"github.com/env0/terraform-provider-env0/client"
+	"fmt"
+	. "github.com/env0/terraform-provider-env0/client"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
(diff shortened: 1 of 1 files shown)


env0/resource_team_project_assignment.go
```diff
@@ -32,7 +33,14 @@ func resourceTeamProjectAssignment() *schema.Resource {
 				Description: "the assigned role (Admin, Planner, Viewer, Deployer)",
 				Required:    true,
 				ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
-					_ = client.Role(val.(string))
+					role := Role(val.(string))
+					if role == "" ||
```
liranfarage89, line 37: empty validation is already validated in `required`,  omitting it..

avnerenv0 (approved):
TBH not loving this implementation,
`Role` should probably be a stronger type than `string` and also it's not the responsibility of `resource_team_project_assignment` to validate `Role` values, that should be in the model or similar.

However - not critical AT ALL so definitely approved 👍

liranfarage89:
@avnerenv0 
I've started something(very quickly), but the tests need to be changed, which is the most of work here.

how would you suggest handle that? opening a PR with "doing on the side"?

https://github.com/env0/terraform-provider-env0/tree/chore-move-role-validation (see the last commit)

avnerenv0:
@liranfarage89 

> opening a PR with "doing on the side"?
Nah, if you wanna do it just go ahead and get it over with, not "on the side".
Just let our dev channel know that you want to complete this part, and keep an eye on the ROI of the task, if it get's messy probably leave it.

BTW we talked about breaking that `model` file into file-per-model, you might as well put `Role` in it's own file.

liranfarage89:
that's ROI negative currently, that's why I thought about "doing on the side"  😕

PR #145, opened 2021-09-19
fix gitlab_project_id changes every apply

razbensimon:

`gitlab_project_id` always gets changed on TF apply.

image

### Solution
- remove the `Set` call because this field doesn't exist on the blueprint
- change readme for less mistakes with personal paths
- fix broken existing test + add test


razbensimon:
[Terraform Provider - GitLab Project ID](https://trello.com/c/Ng44pkjH/4165-terraform-provider-gitlab-project-id)

env0/resource_template.go
```diff
@@ -243,7 +243,6 @@ func resourceTemplateRead(ctx context.Context, d *schema.ResourceData, meta inte
 	d.Set("description", template.Description)
 	d.Set("github_installation_id", template.GithubInstallationId)
 	d.Set("token_id", template.TokenId)
-	d.Set("gitlab_project_id", template.GitlabProjectId)
```
yaronya, line 246: Can we cover this with some test?
It feels like we aren't asserting that the value is set after create/update.

razbensimon: 3fbecee

I am not sure if I did it correctly or if I missed something on mocks. please take a look @yaronya

razbensimon:
f9f9cd3 fixed broken test

eranelbaz:
@razbensimon looks like something is still broken here

env0/resource_template.go
razbensimon: fixed 7ae0d91

razbensimon:
@yaronya please rereview

yaronya (approved):

PR #151, opened 2021-10-11
Feat: Add project_policy resource and data source

### LINKED ISSUE #149: Add Project Policies support
We'd like to be able to configure all of env0's [Project Policies](https://docs.env0.com/docs/policies) using Terraform.

bernot-dev:
Project policy resource: Partially resolves #149 

### Solution
* Create resource for project policy
* Create data source for project policy


client/model.go
```diff
@@ -241,22 +241,36 @@ type TeamProjectAssignment struct {
...
+	SkipApplyWhenPlanIsEmpty   bool   `json:"skipApplyWhenPlanIsEmpty"`
+	DisableDestroyEnvironments bool   `json:"disableDestroyEnvironments"`
+	SkipRedundantDepolyments   bool   `json:"skipRedundantDeployments"`
 }
 
 type PolicyUpdatePayload struct {
 	ProjectId                  string `json:"projectId"`
-	NumberOfEnvironments       int    `json:"numberOfEnvironments"`
+	NumberOfEnvironments       int    `json:"numberOfEnvironments,omitempty"`
+	NumberOfEnvironmentsTotal  int    `json:"numberOfEnvironmentsTotal,omitempty"`
```
roni-frantchi, line 270: @bernot-dev I think that means we will not be able to disable the `NumberOfEnvironments` limit.  
In other words - if someone were to use the provider to set it, then would make a change in their terraform to remove it (`nil`) then it will _**not**_ be sent as part of the json payload, which mean it will not update its value at all.  

To update its value to `null` the backend expects one to explicitly tell it to `null` in the payload.  
Probably should have been a PATCH verb rather than PUT 🤷 

This is probably true for [these lines as well](https://github.com/env0/terraform-provider-env0/pull/151/files#diff-28a6f76d4db82dd4c73617ca57fa43c2eca0074f8b243cbe06aacb6311e2b168R258-R259)

Consider adding a test there to confirm the behavior I described

roni-frantchi (commented):
@bernot-dev thanks for the PR, looks good!  
I do think we have an issue with disabling the number of environment limitations

client/policy.go
```diff
@@ -18,6 +20,9 @@ func (self *ApiClient) Policy(projectId string) (Policy, error) {
 // PolicyUpdate updates a policy through the API
 func (self *ApiClient) PolicyUpdate(payload PolicyUpdatePayload) (Policy, error) {
 	var result Policy
+	if payload.ProjectId == "" {
+		return Policy{}, fmt.Errorf("projectId is required when updating policy")
+	}
```
yaronya, line 25: This should be validated by the resource schema, hence there's no use case where we'd be firing this API w/o project id

env0/data_policy.go
```diff
@@ -0,0 +1,107 @@
...
+	d.Set("id", policy.Id)
+	d.Set("project_id", policy.ProjectId)
+	d.Set("number_of_environments", policy.NumberOfEnvironments)
+	d.Set("number_of_environments_total", policy.NumberOfEnvironmentsTotal)
+	d.Set("requires_approval_default", policy.RequiresApprovalDefault)
+	d.Set("include_cost_estimation", policy.IncludeCostEstimation)
+	d.Set("skip_apply_when_plan_is_empty", policy.SkipApplyWhenPlanIsEmpty)
+	d.Set("disable_destroy_environments", policy.DisableDestroyEnvironments)
+	d.Set("skip_redundant_deployments", policy.SkipRedundantDepolyments)
+	d.Set("updated_by", policy.UpdatedBy)
```
yaronya, line 91: Aren't these redundant if you call `setPolicySchema` right after?

env0/data_policy.go
```diff
@@ -0,0 +1,107 @@
...
+
+	setPolicySchema(d, policy)
+	return nil
+}
+
+func getPolicyByProjectId(projectId string, meta interface{}) (client.Policy, diag.Diagnostics) {
+	apiClient := meta.(client.ApiClientInterface)
+	policy, err := apiClient.Policy(projectId)
+	if err != nil {
+		return client.Policy{}, diag.Errorf("Could not query template: %v", err)
```
yaronya, line 101: Looks like error message is wrong

env0/data_policy.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	if policy.Id == "" {
+		return client.Policy{}, diag.Errorf("Policy has empty id")
+	}
```
yaronya, line 105: How's this scenario is possible?
I mean - if policy doesn't exist then API would fail, so there is no scenario where API returns an invalid policy...

env0/data_policy_test.go
```diff
@@ -0,0 +1,48 @@
...
+		return resource.TestCase{
+			Steps: []resource.TestStep{
+				{
+					Config: dataSourceConfigCreate(resourceType, resourceName, map[string]interface{}{
+						"project_id": policy.ProjectId,
+					}),
+					Check: resource.ComposeAggregateTestCheckFunc(
+						resource.TestCheckResourceAttr(accessor, "id", policy.Id),
+						resource.TestCheckResourceAttr(accessor, "project_id", policy.ProjectId),
+						resource.TestCheckResourceAttr(accessor, "skip_apply_when_plan_is_empty", "true"),
```
yaronya, line 36: How about also validating that we set all other attributes properly?

env0/provider.go
```diff
@@ -38,6 +39,7 @@ func Provider(version string) plugin.ProviderFunc {
 			DataSourcesMap: map[string]*schema.Resource{
 				"env0_organization":           dataOrganization(),
 				"env0_project":                dataProject(),
+				"env0_policy":                 dataPolicy(),
```
yaronya, line 42: This is some sort of general comment about naming of the data/resource entities.
This entity should represent the policies defined under a project. We do have organization policies but it's not so robust nor represented yet by this TF provider.

Due to the above - I'd suggest to rename this from `env0_policy` to `env0_project_policy` (and also the relevant files...)

env0/resource_policy.go
```diff
@@ -0,0 +1,182 @@
...
+				Type:        schema.TypeString,
+				Description: "id  of the project",
+				Required:    true,
+				ValidateDiagFunc: func(i interface{}, path cty.Path) diag.Diagnostics {
+					projectId := i.(string)
+					if projectId == "" {
+						return diag.Errorf("Project ID cannot be empty")
+					}
+					return nil
+				},
```
yaronya, line 40: This is redundant. Having the `project_id` attribute schema to include the `Required` setting is enough...

yaronya (commented):
Nice job 🥇 
Left some comments.

env0/resource_project_policy.go
```diff
@@ -0,0 +1,174 @@
...
+			"number_of_environments": {
+				Type:        schema.TypeInt,
+				Description: "number of environments per project, 0 indicates no limit",
+				Optional:    true,
+			},
+			"number_of_environments_total": {
+				Type:        schema.TypeInt,
+				Description: "number of environments total, 0 indicates no limit",
+				Optional:    true,
+			},
```
yaronya, line 43: I think we should still be validating that these int are positive (and larger than `0`), otherwise API would fail so that's a bad UX

client/model.go
```diff
@@ -241,22 +243,56 @@ type TeamProjectAssignment struct {
...
+	NumberOfEnvironmentsTotal  int    `json:"numberOfEnvironmentsTotal"`
 	RequiresApprovalDefault    bool   `json:"requiresApprovalDefault"`
 	IncludeCostEstimation      bool   `json:"includeCostEstimation"`
 	SkipApplyWhenPlanIsEmpty   bool   `json:"skipApplyWhenPlanIsEmpty"`
 	DisableDestroyEnvironments bool   `json:"disableDestroyEnvironments"`
+	SkipRedundantDepolyments   bool   `json:"skipRedundantDeployments"`
+}
+
+func (p PolicyUpdatePayload) MarshalJSON() ([]byte, error) {
+	type serial struct {
```
yaronya, line 270: I tend to dislike this anti-pattern of arbitrary types. This is quite uncommon in golang.
Why can't we set`PolicyUpdatePayload` to simply support pointers instead?

env0/resource_project_policy_test.go
```diff
@@ -0,0 +1,94 @@
...
+	testCase := resource.TestCase{
+		Steps: []resource.TestStep{
+			{
+				Config: resourceConfigCreate(resourceType, resourceName, map[string]interface{}{
+					"project_id":             policy.ProjectId,
+					"number_of_environments": policy.NumberOfEnvironments,
+				}),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr(accessor, "project_id", policy.ProjectId),
+					resource.TestCheckResourceAttr(accessor, "number_of_environments", fmt.Sprintf("%d", policy.NumberOfEnvironments)),
```
yaronya, line 39: Can we please also validate all other fields? (relevant both for create & update)

tests/integration/011_policy/expected_outputs.json
```diff
@@ -0,0 +1 @@
+{}
```
yaronya, line 1: Do we require this file to exist? 🤔

tests/integration/011_policy/main.tf
```diff
@@ -0,0 +1,13 @@
...
+}
+
+resource "env0_project_policy" "test_policy" {
+  project_id                   = data.env0_project.default.id
+  number_of_environments_total = 1
+}
+
+resource "env0_project_policy" "test_policy_2" {
+  project_id                   = data.env0_project.default.id
+  number_of_environments_total = 0
```
yaronya, line 12: This should fail, did you try to run the harness test?

tests/integration/011_policy/main.tf
```diff
@@ -0,0 +1,13 @@
+data "env0_project" "default" {
+  name = "Default Organization Project"
+}
+
+resource "env0_project_policy" "test_policy" {
+  project_id                   = data.env0_project.default.id
+  number_of_environments_total = 1
```
yaronya, line 7: Adding some more relevant fields here would be great (mayne add them all?)

env0/resource_project_policy.go
```diff
@@ -0,0 +1,174 @@
...
+
+	_, err := apiClient.PolicyUpdate(payload)
+	if err != nil {
+		return diag.Errorf("could not delete policy: %v", err)
+	}
+
+	return nil
+}
+
+func resourcePolicyImport(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
```
yaronya, line 160: 1. One should be able to import a policy by the `project id` as well.
2. I think you're missing a `d.setId()` call here. IIRC otherwise terraform wouldn't save it to state. Correct me if I'm wrong please.

tests/integration/011_policy/expected_outputs.json
bernot-dev: It certainly appears so.

https://github.com/env0/terraform-provider-env0/blob/main/tests/harness.go#L115

env0/resource_project_policy.go
bernot-dev: As noted in the field description, 0 is being used as a sentinel value to indicate no limit. Added validation to handle negative numbers.

yaronya: Right, haven't seen it. What would happen if a user had set value `1` and then removed the attribute completely?

env0/resource_project_policy.go
bernot-dev: `id` is the `project_id`. I have updated the naming to reflect that this. That is the only way to look up a project policy, according to the API docs.

The `terraform.tfstate.backup` file saved after running this integration test indicates that the objects are being saved correctly. The `id` is set in the state, so it does not need to be set again.

bernot-dev: `id` is the `project_id`. I have updated the naming to reflect that this. That is the only way to look up a project policy, according to the API docs.

The `terraform.tfstate.backup` file saved after running this integration test indicates that the objects are being saved correctly. The `id` is set in the state, so it does not need to be set again.

client/model.go
bernot-dev: This would substantially complicate the code, because there would need to be `nil` checks everywhere these attributes are referenced. With 0 as an invalid value for these attributes that can be caught by the `ValidateDiagFunc`, I'm not sure what is gained by converting the `Policy` types to use pointers.

bernot-dev: This would substantially complicate the code, because there would need to be `nil` checks everywhere these attributes are referenced. With 0 as an invalid value for these attributes that can be caught by the `ValidateDiagFunc`, I'm not sure what is gained by converting the `Policy` types to use pointers.

yaronya (approved):

liranfarage89:
[TF provider - Project Policies](https://trello.com/c/0B6zuS90/4200-tf-provider-project-policies)

PR #154, opened 2021-10-20
Add description to configuration variable

roni-frantchi:
A configuration variable may have an optional short description.  
This PR adds it to both TF resource and data source so it may be set or consumed.


razbensimon (approved):
LGTM 👍🏼

PR #156, opened 2021-10-21
Chore: Add Sensitive on configuration variables

RLRabinowitz:
Currently, the value of configuration variables with `isSensitive: true` is displayed in the CLI output of terraform, on `terraform plan` and `terraform apply`

### Solution
Mark configuration variable values to always be sensitive


omry-hay (commented):
Can we add tests here?

omry-hay (approved):

PR #173, opened 2021-12-01
Feat trigger deployments for environment

alonnoga:
Trigger deployments when relevant fields change,
Left comments along the pr to clarify\emphasize important points


alonnoga:
[Terraform provider - Environment Resource Part 2](https://trello.com/c/K68z3UXy/4784-terraform-provider-environment-resource-part-2)

env0/resource_environment.go
```diff
@@ -23,27 +23,147 @@ func resourceEnvironment() *schema.Resource {
...
+	d.Set("template_id", environment.LatestDeploymentLog.BlueprintId)
+	d.Set("workspace", environment.WorkspaceName)
+	d.Set("revision", environment.LatestDeploymentLog.BlueprintRevision)
+	d.Set("repository", environment.LatestDeploymentLog.BlueprintRepository)
+	d.Set("run_plan_on_pull_requests", environment.PullRequestPlanDeployments)
+	d.Set("approve_plan_automatically", !environment.RequiresApproval)
+	d.Set("deploy_on_push", environment.ContinuousDeployment)
+	d.Set("deployment_id", environment.LatestDeploymentLogId)
+	d.Set("auto_deploy_by_custom_glob", environment.AutoDeployByCustomGlob)
+	//TODO: TTL and env\terraform variables
```
alonnoga, line 166: gonna do this on a separate pr

env0/resource_environment.go
```diff
@@ -78,13 +198,23 @@ func resourceEnvironmentRead(ctx context.Context, d *schema.ResourceData, meta i
...
+		deployPayload := getDeployPayload(d)
+		deployResponse, err := apiClient.EnvironmentDeploy(d.Id(), deployPayload)
+		if err != nil {
+			return diag.Errorf("failed deploying environment: %v", err)
+		}
+		d.Set("deployment_id", deployResponse.Id)
+	}
 
-	// TODO: deploy if needed
+	// TODO: update TTL if needed, also consider not updating ttl if deploy happened (cause we update ttl there too)
```
alonnoga, line 210: gonna do this on another pr

env0/resource_environment.go
```diff
@@ -23,27 +23,142 @@ func resourceEnvironment() *schema.Resource {
...
+			"deploy_on_push": {
+				Type:        schema.TypeBool,
+				Description: "should run terraform deploy on push events",
+				Optional:    true,
+				Default:     false,
+			},
+			"auto_deploy_by_custom_glob": {
+				Type: schema.TypeString,
+				// TODO: description
+				Description: "should deploy by custom glob",
```
alonnoga, line 74: not sure what this does, mind helping with the description?

env0/resource_environment_test.go
```diff
@@ -81,6 +88,118 @@ func TestUnitEnvironmentResource(t *testing.T) {
 		})
 	})
 
+	// TODO: test deploy with variables ( needs configuration changes read )
```
alonnoga, line 91: gonna do this in another pr (it requires the env\tf vars read)

env0/resource_environment.go
```diff
@@ -23,27 +23,142 @@ func resourceEnvironment() *schema.Resource {
...
+var VariableTypes = map[string]client.ConfigurationVariableType{
+	"terraform":   client.ConfigurationVariableTypeTerraform,
+	"environment": client.ConfigurationVariableTypeEnvironment,
+}
+
 func setEnvironmentSchema(d *schema.ResourceData, environment client.Environment) {
+	d.Set("id", environment.Id)
 	d.Set("name", environment.Name)
 	d.Set("project_id", environment.ProjectId)
-	d.Set("template_id", environment.TemplateId)
```
alonnoga, line 46: This is important,
When a deployment happens I want to set the deployment id from the response.
If I left this here it would get overridden to latestDeploymentLog.Id on the next read ( i.e when the user runs apply\plan\refresh etc).
Now create & update set deployment id separately,
and the deployment id of an environment wont be updated if triggered from elsewhere

if you think this is a problem please bring this up.

env0/resource_environment_test.go
```diff
@@ -81,6 +88,118 @@ func TestUnitEnvironmentResource(t *testing.T) {
...
+						schema_type = "%s"
+						schema_enum = ["%s"]
+					}
+				}`,
+			resourceType, resourceName, environment.Name,
+			updatedEnvironment.ProjectId, updatedEnvironment.LatestDeploymentLog.BlueprintId,
+			updatedEnvironment.LatestDeploymentLog.BlueprintRevision, configurationVariables.Name,
+			configurationVariables.Value, configurationVariables.Schema.Type,
+			strings.Join(configurationVariables.Schema.Enum, "\",\""),
+		)
```
alonnoga, line 143: I used sprintF instead of the resource build helper because
I started writing this test with config variables in mind
The builder doesn't support resources inside resources
I left it like this so its easier to add them in the. upcoming pr

resource with configuration changes example: 
```
resource "env0_environment" "test" {
  name        = "env name"
  project_id  = "ed7c54d2-ce49-4ad9-ab4e-be5f62107f03"
  template_id = "6a030803-4643-4ad0-a1fe-037feed14f21"
  configuration {
  	name = "var_name"
  	value = "a"
    schema_type = "string"
    schema_enum = ["a","b","c"]
  }
}
```

env0/resource_environment.go
```diff
@@ -23,27 +23,142 @@ func resourceEnvironment() *schema.Resource {
...
+				Optional:    true,
+				ForceNew:    true,
+				Computed:    true,
+			},
+			"revision": {
+				Type:        schema.TypeString,
+				Description: "the revision the environment is to be run against",
+				Optional:    true,
+				Computed:    true,
+			},
```
alonnoga, line 52: Please notice that these two are computed
I wanted those fields to override those values when specified.
Without computed, when they were left empty after a read would happen ( happens on plan\apply\refresh etc)
they were set to the terraform state (since the api returned values for them).
So, without actually changing the terraform config it would appear as if you changed them back to null.

Basically this is the only method I found to handle anything that may be overridden (and more importantly can be left empty) and is generated on the backend if it's not on the request.

*This may cause problems with drift detection

An example of what happened without computed: 

terraform apply this tf config
```
resource "env0_environment" "test" {
  name        = "env name"
  project_id  = "ed7c54d2-ce49-4ad9-ab4e-be5f62107f03"
  template_id = "6a030803-4643-4ad0-a1fe-037feed14f21"
}
```
terraform plan (without changing anything)
terraform would perform a read 
setting the resource to
```
resource "env0_environment" "test" {
  name        = "env name"
  project_id  = "ed7c54d2-ce49-4ad9-ab4e-be5f62107f03"
  template_id = "6a030803-4643-4ad0-a1fe-037feed14f21"
  workspace = "env0ae77fe"
}
```
then it would show that the plan would be
```
resource "env0_environment" "test" {
  name        = "env name"
  project_id  = "ed7c54d2-ce49-4ad9-ab4e-be5f62107f03"
  template_id = "6a030803-4643-4ad0-a1fe-037feed14f21"
  workspace = "env0ae77fe" -> null
}
```
because no workspace was specified in the tf config.

client/model.go
```diff
@@ -98,17 +98,17 @@ type ConfigurationVariableSchema struct {
...
-	Scope          Scope                       `json:"scope"`
-	Id             string                      `json:"id"`
-	Name           string                      `json:"name"`
-	Description    string                      `json:"description"`
-	Type           ConfigurationVariableType   `json:"type"`
-	Schema         ConfigurationVariableSchema `json:"schema"`
+	ScopeId        string                       `json:"scopeId,omitempty"`
+	Value          string                       `json:"value"`
+	OrganizationId string                       `json:"organizationId,omitempty"`
+	UserId         string                       `json:"userId,omitempty"`
```
eranelbaz, line 104: idk what i feel about those
we need to add `userId` and `orgId` and if for some reason we didn't add those i think we want to get runtime error
no?

client/model.go
```diff
@@ -229,11 +229,25 @@ type Template struct {
...
+	ProjectId                   string        `json:"projectId"`
+	WorkspaceName               string        `json:"workspaceName,omitempty"`
+	RequiresApproval            bool          `json:"requiresApproval,omitempty"`
+	ContinuousDeployment        bool          `json:"continuousDeployment,omitempty"`
+	PullRequestPlanDeployments  bool          `json:"pullRequestPlanDeployments,omitempty"`
+	AutoDeployOnPathChangesOnly bool          `json:"autoDeployOnPathChangesOnly,omitempty"`
+	AutoDeployByCustomGlob      string        `json:"autoDeployByCustomGlob,omitempty"`
+	Status                      string        `json:"status"`
+	LatestDeploymentLogId       string        `json:"latestDeploymentLogId"`
+	LatestDeploymentLog         DeploymentLog `json:"latestDeploymentLog"`
```
eranelbaz, line 243: this model is what about BE sending back?

env0/data_configuration_variable.go
```diff
@@ -127,12 +127,12 @@ func dataConfigurationVariableRead(ctx context.Context, d *schema.ResourceData,
 	d.Set("is_sensitive", variable.IsSensitive)
 	d.Set("scope", variable.Scope)
 	d.Set("enum", variable.Schema.Enum)
-	if variable.Type == client.ConfigurationVariableTypeEnvironment {
+	if *variable.Type == client.ConfigurationVariableTypeEnvironment {
```
eranelbaz, line 130: what this change means?

env0/resource_environment.go
```diff
@@ -23,27 +23,142 @@ func resourceEnvironment() *schema.Resource {
...
+				Type:        schema.TypeBool,
+				Description: "should run terraform plan on pull requests creations",
+				Optional:    true,
+				Default:     false,
+			},
+			"approve_plan_automatically": {
+				Type:        schema.TypeBool,
+				Description: "should deployments require manual approvals ( defaults to true )",
+				Optional:    true,
+				Default:     true,
```
eranelbaz, line 63: why those are the defaults?
i'd rather us to make another request to fetch the policies and choose from them

env0/resource_environment.go
eranelbaz: The glob file filter the env0 will check for changes

image

env0/resource_environment.go
```diff
@@ -23,27 +23,142 @@ func resourceEnvironment() *schema.Resource {
...
+			"deployment_id": {
+				Type:        schema.TypeString,
+				Description: "id of the last deployment",
+				Computed:    true,
+			},
+			"configuration": {
+				Type:        schema.TypeList,
+				Description: "terraform and environment variables for the environment",
+				Optional:    true,
+				Elem: &schema.Resource{
```
eranelbaz, line 86: Damn,
can't we reference the schema from `resource_configuration_variable`?

env0/resource_environment.go
```diff
@@ -156,6 +272,84 @@ func getUpdatePayload(d *schema.ResourceData) client.EnvironmentUpdate {
 	return payload
 }
 
+func getDeployPayload(d *schema.ResourceData) client.DeployRequest {
```
eranelbaz, line 275: where we are setting the deployed / creator user id?

env0/resource_environment.go
```diff
@@ -78,13 +194,23 @@ func resourceEnvironmentRead(ctx context.Context, d *schema.ResourceData, meta i
 func resourceEnvironmentUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	apiClient := meta.(client.ApiClientInterface)
 
-	payload := getUpdatePayload(d)
+	if d.HasChanges("template_id", "revision", "configuration") {
```
eranelbaz, line 197: why do we need this part?
We have `ForceNew`, this one will means we create another resource on the same tf id
idk if we want that

env0/resource_environment.go
```diff
@@ -156,6 +272,84 @@ func getUpdatePayload(d *schema.ResourceData) client.EnvironmentUpdate {
...
+	return configurationChanges
+}
+
+func getConfigurationVariableForEnvironment(variable map[string]interface{}) client.ConfigurationVariable {
+	varType := VariableTypes[variable["type"].(string)]
+
+	configurationVariable := client.ConfigurationVariable{
+		Name:  variable["name"].(string),
+		Value: variable["value"].(string),
+		Scope: client.Scope(variable["scope"].(string)),
```
eranelbaz, line 320: i'd rather us to use the same convention from the resource
where we infer the scope and the scope id from what the user has given us

eranelbaz (commented):
left you couple of comments
some of them are big 
if you want to talk about them LMK

client/model.go
alonnoga: they can be defined as required on the variable resource this is just the type for the client

env0/data_configuration_variable.go
alonnoga: this (and all the other chnages to the variable client\resource test files) was so that I can omit none mandatory fields from the request
omitEmpty omits nullish values not false values, so inorder for it to work I changed the omittable values to pointers (so its null for nothing and if not it will be included)

client/model.go
alonnoga: yes, Environment is the environment response (its the same for create and update)
when I had special responses I made a struct just for them (for example deployResponse)

env0/resource_environment.go
alonnoga: I just went with whats on the [ticket](https://trello.com/c/vCEIi9gh/3988-terraform-provider-environment-resource-part-1?filter=*)

env0/resource_environment.go
alonnoga: maybe, didn't see any example for it, and it will come with extra baggage (the crud funcs) but maybe, lest talk about it

env0/resource_environment.go
alonnoga: I didn't, maybe we should, lets talk about irl
(for the record, if the api doesn't force it I think its fine \ can be added later)

env0/resource_environment.go
alonnoga: there is no force new on any of these fields
this is the deploy condition

eranelbaz: you can't change environment template id this will have to force new
you can change the revision or configuration

env0/resource_environment.go
alonnoga: I don't understand
thats how the api expects it wdym infer it like the resource the variable resource?

env0/resource_environment.go
alonnoga: e997b6f

env0/resource_environment.go
alonnoga: 7dd986c

alonnoga: e485770

env0/resource_environment.go
```diff
@@ -23,27 +23,147 @@ func resourceEnvironment() *schema.Resource {
...
 				Type:        schema.TypeString,
 				Description: "project id of the environment",
 				Required:    true,
+				ForceNew:    true,
 			},
 			"template_id": {
 				Type:        schema.TypeString,
 				Description: "the template id the environment is to be created from",
 				Required:    true,
+				ForceNew:    true,
```
yaronya, line 39: noice

env0/resource_environment.go
yaronya: Thanks for the explanation 👑

env0/resource_environment.go
```diff
@@ -23,27 +23,147 @@ func resourceEnvironment() *schema.Resource {
...
+				Optional:    true,
+				Computed:    true,
+			},
+			"deploy_on_push": {
+				Type:        schema.TypeBool,
+				Description: "should run terraform deploy on push events",
+				Optional:    true,
+				Computed:    true,
+			},
+			"auto_deploy_on_path_changes_only": {
```
yaronya, line 72: Can this attribute and `auto_deploy_by_custom_glob` co-exist?

env0/resource_environment.go
yaronya: @alonnoga I would actually try to DRY it out.
You don't have to extract it including the CRUD methods, you could simply just set the schema to some variable and used it both here and on `configuration_variable` (it's a common thing in TF providers)

env0/resource_environment.go
```diff
@@ -23,27 +23,147 @@ func resourceEnvironment() *schema.Resource {
...
+							},
+						},
+					},
+				},
 			},
 		},
 	}
 }
 
+var VariableTypes = map[string]client.ConfigurationVariableType{
```
yaronya, line 141: I'm guessing this should be configured under `configuration_variable` rather than here, isn't it?

env0/resource_environment.go
yaronya: I think `LatestDeploymentLog` can be empty so I suggest to check if it exists first...

env0/resource_environment.go
```diff
@@ -78,13 +199,23 @@ func resourceEnvironmentRead(ctx context.Context, d *schema.ResourceData, meta i
 func resourceEnvironmentUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
yaronya, line 199: This is a gigantic function. Mind extracting its logic to some other functions? 🙏

env0/resource_environment.go
```diff
@@ -156,6 +296,85 @@ func getUpdatePayload(d *schema.ResourceData) client.EnvironmentUpdate {
...
+
+	if configuration, ok := d.GetOk("configuration"); ok {
+		configurationChanges := getConfigurationVariables(configuration.([]interface{}))
+		payload.ConfigurationChanges = &configurationChanges
+	}
+
+	if ttl, ok := d.GetOk("ttl"); ok {
+		payload.TTL = &client.TTL{
+			Type:  ttl.(map[string]interface{})["type"].(string),
+			Value: ttl.(map[string]interface{})["value"].(string),
```
yaronya, line 318: Would be cool if you can define a struct for the TTL schema...

yaronya (commented):
Nicely done! 💯 
Left some comments/questions

env0/resource_environment.go
alonnoga: deleted it here https://github.com/env0/terraform-provider-env0/pull/174 😄

env0/data_configuration_variable.go
```diff
@@ -127,12 +127,12 @@ func dataConfigurationVariableRead(ctx context.Context, d *schema.ResourceData,
...
 		d.Set("type", "environment")
-	} else if variable.Type == client.ConfigurationVariableTypeTerraform {
+	} else if *variable.Type == client.ConfigurationVariableTypeTerraform {
 		d.Set("type", "terraform")
 	} else {
-		return diag.Errorf("Unknown variable type: %d", int(variable.Type))
+		return diag.Errorf("Unknown variable type: %d", int(*variable.Type))
```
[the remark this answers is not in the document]
alonnoga: idk I just changed the type to a pointer

env0/resource_configuration_variable.go
```diff
@@ -172,12 +172,12 @@ func resourceConfigurationVariableRead(ctx context.Context, d *schema.ResourceDa
 			d.Set("description", variable.Description)
 			d.Set("value", variable.Value)
 			d.Set("is_sensitive", variable.IsSensitive)
-			if variable.Type == client.ConfigurationVariableTypeTerraform {
+			if variable.Type != nil && *variable.Type == client.ConfigurationVariableTypeTerraform {
```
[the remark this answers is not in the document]
alonnoga: trying to access null pointers will throw `invalid memory address or nil pointer dereference`

env0/resource_environment.go
alonnoga: tight

env0/resource_environment.go
alonnoga: I'll throw it model, it has consts similar to this there 👍🏻

env0/resource_environment.go
yaronya: So can we reflect it in schema validation somehow?

env0/resource_environment.go
alonnoga: There are several differences between the schemas.
stuff that are required when creating config variables but not when including them in deployments.
I could extract individual fields but with go it will create so much fucking code that I don't want to

env0/resource_environment.go
alonnoga: a99278f

env0/resource_environment.go
alonnoga: 9bdf19b

env0/resource_environment.go
alonnoga: c5609c4 sure thing

env0/resource_environment_test.go
alonnoga: ended up testing the vars so Im resolving this

env0/resource_environment.go
alonnoga: this is the only thing I found (and its not even documented on their site)
e47f7fb
it will throw this error when not both are specified (but it doesn't look like it force the field to be set to true so its still not 100%)
image

yaronya: Yeah nothing is documented on their site, they have nice godoc structure tho
I think it should be set the other way around also right?

alonnoga: thought so too but manual testing proved otherwise (its more like a dependency)

yaronya (approved):

PR #176, opened 2021-12-08
Fix environment resource variables on create

alonnoga:
For some reason when putting the variables in the deploy request doesn't take affect when creating an environment ( but it does work on deployment requests)
Fix: When creating an environment add configuration changes under the environment.


alonnoga:
[Environment Resource - Bug - When creating an environment with variables they don't apply](https://trello.com/c/LT6tBj82/4863-environment-resource-bug-when-creating-an-environment-with-variables-they-dont-apply)

HeverFarber (approved):
don't we have a test for it?

PR #177, opened 2021-12-12
Fix: Environment Update should not omit empty fields

roni-frantchi:
### Issue
1. Add an environment with  "auto_deploy_by_custom_glob"
(note that you also need "auto_deploy_on_path_changes_only" true for this one)
2. delete "auto_deploy_by_custom_glob" from tf config
3. tf apply again
4. tf show (will show that its gone), then tf refresh and it will be back cause it was never actually removed (the update request didn't delete it)

### Solution
Set the default values of all optional to `false`


roni-frantchi:
[Environment Resource bug - when removing fields they are not actually deleted](https://trello.com/c/ePmXK8PD/4859-environment-resource-bug-when-removing-fields-they-are-not-actually-deleted)

eranelbaz:
@roni-frantchi i see you also get 
`
test_reporter.go:43: Step 1/2 error: unsupported state format version: expected ["0.1" "0.2"], got "1.0"
`

fixing that as part of #178

roni-frantchi:
Yeah I've seen your PR but thanks for the heads up @eranelbaz

env0/resource_environment.go
```diff
@@ -311,25 +307,21 @@ func getCreatePayload(d *schema.ResourceData) client.EnvironmentCreate {
...
+	requiresApproval := !d.Get("approve_plan_automatically").(bool)
+	payload.RequiresApproval = &requiresApproval
+
+	pullRequestPlanDeployments := d.Get("run_plan_on_pull_requests").(bool)
+	payload.PullRequestPlanDeployments = &pullRequestPlanDeployments
+
+	autoDeployOnPathChangesOnly := d.Get("auto_deploy_on_path_changes_only").(bool)
+	payload.AutoDeployOnPathChangesOnly = &autoDeployOnPathChangesOnly
+
+	payload.AutoDeployByCustomGlob = d.Get("auto_deploy_by_custom_glob").(string)
```
alonnoga, line 323: doesn't this change always send them? (even if not specified)

env0/resource_environment_test.go
```diff
@@ -343,6 +343,105 @@ func TestUnitEnvironmentResource(t *testing.T) {
 			mock.EXPECT().EnvironmentDestroy(environment.Id).Times(1)
 		})
 	})
+
+	t.Run("Deleting triggers should set them to false", func(t *testing.T) {
+		falsey := false
+		truthyFruity := true
```
alonnoga, line 349: 😄

env0/resource_environment.go
```diff
@@ -57,25 +57,21 @@ func resourceEnvironment() *schema.Resource {
...
 				Type:        schema.TypeBool,
 				Description: "should run terraform deploy on push events",
 				Optional:    true,
-				Computed:    true,
 			},
 			"auto_deploy_on_path_changes_only": {
 				Type:        schema.TypeBool,
 				Description: "redeploy only on path changes only",
 				Optional:    true,
-				Computed:    true,
```
alonnoga, line 78: just so you know, if we set them (from the response value) theres a warning from terraform
(but we do that in other resources as well)

env0/resource_environment.go
roni-frantchi: Hmm. I'm not sure. Good catch though, think you're right - let me add a UT and see

env0/resource_environment.go
roni-frantchi: Gonna have to live with that I guess

env0/resource_environment.go
roni-frantchi: Actually there is a test there and it fails 😄   
I'll attend to it

roni-frantchi: https://github.com/env0/terraform-provider-env0/pull/177/commits/4dab816d97d3818110dbd5167e7771d6148d66fa

alonnoga (approved):
nice 👍🏻

PR #185, opened 2021-12-16
Fix: We don't import the deployment_id when importing an environment

RLRabinowitz:
We don't import `deployment_id` when importing an environment.
We should import the latest deployment log

### Solution
Import latest deployment log
No tests, because we don't have tests for imports + no integration tests yet


RLRabinowitz:
[TF Environment - import doesn't set deployment_id value](https://trello.com/c/aLW2raRY/4962-tf-environment-import-doesnt-set-deploymentid-value)

liranfarage89 (approved):

PR #188, opened 2021-12-19
Fix - fix support in any cnange deploy

GiliFaroEnv0:
fix support in any cnange deploy
### Solution

we needed to send the `auto_deploy_on_path_changes_only` as `false` any way so added default

### DIFF
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -77,12 +77,14 @@ func resourceEnvironment() *schema.Resource {
 				Type:        schema.TypeBool,
 				Description: "redeploy only on path changes only",
 				Optional:    true,
+				Default:     false,
 			},
 			"auto_deploy_by_custom_glob": {
 				Type:         schema.TypeString,
 				Description:  "redeploy on file filter pattern",
 				RequiredWith: []string{"auto_deploy_on_path_changes_only"},
 				Optional:     true,
+				Default:      "",
 			},
 			"deployment_id": {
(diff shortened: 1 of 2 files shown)


GiliFaroEnv0:
[Environment resource doesn't support "run on any change" trigger](https://trello.com/c/9rQtigkZ/4966-environment-resource-doesnt-support-run-on-any-change-trigger)

eranelbaz (approved):

env0/resource_environment.go
```diff
@@ -77,12 +77,14 @@ func resourceEnvironment() *schema.Resource {
 				Type:        schema.TypeBool,
 				Description: "redeploy only on path changes only",
 				Optional:    true,
+				Default:     false,
```
liranfarage89, line 80: it's supposed to be true, but if the user set that to false and no glob pattern, then it should be anyChange no?
@GiliFaroEnv0

eranelbaz: if we get to that we need to fetch the default from the BE

GiliFaroEnv0: @liranfarage89 since we want the default behavior to be "on any change" this need to be false

liranfarage89: on any change is not the default behavior as far as I know..

liranfarage89: check the ui and see

liranfarage89: @GiliFaroEnv0

PR #190, opened 2021-12-19
Fix: Add glob_pattern validation when cd/PrPlan enabled

liranfarage89:
Currently we don't validate that CD/PrPlan are enabled if the user set a custom_glob_pattern.  By default, the `runOnAnyChange` is enabled, so that should be true unless the user configured a glob pattern.
### Solution

1. Add validation in both create & update cases.
2. auth_deploy_on_path_changes_only must be `true` for the glob to take effect.

### DIFF
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -81,11 +81,10 @@ func resourceEnvironment() *schema.Resource {
 				Default:     true,
 			},
 			"auto_deploy_by_custom_glob": {
-				Type:         schema.TypeString,
-				Description:  "redeploy on file filter pattern",
-				RequiredWith: []string{"auto_deploy_on_path_changes_only"},
-				Optional:     true,
-				Default:      "",
+				Type:        schema.TypeString,
+				Description: "redeploy on file filter pattern",
+				Optional:    true,
+				Default:     "",
 			},
 			"deployment_id": {
 				Type:        schema.TypeString,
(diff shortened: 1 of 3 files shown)


liranfarage89:
[Environment update - continuous deployment bugs](https://trello.com/c/kUIxpkWF/4970-environment-update-continuous-deployment-bugs)

env0/resource_environment.go
```diff
@@ -81,11 +81,10 @@ func resourceEnvironment() *schema.Resource {
...
 			"auto_deploy_by_custom_glob": {
-				Type:         schema.TypeString,
-				Description:  "redeploy on file filter pattern",
-				RequiredWith: []string{"auto_deploy_on_path_changes_only"},
-				Optional:     true,
-				Default:      "",
+				Type:        schema.TypeString,
+				Description: "redeploy on file filter pattern",
+				Optional:    true,
+				Default:     "",
```
alonnoga, line 87: this could be redundant (I think* that d.get always returns the false value of that type)

alonnoga (approved):

env0/resource_environment.go
```diff
@@ -384,10 +396,23 @@ func getCreatePayload(d *schema.ResourceData, apiClient client.ApiClientInterfac
...
+		if (continuousDeployment == false) &&
+			(pullRequestPlanDeployments == false) {
+			return diag.Errorf("run_plan_on_pull_requests or deploy_on_push must be enabled for auto_deploy_by_custom_glob")
+		}
+		if autoDeployOnPathChangesOnly == false {
+			return diag.Errorf("cannot set auto_deploy_by_custom_glob when auto_deploy_on_path_changes_only is disabled")
+		}
+	}
+	return nil
+}
```
alonnoga, line 413: 👍🏻

env0/resource_environment.go
liranfarage89: f99f7e4

Issue #155, opened 2021-10-21
Deleting a resource manually corrupts the provider


RLRabinowitz:
1. Create the following resource and apply it:
```
resource "env0_template_project_assignment" "assignment" {
  template_id = "my_template_id"
  project_id = "my_project_id"
}
```
2. Manually using env0 UI, remove the template or simply un-assign it from the project
3. Run `terraform plan` again
4. Plan fails, `could not find projectId in template`

Reason:
On every Read, we fail if the resource already exists. However, we should only fail those if this is a new resource `d.IsNewResource`.
Otherwise, we need to `d.SetId('')`.

This bug applies to all resources

See [these lines from aws TF provider](https://github.com/hashicorp/terraform-provider-aws/blob/e86fca35f8f14709be928028a5550f29bbb21bc6/internal/service/s3/bucket.go#L830-L834) for an example.

eranelbaz:
Resolved by #202

PR #202, opened 2022-01-13
Fix - delete resource manually corrupts provider

razbensimon:
https://github.com/env0/terraform-provider-env0/issues/155
### Solution
- add `isDeleted` to template model
- "detect drift" and call `setId("")` for destroy the resource as suggested in TF [docs](https://learn.hashicorp.com/tutorials/terraform/provider-setup#implement-read)
    - when template **deleted** (we dont return NotFound in our api)
    - when template **detached** from project

- Extra: 
    - run `go mod tidy` to sync our `go.sum`
    - `harness.go`
        - return `error` and print it in order to understand when thing goes wrong
        - inject ENV0 variables in order to use them inside harness test


razbensimon:
[Fix Deleting a resource manually corrupts the provider](https://trello.com/c/cNLab9kc/5122-fix-deleting-a-resource-manually-corrupts-the-provider)

razbensimon:
Tests TL;DR - 
I tried to implement a new integration test in `harness` which:
- create a template, project, and its assignment
- delete the template by API request (create drift)
- run terraform again and see that my fix actually fix the crash and now we handle it

But that can't be achieved by `terraform` alone, because of Chiken&Egg kind of problem when need the `template_id` for delete by the API (tried HTTP call providers).
Anyways, we need a kind of `before/after` logic hooks solution.
So in terms of ROI, I leave it for now because it's too much work and obviously out of scope.

yaronya (approved):

eranelbaz:
Resolved https://github.com/env0/terraform-provider-env0/pull/202

Issue #183, opened 2021-12-14
Unclear how to "turn off" number_of_environments and number_of_environments_total options


away168:
For the resource "env0_project_policy" 
It is unclear how to mark "uncheck" like in the UI the following options:

      number_of_environments
      number_of_environments_total

mmclane:
I am seeing the same thing.

eranelbaz:
Hey, @away168 + @mmclane, we'll take a look at that,
my suggestion is If `number_of_environments` or `number_of_environments_total` set to `null` we'll uncheck those checkboxes
WDYT?

yaronya:
@eranelbaz I believe both removal and explicit `null` should disable the checkboxes. I believe currently only the former takes place but double check me

samuel-br:
>  I believe both removal and explicit `null` should disable the checkboxes. I believe currently only the former takes place but double check me

@yaronya 

for what  I checked when explicit null is disable the checkbox, I don't really understand what you meant about "removal", do you mean remove by enter '0' to the value? cause this is enforced  as error or to remove the fields from the TF file?

eranelbaz:
We confirmed when the values are null those checkboxes are uncheck

PR #208, opened 2022-01-18
Feat: environment scheduling resource

### LINKED ISSUE #194: environment scheduling
We should add support for environment scheduling as we have in our frontend under Environment -> Triggers ->  Scheduling

```
resource "env0_environment_scheduling" "name" {
	environment_id = "id" 
	deploy = "cron string"
	destroy = "cron string"
}
```

the deploy + destroy are optional but must at least one of them
need to validate they are valid cron string

data - 
```
data "env0_environment" "environment" {
  name   =  "name"
}

data "env0_environment_scheduling" "name" {
	environment_id = data.env0_environment.environment.id
}

output "deploy_scheduling" {
	value = data.env0_environment_scheduling.name.deploy
}


output "destroy_scheduling" {
	value = data.env0_environment_scheduling.name.destroy
}
```

yaronya:
Add `resource_environment_scheduling`
Resolves #194 

### Solution
☝️ 
Also changed the environment scheduling expression model to cope with empty values, hence turned it to a pointer


yaronya:
[Terraform provider - adding environment scheduling resource](https://trello.com/c/xRukN8AL/5225-terraform-provider-adding-environment-scheduling-resource)

env0/provider.go
```diff
@@ -59,6 +59,7 @@ func Provider(version string) plugin.ProviderFunc {
 				"env0_team_project_assignment":              resourceTeamProjectAssignment(),
 				"env0_team":                                 resourceTeam(),
 				"env0_environment":                          resourceEnvironment(),
+				"env0_environment_scheduling":               resourceEnvironmentScheduling(),
```
eranelbaz, line 62: what about data?

env0/resource_environment_scheduling.go
```diff
@@ -0,0 +1,128 @@
...
+		Schema: map[string]*schema.Schema{
+			"environment_id": {
+				Type:        schema.TypeString,
+				Description: "the environment's id",
+				Required:    true,
+				ForceNew:    true,
+			},
+			"destroy_cron": {
+				Type:             schema.TypeString,
+				Description:      "Cron expression for scheduled destroy of the environment",
```
eranelbaz, line 46: maybe add `should be different from deploy_cron`?

env0/resource_environment_scheduling.go
```diff
@@ -0,0 +1,128 @@
...
+	environmentScheduling, err := apiClient.EnvironmentScheduling(environmentId)
+
+	if err != nil {
+		return diag.Errorf("could not get environment scheduling: %v", err)
+	}
+
+	if environmentScheduling.Deploy != nil {
+		d.Set("deploy_cron", environmentScheduling.Deploy.Cron)
+	} else {
+		d.Set("deploy_cron", "")
```
eranelbaz, line 76: why you need to set that?
isn't that empty string by default? or why does it matter?

env0/resource_environment_scheduling.go
```diff
@@ -0,0 +1,128 @@
...
+
+	payload := EnvironmentScheduling{}
+
+	if deployCron != "" {
+		payload.Deploy = &EnvironmentSchedulingExpression{Cron: deployCron, Enabled: true}
+	}
+
+	if destroyCron != "" {
+		payload.Destroy = &EnvironmentSchedulingExpression{Cron: destroyCron, Enabled: true}
+	}
```
eranelbaz, line 104: i have a feeling this will create the same issue as https://github.com/env0/terraform-provider-env0/issues/183
can you make sure thats not happening here?

eranelbaz (commented):
code looking good, just have other issues IMO

please also attach the issue you are resolving
generate docs
add examples
`harness.go`?
data?

GiliFaroEnv0 (approved):
LGTM!
just look at my suggestion

GiliFaroEnv0 (dismissed):
look on @eranelbaz requests

env0/provider.go
yaronya: There is no real usage of having a data resource for environment scheduling. 
What would you use out of it? the cron expression?

env0/resource_environment_scheduling.go
yaronya: Since this isn't something we state in UI, didn't feel I need to add it here. It would immediately fail once a user enters the same value...

env0/resource_environment_scheduling.go
yaronya: That's how you handle drift detection. In case something removed one of the crons, the read operation needs to be update the state accordingly

env0/resource_environment_scheduling.go
yaronya: Cron's can't be `""` as it's not a valid cron expression, so `""` means default empty value. In your example, the value is number so either `0` or deleting it means removal (which unlike environment scheduling, is quite confusing)

env0/resource_environment_scheduling.go
eranelbaz: but the user will try to run `terraform apply` with that and we'll fail him on runtime 
i think that bad UX...

env0/provider.go
eranelbaz: 🤷 
we can leave that behind until someone will ask

env0/resource_environment_scheduling.go
yaronya: WDYM bad UX? That's the whole purpose of schema validations in terraform - it fails on plan time

eranelbaz: so idk if that will fail on plan since the validation happen on the `create` which is `apply`
why not just write that in the description?
then people will see "huhh i can't have the same cron here and there" when they write there tf

yaronya: It'd fail on plan, that's how schema validations work in TF provider. You can read it in their docs and also check it yourself

Since we don't explicitly state every validation or restriction we have in our docs, it neither felt to me as time to start doing it

yaronya: I thought we were talking about the cron expression validation for a sec. Anyways, I've added it in 2cbc05e but bear in mind that we don't do that usually (for any other validation that runs on CRUD)

eranelbaz (approved):

client/environment_scheduling.go
```diff
@@ -15,8 +15,10 @@ func (self *ApiClient) EnvironmentScheduling(environmentId string) (EnvironmentS
 func (self *ApiClient) EnvironmentSchedulingUpdate(environmentId string, payload EnvironmentScheduling) (EnvironmentScheduling, error) {
 	var result EnvironmentScheduling
 
-	if payload.Deploy.Cron == payload.Destroy.Cron {
-		return EnvironmentScheduling{}, errors.New("deploy and destroy cron expressions must not be the same")
+	if payload.Deploy != nil && payload.Destroy != nil {
+		if payload.Deploy != nil && payload.Deploy.Cron == payload.Destroy.Cron {
```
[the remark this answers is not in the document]
yaronya: Oops
e1de49f

GiliFaroEnv0 (approved):
GJ!

Issue #193, opened 2021-12-21
Workflow triggers support


eranelbaz:
We should add support for environment workflow triggers as we have in our frontend under Environment -> Triggers -> Workflow Triggers

```hcl
resource "env0_workflow_triggers" "name" {
	environment_id = "id"
	downstream_environment_ids = [
		"id1",
		"id2"
	]
}
```

Same for `data`

```hcl
data "env0_environment" "environment" {
  name   =  "name"
}

data "env0_workflow_triggers" "name" {
	environment_id = data.env0_environment.environment.id
}

output "downstream_envs" {
	value = data.env0_workflow_triggers.name.downstream_environment_ids
}
```

PR #211, opened 2022-01-20
downstream trigger api client

eranelbaz:
add api client for #193

### Solution
this pr is to add the api client
next pr will be the resource

moved all the `const` variables into the functions instead of global scope

### DIFF
--- a/client/api_client.go
+++ b/client/api_client.go
@@ -61,7 +61,7 @@ type ApiClientInterface interface {
 	EnvironmentSchedulingUpdate(environmentId string, payload EnvironmentScheduling) (EnvironmentScheduling, error)
 	EnvironmentSchedulingDelete(environmentId string) error
 	WorkflowTrigger(environmentId string) ([]WorkflowTrigger, error)
-	WorkflowTriggerCreate(environmentId string, request WorkflowTriggerCreatePayload) (WorkflowTrigger, error)
+	WorkflowTriggerUpsert(environmentId string, request WorkflowTriggerUpsertPayload) ([]WorkflowTrigger, error)
 }
 
 func NewApiClient(client http.HttpClientInterface) ApiClientInterface {
--- a/client/api_client.go
+++ b/client/api_client.go
@@ -60,6 +60,8 @@ type ApiClientInterface interface {
 	EnvironmentScheduling(environmentId string) (EnvironmentScheduling, error)
 	EnvironmentSchedulingUpdate(environmentId string, payload EnvironmentScheduling) (EnvironmentScheduling, error)
 	EnvironmentSchedulingDelete(environmentId string) error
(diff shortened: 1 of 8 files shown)


client/api_client.go
```diff
@@ -60,6 +60,8 @@ type ApiClientInterface interface {
 	EnvironmentScheduling(environmentId string) (EnvironmentScheduling, error)
 	EnvironmentSchedulingUpdate(environmentId string, payload EnvironmentScheduling) (EnvironmentScheduling, error)
 	EnvironmentSchedulingDelete(environmentId string) error
+	WorkflowTrigger(environmentId string) ([]WorkflowTrigger, error)
+	WorkflowTriggerCreate(environmentId string, request WorkflowTriggerCreatePayload) (WorkflowTrigger, error)
```
RLRabinowitz, line 64: It's Update, not Create
Or Upsert

client/aws_credentials_test.go
```diff
@@ -7,9 +7,8 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-const awsCredentialsName = "credential_test"
-
 var _ = Describe("AwsCredentials", func() {
+	const awsCredentialsName = "credential_test"
 	var apiKey ApiKey
```
RLRabinowitz, line 12: Why is that? This applies to all other unrelated tests

client/model.go
```diff
@@ -399,6 +399,18 @@ type EnvironmentScheduling struct {
...
 
+type WorkflowTrigger struct {
+	Id                  string        `json:"id"`
+	Name                string        `json:"Name"`
+	WorkspaceName       string        `json:"WorkspaceName"`
+	ProjectId           string        `json:"ProjectId"`
+	LatestDeploymentLog DeploymentLog `json:"latestDeploymentLog"`
+}
+
+type WorkflowTriggerCreatePayload struct {
```
RLRabinowitz, line 410: Same as above, should be Update

RLRabinowitz (approved):
👍🏻

client/aws_credentials_test.go
eranelbaz: those were in the global scope so you could get them from other tests file
if i want to declare `dummyProjectId` i couldn't since its already been declared
so i put them inside the test scope

client/api_client.go
eranelbaz: b0c6ca74(upsert + type)
also i was wrong on the return type

client/model.go
eranelbaz: b0c6ca74(upsert + type)

client/aws_credentials_test.go
RLRabinowitz: > those were in the global scope so you could get them from other tests file

🤮

eranelbaz:
81f6e732 fixed broken tests

PR #216, opened 2022-01-24
Enable query by configuration variable id in order to resolve #215

### LINKED ISSUE #215: Terraform Provider - Overriding a var with is_read_only corrupts the state
1. Add the following resources:

```
resource "env0_configuration_variable" "orgvar" {
  name = "myvar"
  value = "orgvalue"
  is_read_only = true
}

resource "env0_configuration_variable" "projvar" {
  name = env0_configuration_variable.orgvar.name
  value = "projvar"
  project_id = "<some_project_id>"
}
```

2. Apply the TF file. That was allowed
3. Going to env0, you don't see the project variable, but it was created
4. Try terraform apply again, you get an error. You now can't even remove the resource from terraform. The only thing you can do is manually remove the variables from env0

```
Error: variable 8d7d8d31-7600-409f-af3c-269ca09c5fdf not found (under this scope): <nil>
│
│   with env0_configuration_variable.projvar,
│   on main.tf line 24, in resource "env0_configuration_variable" "projvar":
│   24: resource "env0_configuration_variable" "projvar" {
```

eranelbaz:
fixes #215 
The problem is that when we query by scope with `readonly` we don't return the lower scopes

### Solution
added query by id function to our client
if we have id use it

also added test that check #215

also saw we are missing 
https://github.com/env0/terraform-provider-env0/blob/11466150976f86a7615b14116d8fa0c112613563/env0/resource_configuration_variable.go#L211-L212
in our read function


eranelbaz:
[Terraform Provider - Overriding a var with is_read_only corrupts the state](https://trello.com/c/qXjSVCH9/5309-terraform-provider-overriding-a-var-with-isreadonly-corrupts-the-state)

client/api_client.go
```diff
@@ -13,6 +13,7 @@ type ApiClient struct {
 
 type ApiClientInterface interface {
 	ConfigurationVariables(scope Scope, scopeId string) ([]ConfigurationVariable, error)
```
yaronya, line 15: Maybe worth changing existing method to `ConfigurationVariablesByScope` ?
No biggie, your call

client/configuration_variable_test.go
```diff
@@ -35,6 +35,24 @@ var _ = Describe("Configuration Variable", func() {
...
+			httpCall = mockHttpClient.EXPECT().
+				Get("/configuration/"+id, nil, gomock.Any()).
+				Do(func(path string, request interface{}, response *ConfigurationVariable) {
+					*response = mockConfigurationVariable
+				})
+
+			found, _ = apiClient.ConfigurationVariablesById(id)
+		})
+
+		It("Should return created triggers", func() {
```
yaronya, line 51: Text is wrong

env0/resource_configuration_variable.go
```diff
@@ -198,36 +198,34 @@ func resourceConfigurationVariableRead(ctx context.Context, d *schema.ResourceDa
...
 
-				if variable.Schema.Format != "" {
-					d.Set("format", variable.Schema.Format)
-				}
-			}
+	d.Set("name", variable.Name)
+	d.Set("description", variable.Description)
+	d.Set("value", variable.Value)
+	d.Set("is_sensitive", variable.IsSensitive)
+	d.Set("is_read_only", variable.IsReadonly)
```
yaronya, line 211: This means that we haven't QA-ed drift detection properly for configuration properties 😞

yaronya (approved):
LGTM
Left some minor comments

client/configuration_variable_test.go
eranelbaz: 1fde85e7(rename2)

client/api_client.go
eranelbaz: a66cb455(rename)

PR #219, opened 2022-02-02
Fix handle drift #205

### LINKED ISSUE #205: Handle drift - general
as we did in https://github.com/env0/terraform-provider-env0/pull/202

we need to do that for ALL of our resources.

Generally:
- add a condition check for determine if (isNotANewResource && NotFound/Deleted) then
- call `setId("")` to invoke destroy
- now state is synced.

samuel-br:
fixes #205

### DIFF
--- a/client/api_client_mock.go
+++ b/client/api_client_mock.go
@@ -181,11 +181,11 @@ func (mr *MockApiClientInterfaceMockRecorder) ConfigurationVariableUpdate(arg0 i
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "ConfigurationVariableUpdate", reflect.TypeOf((*MockApiClientInterface)(nil).ConfigurationVariableUpdate), arg0)
 }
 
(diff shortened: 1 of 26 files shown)


razbensimon:
@samuel-br Yo!

just here to note that we had [that](https://github.com/env0/terraform-provider-env0/pull/202) PR to fix that for templates.
Also please note that on most of resources, probably need to fix it in the regular way as Arel mentioned [here](https://github.com/env0/terraform-provider-env0/issues/155).
See [these lines from aws TF provider](https://github.com/hashicorp/terraform-provider-aws/blob/e86fca35f8f14709be928028a5550f29bbb21bc6/internal/service/s3/bucket.go#L830-L834) for an example. 👈🏼

samuel-br:
@razbensimon yap, I looked over what you fixed before, but as you mentioned not all resources  have `isDeleted` in response body, so my way here is to check err is 404, correct me if I wrong here please

PR #226, opened 2022-02-17
Add drift handling

### LINKED ISSUE #225: Handle drift - specific resources
When deleting a managed resource from below through  Env0 GUI,  drift isn't detected.
resources : 
Cloud credentials project assignment
Team project assignment 
Template project assignment 
Template

**this issue related to #205**

samuel-br:
fixes #225 
### Solution
Add check if `deleted` in read function to determine if drift happened , and invoke destroy that remove the resource from the state, instead of throwing error


env0/resource_cloud_credentials_project_assignment.go
```diff
@@ -60,6 +60,11 @@ func resourceCloudCredentialsProjectAssignmentRead(ctx context.Context, d *schem
 		}
 	}
 	if !found {
+
+		if !d.IsNewResource() {
```
yaronya, line 64: Can you explain why we need to assert this one? what's the idea behind it?
Also you shouldn't be throwing here, it's either throw or remove resource from state, not both

env0/resource_environment.go
```diff
@@ -254,6 +254,12 @@ func resourceEnvironmentRead(ctx context.Context, d *schema.ResourceData, meta i
 	if err != nil {
 		return diag.Errorf("could not get environment: %v", err)
 	}
+
+	if environment.IsArchived {
```
yaronya, line 258: Can you please bring this behavior on the slack channel? it might happen to users a lot so I think we should discuss this first

env0/resource_team_project_assignment.go
```diff
@@ -59,14 +60,20 @@ func resourceTeamProjectAssignmentRead(ctx context.Context, d *schema.ResourceDa
...
 			d.Set("project_id", assignment.ProjectId)
 			d.Set("team_id", assignment.TeamId)
 			d.Set("role", assignment.ProjectRole)
+			found = true
 			break
 		}
 	}
+	if !found {
+		d.SetId("")
+		return nil
```
yaronya, line 75: Seems like a redundant return statement

env0/resource_cloud_credentials_project_assignment_test.go
```diff
@@ -90,20 +97,41 @@ func TestUnitResourceCloudCredentialsProjectAssignmentResource(t *testing.T) {
 			mock.EXPECT().RemoveCloudCredentialsFromProject(assignment.ProjectId, assignment.CredentialId).Times(1).Return(nil)
 		})
 	})
-	t.Run("Read didnt api didnt return correct stuff", func(t *testing.T) {
+	t.Run("detect drift", func(t *testing.T) {
```
yaronya, line 100: This tests that when assignment changes then we update its values.
What about a test that asserts we removes the item from the state (i.e `d.SetId("")`?

yaronya (commented):
Nice work @samuel-br 
Left some comments/questions

Please also see my comments about the missing test case. It looks like you're missing this tests in all relevant resources

env0/resource_cloud_credentials_project_assignment.go
samuel-br: you right, I fix this. its need to be set to ""

samuel-br: and we need to add  it because if `found `is false its mean that no credentials in that context was found, so the resource manually changed

env0/resource_environment.go
samuel-br: I will, but why does it need to happen a lot? its just when they deleted the environment which means we need to destroy this in state

env0/resource_team_project_assignment.go
samuel-br: Yes it is, will be fixed

env0/resource_team_project_assignment.go
```diff
@@ -59,14 +60,20 @@ func resourceTeamProjectAssignmentRead(ctx context.Context, d *schema.ResourceDa
 		return diag.Errorf("could not get TeamProjectAssignment: %v", err)
 	}
 
+	found := false
 	for _, assignment := range assignments {
 		if assignment.Id == id {
 			d.Set("project_id", assignment.ProjectId)
 			d.Set("team_id", assignment.TeamId)
 			d.Set("role", assignment.ProjectRole)
+			found = true
```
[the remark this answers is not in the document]
samuel-br: the behavior will be the same as TF search for the id specific to determine if destroy is needed.

env0/resource_cloud_credentials_project_assignment_test.go
samuel-br: all the tests for drift scenario are the same, I create a resource with specific values and then call to update the same resource with other values, but when reading the state before `update`, I return values that cause destroy.
the tests show that no update was called and another create is invoked, same as the real behavior of TF when drift detected

env0/resource_environment.go
yaronya: As discussed we are not going to handle `environment` and `project` resources because they aren't really deleted but "archived"

env0/resource_cloud_credentials_project_assignment.go
samuel-br: 3868f818e1cd9e36156e9db1d2ce99f79f610410

env0/resource_team_project_assignment.go
samuel-br: 234334f5b27de1837aad58db525caff70bd002e0

yaronya (dismissed):

razbensimon:
https://github.com/env0/terraform-provider-env0/pull/219#issuecomment-1027975349
according to that comment, I am wondering why we don't have any call to `d.IsNewResource` ?
general behavior IIUC that drift flow - is when `d.IsNewResource` is true, we should fail/error when exists.
if `d.IsNewResource` is false **AND** we got some relevant Http Code (eg. NotFound) we should understand it as **drift** and call `SetId("")`.

I'm not 100% sure about what I said, and just raise a flag for you, because that's what I understood from the TF `Read` examples over the web (eg [AWS](https://github.com/hashicorp/terraform-provider-aws/issues/16796)).
also I saw [that](https://www.terraform.io/plugin/sdkv2/best-practices/detecting-drift) maybe it can help.

@samuel-br just validate that please, and QA your changes by running the provider after drift changes (in env0).
don't hesitate to ask for help if needed.

samuel-br:
> I think you right but as you said i want to validate it, the explenation i think, its because the d.isnewresource is true when we plan a resource its pushing it to TF state, but its not yet created on the "real world" and then every read will get exeption, but what i dont understand is how we dont use this for now and there is still no error when apply a new resource.

env0/resource_team_project_assignment.go
```diff
@@ -59,14 +60,21 @@ func resourceTeamProjectAssignmentRead(ctx context.Context, d *schema.ResourceDa
...
 			break
 		}
 	}
+	if !found {
+		if !d.IsNewResource() {
+			d.SetId("")
+		}
+	}
```
razbensimon, line 77:
```
if !d.IsNewResource() && !found {
		log.Printf("[WARN] Project (%s) not found, removing from state", d.Id())
		d.SetId("")
		return nil
	}
```

Looks better to me. I took it from [S3 bucket provider](https://github.com/hashicorp/terraform-provider-aws/blob/e86fca35f8f14709be928028a5550f29bbb21bc6/internal/service/s3/bucket.go#L830-L834) as example / reference.
Please do the same for others if can.

I am still not sure about this one. 
IIUC we should still `throw` on `d.IsNewResource()==true && someErrorOrBadCodeExists`.
I don't see any error returned on that case. @yaronya maybe you can revisit my thoughts here?

client/model.go
```diff
@@ -286,6 +286,7 @@ type Environment struct {
 	LifespanEndAt               string        `json:"lifespanEndAt"`
 	LatestDeploymentLogId       string        `json:"latestDeploymentLogId"`
 	LatestDeploymentLog         DeploymentLog `json:"latestDeploymentLog"`
+	IsArchived                  bool          `json:"isArchived"`
```
razbensimon, line 289: don't we need to use it on the logic somewhere?
just wondering

razbensimon (commented):
@samuel-br I am not sure i understood your comment.
also please write a propper description of that PR.

@yaronya @eranelbaz please revisit my comments here.
I am not sure about it...

env0/resource_team_project_assignment.go
eranelbaz: i don't think we need to throw something, but we do need to return `nil` here

yaronya: Yes, @eranelbaz is correct - we don't need to throw anything. This is what we wanted to avoid in first place.
We should be removing the resource from state.

@samuel-br haven't we previously discussed it and said that we don't really need to check whether `IsNewResource` because TF doesn't `read` the resource before `create` (only before `update`/`delete`) so this would always return `false`?

client/model.go
yaronya: The PR handled also environment resource at start but since it isn't so trivial to re-create an environment on some drift and there's a deeper product decision to make before implementing it, we've decided to ditch it for now and create a separate issue for environment and project resources.
This looks like a leftover tho.

@samuel-br did you create that seperate issue btw?

env0/resource_team_project_assignment.go
razbensimon: @yaronya i think `read ` is invoked on `create` too..

eranelbaz: `read` invokes on terraform refresh IIRC

client/model.go
eranelbaz: @yaronya do you mean for https://github.com/env0/terraform-provider-env0/issues/225?

yaronya: I'm not so sure. I expect to see an issue that includes environment and project resources in its description, but couldn't find it.

If I get that correctly:
 - #205 talks about our drift issues in general
 - #225 tries to tackle specific resources, such as template for e.g
 - We're missing another issue that tackles environment and project

env0/resource_team_project_assignment.go
samuel-br: So I checked it with some test and @yaronya is right about `d.isNewResource()` but it still can be a factor on less common senarios as @eranelbaz said about refresh and maybe on `import `too (for now those resources dont have import but to be safe). 
So ill just change later for the changes @razbensimon  suggested

client/model.go
samuel-br: nope, Ill do it later today, but I want  to summary this before on channel, cause for now there is three or for types of resources in this contexts of problem

samuel-br: I added #250 now.
In conclusion we have the main issue #205 that blocked for now because lack of info in returned errors
#225 take care the resources that have `IsDeleted ` attribute except project and environment, those mentioned in #250

eranelbaz: @samuel-br maybe just change their title?
Handle drift - general
Handle drift - specific resources
etc

and link them

env0/resource_team_project_assignment.go
razbensimon: the `if` is good now, just suggested adding logs to the user as I saw in the S3 bucket example.
for your decision @samuel-br

client/model.go
razbensimon: i am confused with all of those issues.
I just asked to make the PR with a good description (copy-paste the issue if needed to the PR).

anyways, if that line of code is related to this PR keep it.
if not - remove it

@samuel-br 
cc @eranelbaz

razbensimon (approved):

client/model.go
yaronya: I've changed the title for #250 already FYI.
And yes, prefer not having multiple issues with the same name.

yaronya: @samuel-br can you please share on the main issue (i.e #205) what's needed to be done basically and why it differs from the other issues (parsing the error codes from http client layer and etc.)

env0/resource_team_project_assignment.go
samuel-br: "drift detected: TerraForm will remove {this resource} from state" is that good wording?

razbensimon: "[WARN] Drift Detected: Terraform will remove {this resource} from state"

client/model.go
samuel-br: @yaronya @eranelbaz  I changed the issues names (@yaronya thanks on #250 ) and add https://github.com/env0/terraform-provider-env0/issues/205#issuecomment-1062822911 on the main issue, please take a look on that comment to ensure its clear enough

env0/resource_team_project_assignment.go
samuel-br: c2ec119d609520f9038ee374e765173e65633ed5

PR #227, opened 2022-02-21
Fix: template resource token_id attribute validation

yaronya:
On #224 we introduced some new schema validations for the template resource. Specifically the validation that requires `token_id` attribute to be set with `gitlab_project_id` attribute is incorrect as `token_id` is used as git token for simple git fetch when template is of type "Other"

### Solution
Remove that validation


yaronya:
[Fix template resource token_id attribute validation](https://trello.com/c/ob3YNMS5/5575-fix-template-resource-tokenid-attribute-validation)

razbensimon (approved):

PR #237, opened 2022-03-07
Add client support for drift detection

eranelbaz:
We want to add drift detection support for our terraform provider
for that, we first need to add API client support

### Solution
add client support :)

### DIFF
--- a/client/environment_drift_detection.go
+++ b/client/environment_drift_detection.go
@@ -1,0 +1,31 @@
+package client
+
+func (self *ApiClient) EnvironmentDriftDetection(environmentId string) (EnvironmentSchedulingExpression, error) {
+	var result EnvironmentSchedulingExpression
+
+	err := self.http.Get("scheduling/drift-detection/environments/"+environmentId, nil, &result)
+	if err != nil {
+		return result, err
+	}
+	return result, nil
+}
+
+func (self *ApiClient) EnvironmentUpdateDriftDetection(environmentId string, payload EnvironmentSchedulingExpression) (EnvironmentSchedulingExpression, error) {
+	var result EnvironmentSchedulingExpression
+
+	err := self.http.Put("scheduling/drift-detection/environments/"+environmentId, payload, &result)
+	if err != nil {
+		return EnvironmentSchedulingExpression{}, err
+	}
+
+	return result, nil
+}
+
+func (self *ApiClient) EnvironmentStopDriftDetection(environmentId string) error {
+	err := self.http.Put("scheduling/drift-detection/environments/"+environmentId, EnvironmentSchedulingExpression{Enabled: false, Cron: ""}, &EnvironmentScheduling{})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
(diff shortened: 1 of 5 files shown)


eranelbaz:
[Terraform Provider - Drift Detection Deployments](https://trello.com/c/7Ko5En8w/5608-terraform-provider-drift-detection-deployments)

client/environment_drift_detection.go
```diff
@@ -0,0 +1,31 @@
...
+	err := self.http.Put("/scheduling/drift-detection/environments/"+environmentId, payload, &result)
+	if err != nil {
+		return EnvironmentSchedulingExpression{}, err
+	}
+
+	return result, nil
+}
+
+func (self *ApiClient) EnvironmentStopDriftDetection(environmentId string) error {
+	err := self.http.Put("/scheduling/drift-detection/environments/"+environmentId, EnvironmentSchedulingExpression{Enabled: false, Cron: ""}, &EnvironmentScheduling{})
```
RLRabinowitz, line 25: We've discussed this briefly at sync today, @alonnoga 's changing the API to be PATCH in [this PR](https://github.com/env0/env0/pull/6758)

Please use `Patch` here

RLRabinowitz, line 25: Sorry @eranelbaz but another thing that was just raised. `cron` would now be optional, and `''` wouldn't really work well in the API, as it's a patch

https://env0.slack.com/archives/C02TRUU5YS0/p1646659554487259

RLRabinowitz, line 25: Sorry @eranelbaz but we're changing `cron` to be optional, and `''` wouldn't really work here
https://env0.slack.com/archives/C02TRUU5YS0/p1646659554487259

eranelbaz: 7b8a0530(convert to Patch)

eranelbaz: same as above

client/model.go
```diff
@@ -393,7 +393,7 @@ type PolicyUpdatePayload struct {
 
 type EnvironmentSchedulingExpression struct {
 	Cron    string `json:"cron"`
-	Enabled bool   `json:"enabled"`
+	Enabled bool   `json:"enabled,omitempty"`
```
RLRabinowitz, line 396: I think the `omitempty` should be on `cron` and not on `enabled`.

A simple test in `model_test.go` could test this behavior

client/environment_drift_detection_test.go
```diff
@@ -0,0 +1,114 @@
...
+	. "github.com/env0/terraform-provider-env0/client"
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("EnvironmentDriftDetection", func() {
+	environmentId := "env"
+	path := "/scheduling/drift-detection/environments/" + environmentId
+	mockError := errors.New("I don't like milk")
```
RLRabinowitz, line 14: Liar

RLRabinowitz (approved):
Nice one.
Since we've made some changes ([here](https://github.com/env0/env0/pull/6820/files)), now `enabled` is mandatory and `cron` is not. So you'd need to adjust to model before merging (`omitempty` is on the wrong field)

PR #238, opened 2022-03-07
Fix broken mocks

eranelbaz:
Found this problem while I generate mocks for #237 
we changed the client api but didn't generate our mocks...

### Solution
generate mocks and update code to fix `main`....
also added step to our ci that will generate mocks and will protect us from this issue


GiliFaroEnv0 (approved):
LGTM !

PR #244, opened 2022-03-08
Add drift detection resource

eranelbaz:
Add drift detection resource

### Solution
Add it

Didn't add any integration test since I feel like it will be stupid test...


docs/resources/environment_drift_detection.md
```diff
@@ -0,0 +1,48 @@
...
+## Schema
+
+### Required
+
+- **cron** (String) Cron expression for scheduled drift detection of the environment
+- **environment_id** (String) The environment's id
+
+### Optional
+
+- **id** (String) The ID of this resource.
```
GiliFaroEnv0, line 46: why?

GiliFaroEnv0 (approved):
Looks good!
left 1 comment on the generated docs If you know why it add the id as optional?
If it really allows it you need o changed your logic using the id

docs/resources/environment_drift_detection.md
GiliFaroEnv0: If it really allows it you need o changed your logic using the id

eranelbaz: each tf resource have `id` which is the resource `id`
in this case the actual resource `id` doesn't matter but only the environment `id` which we set

eranelbaz: and it's optional since i didn't specify it as part of the schema

PR #247, opened 2022-03-08
Fix: env0_project data error message (#89)

### LINKED ISSUE #89: Confusing error when ID/name doesn't exists for `data "env0_project"`
1. Create a TF file with an `env0_project` data with a non-existent ID
```
data "env0_project" "non_existent" {
  id = "non-existent-id"
}
```
2. `terraform apply`
3. The failure is unclear, mentions failure to query template: "Could not query template: 404 Not found: not found"

TomerHeber:
fixes #89 

### Solution
Added a new error type. 
If a request is unsuccessful (E.g., the response returned 404), create the new error type.
In env0_project data, returns an appropriate message based on the status code. (In the case 404 returns an "id not found" message).


client/http/errors.go
```diff
@@ -0,0 +1,15 @@
...
+
+import "github.com/go-resty/resty/v2"
+
+type FailedResponseError struct {
+	res *resty.Response
+}
+
+func (e *FailedResponseError) Error() string {
+	return e.res.Status() + ": " + string(e.res.Body())
+}
```
RLRabinowitz, line 11: When is this ever used? Is `Error` called anywhere along the way when displaying errors?

env0/data_project.go
```diff
@@ -101,6 +102,11 @@ func getProjectById(id string, meta interface{}) (client.Project, diag.Diagnosti
 	apiClient := meta.(client.ApiClientInterface)
 	project, err := apiClient.Project(id)
 	if err != nil {
+		if frerr, ok := err.(*http.FailedResponseError); ok {
+			if frerr.NotFound() {
+				return client.Project{}, diag.Errorf("Could not find a project with id: %s", id)
```
RLRabinowitz, line 107: Please add a test case for this in `data_project_test.go`
Mock 404 and expect this error

RLRabinowitz (commented):
Looks good!

We're just missing a test for this fix

client/http/errors.go
TomerHeber: In general for a struct to be considered an error, it must implement Error().

env0/data_project.go
TomerHeber: This has been added.

yaronya (approved):

Issue #232, opened 2022-02-28
Add support for AWS access key Credentials


RLRabinowitz:
Terraform Provider should support upcoming cloud deployment credentials

### `aws_credentials` resource:

Should also support AWS access keys:
- Add new fields to `aws_credentials` - `access_key_id` and `secret_access_key`
- [`arn`, `external_id`] and `[`access_key_id`, `secret_access_key`] should be mutually exclusive!
  - If `arn` is set, then `external_id` is optional, but `access_key_id` and `secret_access_key` cannot be set
  - If `access_key_id` is set then `secret_access_key` is mandatory, and both `arn` and `external_id` cannot be set
- For the create API call
  - If `arn` is set, `type` should be `AWS_ASSUMED_ROLE_FOR_DEPLOYMENT`. If `access_key_id` is set, `type` should be `AWS_ACCESS_KEYS_FOR_DEPLOYMENT`
  - Build the value object according to the type. Either `{ roleArn, externalId }` or `{ accessKeyId, secretAccessKey }`
- When importing, only accept credentials that are AWS (`AWS_ASSUMED_ROLE_FOR_DEPLOYMENT` or `AWS_ACCESS_KEYS_FOR_DEPLOYMENT`). This is can be done in a separate issue though

yaronya:
@GiliFaroEnv0 @RLRabinowitz was this fully implemented? if so, can we please close this issue?

PR #249, opened 2022-03-08
Feat: Add support with AWS access key cred

GiliFaroEnv0:
Solve #232 
### Solution

Add support to AWS access key credentials

### DIFF
--- a/client/aws_credentials.go
+++ b/client/aws_credentials.go
@@ -40,7 +40,6 @@ func (self *ApiClient) AwsCredentialsCreate(request AwsCredentialsCreatePayload)
 		return ApiKey{}, err
 	}
 
-	request.Type = "AWS_ASSUMED_ROLE_FOR_DEPLOYMENT"
 	request.OrganizationId = organizationId
 
 	var result ApiKey
--- a/client/aws_credentials_test.go
+++ b/client/aws_credentials_test.go
@@ -30,15 +30,17 @@ var _ = Describe("AwsCredentials", func() {
 		BeforeEach(func() {
 			mockOrganizationIdCall(organizationId)
 
+			payloadValue := AwsCredentialsValuePayload{
+				RoleArn:    "role",
+				ExternalId: "external",
+			}
+
 			httpCall = mockHttpClient.EXPECT().
 				Post("/credentials", AwsCredentialsCreatePayload{
 					Name:           awsCredentialsName,
 					OrganizationId: organizationId,
 					Type:           "AWS_ASSUMED_ROLE_FOR_DEPLOYMENT",
-					Value: AwsCredentialsValuePayload{
-						RoleArn:    "role",
-						ExternalId: "external",
-					},
+					Value:          payloadValue,
 				},
 					gomock.Any()).
 				Do(func(path string, request interface{}, response *ApiKey) {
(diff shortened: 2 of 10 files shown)


client/aws_credentials.go
```diff
@@ -40,7 +40,6 @@ func (self *ApiClient) AwsCredentialsCreate(request AwsCredentialsCreatePayload)
 		return ApiKey{}, err
 	}
 
-	request.Type = "AWS_ASSUMED_ROLE_FOR_DEPLOYMENT"
```
HeverFarber, line 43: It could break the usage of customers.

I agree this default is ugly - but if someone uses this resource. it will force him to commit change before new deploy. 

maybe use the default with warning message

client/aws_credentials_test.go
```diff
@@ -30,27 +30,27 @@ var _ = Describe("AwsCredentials", func() {
...
 
 			apiKey, _ = apiClient.AwsCredentialsCreate(AwsCredentialsCreatePayload{
-				Name: awsCredentialsName,
-				Value: AwsCredentialsValuePayload{
-					RoleArn:    "role",
-					ExternalId: "external",
-				},
+				Name:  awsCredentialsName,
+				Value: payloadValue,
+				Type:  "AWS_ASSUMED_ROLE_FOR_DEPLOYMENT",
```
liranfarage89, line 53: can't you use the const types you created in `model.go`?

client/aws_credentials.go
liranfarage89: @HeverFarber but I think it is still backward's compatible... the old usage still requires `arn`& `externa_id`, and the `Type` was assigned under the hood.

liranfarage89 (approved):
LGTM - but let's wait for @HeverFarber response before merging it.

client/aws_credentials.go
GiliFaroEnv0: @HeverFarber  as @liranfarage89 said Its backward's compatible.

HeverFarber: We should keep backward's compatible. for the who doesn't add type before. we can add it instead of him

liranfarage89: but the type was not exposed before that...

HeverFarber: Correct - so we added it for the user. so the user didn't add it to his resource.

Now when we removed it - the resource without this type - and when we will send it to the server we will get an error because type is missing

liranfarage89: the Type was not a part of the resource either way, it's computed

eranelbaz: I don't think the user will need to approve a change since the `Type` isn't part of the schema,
and the removal is in the `client` but @GiliFaroEnv0 is calculating the type on runtime in the "resource" side and sends it to the `client` here 

https://github.com/env0/terraform-provider-env0/blob/0ea2c1771e18d41948de435cd344a0ebe548b868/env0/resource_aws_credentials.go#L72-L92

and you guys can check it, run `terraform apply` using `main`, then checkout this branch and run `apply` again and check if there is any change

GiliFaroEnv0: OK it works I merge IT 👍🏿 
bors r+

PR #252, opened 2022-03-09
Feat cloud cost credentials #218

### LINKED ISSUE #218: support cloud cost credentials
We'd like to add support for our cost credentials.
We'd need the following new resources (and data for each resource)

### cost_credentials_project_assignment

Exactly the same as `cloud_credentials_project_assignment` resource and data. The only difference is that we use different APIs:
* [Create](https://developer.env0.com/docs/api/b3A6MTkyNjc1NzI-associate-cost-credentials-with-project)
* [Delete](https://developer.env0.com/docs/api/b3A6MTkyNjc1NzU-dissociate-cost-credentials-from-project)
* [Read](https://developer.env0.com/docs/api/b3A6MTkyNjc1NzM-list-cost-credentials-associated-to-project) - Notice the response format is different than [this](https://developer.env0.com/docs/api/b3A6MTkyNjc1NzE-list-deployment-a-project-s-deployment-credentials)

### aws_cost_credentials

Exactly the same as `aws_credentials` resource and data (which is a deploy credential, not a cost credential). The schema is exactly the same, and the API calls are the same. The only difference is that the `type` that should be `AWS_ASSUMED_ROLE`:
* [Create](https://developer.env0.com/docs/api/b3A6ODc4NjAwOQ-create-cloud-provider-cost-credentials) should be called with correct `type` and data in `value`
* data should filter by type = `AWS_ASSUMED_ROLE`

### azure_cost_credentials

Similar to `aws_cost_credentials` resource and data. Should use the same APIs, where type is `AZURE_CREDENTIALS `. For the schema - see [docs](https://developer.env0.com/docs/api/b3A6ODc4NjAwOQ-create-cloud-provider-cost-credentials) - value:
* clientId - required
* clientSecret - required and sensitive
* tenantId - required
* subscriptionId - required
Other than the schema, differences in API calls are as follows:
* [Create](https://developer.env0.com/docs/api/b3A6ODc4NjAwOQ-create-cloud-provider-cost-credentials) should be called with correct `type` and data in `value`
* data should filter by type = `AZURE_CREDENTIALS `

### google_cost_credentials

Similar to `aws_cost_credentials` resource and data. Should use the same APIs, where type is `GCP_CREDENTIALS`. For the schema - see [docs](https://developer.env0.com/docs/api/b3A6ODc4NjAwOQ-create-cloud-provider-cost-credentials) - value:
* tableId - required
* secret - required and sensitive
Other than the schema, differences in API calls are as follows:
* [Create](https://developer.env0.com/docs/api/b3A6ODc4NjAwOQ-create-cloud-provider-cost-credentials) should be called with correct `type` and data in `value`
* data should filter by type = `GCP_CREDENTIALS`

samuel-br:
resolves #218 
### Solution
Add the relevant resources and data


eranelbaz (changes requested):
I'm sorry @samuel-br but this PR is wayyy to big for someone to review...
please split this pr
one way to split it is by
- pr for api client
- pr for resources

and please try to make the code a bit more generic, maybe the cost model can contain everything with omit empty so we can use the same common code for all the resource but with different schema

LMK if you need help

eranelbaz:
Closing this one as we you open the splits 
so no need for this one

PR #257, opened 2022-03-10
Feat: Add Azure cloud credentials

### LINKED ISSUE #248: Support new GCP & Azure credentials
### New `gcp_credentials` resource:

- Should call same create/list/delete APIs as `aws_credentials` resource, but with type `GCP_SERVICE_ACCOUNT_DEPLOYMENT`
- Fields should be `project_id` and `service_account_key`, on API calls `value` should be an object of those fields. `service_account_key` should be secret
- When importing, only accept credentials that are GCP (`GCP_SERVICE_ACCOUNT_DEPLOYMENT`). This is can be done in a separate issue though

### New `azure_credentials` resource:

- Should call same create/list/delete APIs as `aws_credentials` resource, but with type `AZURE_SERVICE_PRINCIPAL_FOR_DEPLOYMENT`
- Fields should be `clientId`, `clientSecret`, `tenantId` and `subscriptionId`, on API calls `value` should be an object of those fields. `clientSecret` should be secret
- When importing, only accept credentials that are AZURE (`AZURE_SERVICE_PRINCIPAL_FOR_DEPLOYMENT`). This is can be done in a separate issue though

liranfarage89:
resolves #248


liranfarage89:
Opening a new PR

PR #258, opened 2022-03-10
Feat: Add azure credentials

### LINKED ISSUE #248: Support new GCP & Azure credentials

liranfarage89:
resolves #248
### Solution
Add azure cloud credentials


env0/resource_aws_credentials.go
```diff
@@ -120,24 +115,3 @@ func resourceAwsCredentialsDelete(ctx context.Context, d *schema.ResourceData, m
 	}
 	return nil
 }
-
-func resourceAwsCredentialsImport(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
-	id := d.Id()
```
liranfarage89, line 125: Import is not relevant so I removed it from AWS

liranfarage89:
[Add Support for GCP and Azure credentials in our Terraform provider](https://trello.com/c/SCz0Q494/5651-add-support-for-gcp-and-azure-credentials-in-our-terraform-provider)

HeverFarber (approved):
LGTM

env0/data_azure_credentials.go
```diff
@@ -0,0 +1,93 @@
+package env0
+
+import (
```
RLRabinowitz, line 3: Why do we need a `data` source for this? I see we have it for the other credentials, but it seems very unnecessary to me (and it won't even work as far as I understand, because we redact the secret values)

env0/resource_azure_credentials.go
```diff
@@ -0,0 +1,101 @@
...
+				Description: "the azure client id",
+				Required:    true,
+				Sensitive:   true,
+				ForceNew:    true,
+			},
+			"client_secret": {
+				Type:        schema.TypeString,
+				Description: "the azure client secret",
+				Required:    true,
+				Sensitive:   true,
```
RLRabinowitz, line 34: Technically this is the only field that's actually sensitive, but I don't think that matters much
I'm fine with keeping all of those `Sensitive: true`

RLRabinowitz (approved):

env0/data_azure_credentials.go
liranfarage89: [0f20b41](https://github.com/env0/terraform-provider-env0/pull/258/commits/0f20b41f2f2eeeee4eee08cad889f47c3bdb3824)
[d46d1de](https://github.com/env0/terraform-provider-env0/pull/258/commits/d46d1ded48b6adffddb842b135e81fcae9a29d2c)

HeverFarber: I think is very important data here.

1. manage the credential from the UI and only connect it to the project.
2. manage the credential in other stuck (organization stack)

HeverFarber: I meant you manage X stacks for our provider for example:

org stack:
```
resource azure_cred {
   name: 'my_production_keys'
}
```

project stack:
```
data azure_cred {
  name: 'my_production_keys'
}

resource env0_project_credentials  {
  project_id: env0.project.id
  cred_id: data.azure_cred.id
  type: AzureSecret
}
```

PR #263, opened 2022-03-11
Feat: project deletion with force-destroy flag (#67)

### LINKED ISSUE #67: Project deletion should be with flag
Only if there are no environments in the project it should be archived.

If there are - you should get an error.
There should be a flag for "force_destroy" (like in aws_s3_bucket) , when it's `true` then we should archive the project even if it has environments in it.

TomerHeber:
resolves #67 

### Solution
Implemented the API call that returns environments for a given project id.
Added the force-destroy flag to the project schema (state only) (defaults to false).
When destroying a project now checks if it has the flag and any active environments and acts accordingly:

- If no environments - delete the project.
- If has environments and force-destroy is true - delete the project.
- If has environments and force-destroy is false - return an error.


env0/resource_project.go
```diff
@@ -106,8 +114,44 @@ func resourceProjectDelete(ctx context.Context, d *schema.ResourceData, meta int
...
+		for {
+			envs, err := apiClient.ProjectEnvironments(id)
+			if err != nil {
+				return diag.Errorf("could not get project environments: %v", err)
+			}
+
+			// Filter out archived (destroyed) environments.
+			hasActiveEnvs := false
+
+			for _, env := range envs {
```
TomerHeber, line 135: Deleting an environment takes time. One option is to fix the delete environment to wait until it's deleted (my preferred option).

Another option would be to call delete on each "active" environment (I can then increase the timeout from 1 minute to 5 or 10).

(At the moment, there's no point in waiting 5-10 minutes if an environment remains active).

Let me know your thoughts.

env0/resource_project_test.go
```diff
@@ -84,3 +85,84 @@ func TestUnitProjectInvalidParams(t *testing.T) {
...
+			},
+		}
+
+		runUnitTest(t, testCase, func(mock *client.MockApiClientInterface) {
+			mock.EXPECT().ProjectCreate(client.ProjectCreatePayload{
+				Name:        project.Name,
+				Description: project.Description,
+			}).Times(1).Return(project, nil)
+
+			gomock.InOrder(
```
GiliFaroEnv0, line 127: no point to use in order for 1 check

env0/resource_project.go
```diff
@@ -106,8 +114,45 @@ func resourceProjectDelete(ctx context.Context, d *schema.ResourceData, meta int
 	apiClient := meta.(client.ApiClientInterface)
 
 	id := d.Id()
-	err := apiClient.ProjectDelete(id)
-	if err != nil {
+
+	forceDestroy := d.Get("force_destroy").(bool)
+
+	if !forceDestroy {
```
GiliFaroEnv0, line 120: split this logic into smaller functions
so it will be more readable
something like
* hasActiveEnv
* (for the big loop) validateNoActivceEnvironments
just a suggestion since its not clean

GiliFaroEnv0 (commented):
Good job! few minor comments 
And follow the slack conversation about the environment deletion
Also, add it to Integration tests

env0/resource_project.go
GiliFaroEnv0: https://env0.slack.com/archives/C01QBBEGBEH/p1647344145698209

env0/resource_project.go
TomerHeber: Hi @GiliFaroEnv0 ,

Thank you for the comments.
I'm still going over your comments and will refactor it.
However, before I proceed - I have added this comment in Slack:
https://env0.slack.com/archives/C01QBBEGBEH/p1647378021284789?thread_ts=1647344145.698209&cid=C01QBBEGBEH

Happy to discuss further. Thanks!

TomerHeber: @GiliFaroEnv0 - applied changes as discussed in the slack thread.
https://github.com/env0/terraform-provider-env0/pull/263/commits/f850c770528f357af8ad4654d4dbf151970ccaa7

env0/resource_project_test.go
TomerHeber: Fixed.
https://github.com/env0/terraform-provider-env0/pull/263/commits/f850c770528f357af8ad4654d4dbf151970ccaa7

GiliFaroEnv0 (approved):
Great Job!

TomerHeber:
> Great Job!

Thank you @GiliFaroEnv0 !
I think there's an issue with the github hooks (outage).
Is there a way to re-run the checks?

PR #271, opened 2022-03-15
Chore: filter deleted Templates from data

liranfarage89:
We still get errors on multiple templates with the same name, this is because we don't filter out deleted templates.

### Solution

- add it.
- append random string to template name for concurrency.


eranelbaz (approved):

avnerenv0:
@liranfarage89 sorry being very late here
but why isn't this filtered on the API side?

PR #273, opened 2022-03-15
Chore: Add configuration for waiting for environment to finish deployment

### LINKED ISSUE #272: Add option to synchronously wait for environment to finish deploying
Currently the `env0_environment` `resource` is considered "done" when the deployment has started.
If you'd like to rely on the environment only when it's done (for example, get environment variables, or rely on deployment outputs), you can't really do that without adding some sleep

We should have some configuration on the `env0_environment` resource that would allow us to pick whether or not to wait for full deployment

RLRabinowitz:
resolves https://github.com/env0/terraform-provider-env0/issues/272

We should allow users to choose whether or not they'd like to wait for environment to finish deploying before continuing to the next terraform resource dependency

### Solution
Add `wait_for` field which supports `ACK` and `FULLY_DEPLOYED`. 
`ACK` is the default behaviour, which is the current behaviour - Resource is considered created once deployment is starting/queued
`FULLY_DEPLOYED` will wait for the deployment to reach final status before proceeding.

For timeouts while polling deployment status - we'll just rely on TF's default timeout for resources


tests/integration/012_environment/main.tf
```diff
@@ -16,12 +16,17 @@ resource "env0_environment" "example" {
   name          = "environment"
   project_id    = env0_project.test_project.id
   template_id   = env0_template.template.id
+  wait_for      = "FULLY_DEPLOYED"
   configuration {
     name  = "environment configuration variable"
     value = "value"
   }
   approve_plan_automatically = true
-  revision                   = "master"
```
RLRabinowitz, line 24: Changed the integration test back to what it used to be, minus the `time_sleep` to wait for deployment to finish

tests/integration/012_environment/main.tf
```diff
@@ -24,6 +25,11 @@ resource "env0_environment" "example" {
   revision                   = "master"
 }
 
+data "env0_configuration_variable" "env_config_variable" {
```
RLRabinowitz, line 28: Return this environment configuration variable test, now that it's fully support in our TF provider

RLRabinowitz:
[Terraform Provider - Allow for environment creation to be synchronous](https://trello.com/c/LdzDr22f/5970-terraform-provider-allow-for-environment-creation-to-be-synchronous)

RLRabinowitz:
[Terraform Provider - Allow for environment creation to be synchronous](https://trello.com/c/LdzDr22f/5970-terraform-provider-allow-for-environment-creation-to-be-synchronous)

RLRabinowitz:
[Terraform Provider - Allow for environment creation to be synchronous](https://trello.com/c/LdzDr22f/5970-terraform-provider-allow-for-environment-creation-to-be-synchronous)

client/api_client.go
```diff
@@ -62,13 +62,14 @@ type ApiClientInterface interface {
 	ProjectEnvironments(projectId string) ([]Environment, error)
 	Environment(id string) (Environment, error)
 	EnvironmentCreate(payload EnvironmentCreate) (Environment, error)
-	EnvironmentDestroy(id string) (Environment, error)
```
RLRabinowitz, line 65: This API was wrong

jwjs36987:
FYI, since creating the original patch, I have discovered a unfortunate issue in this behaviour. If the deploy fails, then terraform marks the environment as tainted. This causes the next apply to destroy and re-create the environment rather than just re-deploying it. I don't know if there is a way to work around this.

RLRabinowitz:
> If the deploy fails, then terraform marks the environment as tainted

Hey @jwjs36987, that's how terraform automatically behaves when a resource creation fails (doesn't happen for updates/modifications) after some state for that resources has been persisted. 
If we'd like failures to stop the TF provider, the only alternative would be to not store the state at all, which would be worse, in my opinion. That would mean that after a failure, an apply would simply create a new environment, forgetting about the old one.

The only way to fix this issue would be to not consider deployment failure as TF provider failure. That means that if an environment deployment fails, terraform will consider the resource as created successfully, and would continue on to other dependant resources.
I could look into logging a warning in such a scenario, not sure it's possible during apply, though.

What do you think about that behaviour? We can also configure it in such a way that this only happens on create, and not on subsequent redeploys/destroys

jwjs36987:
That makes sense. I think the current behaviour makes the most sense and mirrors what other providers do (e.g. the official kubernetes provider).

liranfarage89 (approved):
🚀 
Left some comments, take a look before merging 🙏

env0/resource_environment.go
```diff
@@ -109,6 +112,19 @@ func resourceEnvironment() *schema.Resource {
 				Description: "destroy safegurad",
 				Optional:    true,
 			},
+			"wait_for": {
+				Type:        schema.TypeString,
+				Description: "whether or not to wait for environment to fully deploy",
+				Optional:    true,
+				Default:     "ACK",
+				ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
```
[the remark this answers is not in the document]
RLRabinowitz: Tried it, and then figured out it's something different.

It means that a field is mutually exclusive with another field, it's not a list of possible values for a single field.
I couldn't find a different solution for this 🤷🏻‍♂️

env0/resource_environment.go
```diff
@@ -109,6 +112,19 @@ func resourceEnvironment() *schema.Resource {
...
+					value := val.(string)
+					if value != "ACK" && value != "FULLY_DEPLOYED" {
```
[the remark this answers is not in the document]
RLRabinowitz: That's current behaviour for customers, we didn't really wanna change that.

However, I'm not entirely sure any users would actually prefer using `"ACK"`, but let's leave it for now and get feedback

PR #289, opened 2022-03-24
Feat: added support for bitbucket server and github enterprise in tem…

### LINKED ISSUE #222: Support GitHub Enterprise on Templates
Right now we don't have a way to add a GitHub Enterprise template.
We should add an is_github_enterprise to match the same field on [our api](https://developer.env0.com/docs/api/b3A6MTg2NzAxMTQ-create-template)

### LINKED ISSUE #223: Support Bitbucket Server Templates
Right now we don't have a way to add a Bitbucket template.
We should add an is_bitbucket_server to match the same field on [our api](https://developer.env0.com/docs/api/b3A6MTg2NzAxMTQ-create-template)

TomerHeber:
…plate resource

resolves #222 
resolves #223 

### Solution

Added the required fields in the resource and the model.
Added additional unit tests.
Refactored the VCS "conflicts with".


env0/resource_template.go
```diff
@@ -22,6 +23,40 @@ func resourceTemplate() *schema.Resource {
...
+	 *		GitHub - githubInstallationId
+	 *		GitLab -  gitlabProjectId and tokenId
+	 *		Bitbucket - bitbucketClientKey
+	 *		GH Enterprise - isGitHubEnterprise
+	 *		GL Enterprise - isGitLabEnterprise
+	 *		BB Server - isBitbucketServer
+	 *		Other - tokenId (optional field)
+	 */
+
+	allVCSAttributes := []string{
```
TomerHeber, line 37: A small refactor to cover all the "ConflictsWith" use-cases.

env0/resource_template_test.go
```diff
@@ -313,10 +407,12 @@ func TestUnitTemplateResource(t *testing.T) {
 				TokenId:              templateUseCase.updatedTemplate.TokenId,
 				Path:                 templateUseCase.updatedTemplate.Path,
 				Revision:             templateUseCase.updatedTemplate.Revision,
-				Type:                 client.TemplateTypeTerragrunt,
+				Type:                 client.TemplateType(templateUseCase.updatedTemplate.Type),
```
TomerHeber, line 410: fixed a small issue with the unit tests.

env0/resource_template.go
```diff
@@ -22,6 +23,40 @@ func resourceTemplate() *schema.Resource {
...
+		"is_gitlab_enterprise",
+		"is_bitbucket_server",
+		"is_github_enterprise",
+	}
+
+	allVCSAttributesBut := func(strs ...string) []string {
+		sort.Strings(strs)
+		ret := []string{}
+
+		for _, attr := range allVCSAttributes {
```
HeverFarber, line 51: IMO - with [map](https://stackoverflow.com/questions/38579485/golang-convert-slices-into-map) is more elegant

HeverFarber (approved):
Hi Tomer :-) 

I like the conflict refactor

PR #290, opened 2022-03-24
Fix different behavior when creating policy #201

### LINKED ISSUE #201: Project Policy - Requires Approval Default behavior is inconsistent with manual project creation
When creating a project manually, the project policy setting "Approve plans automatically for new environments" is un-checked by default. 

When creating a project with terraform provide the project policy setting "Approve plans automatically for new environments" is checked by default.

This is also confusing because the UI and Terraform Provider field are opposite statements:
 "Approve Plans Automatically for New Environments" unchecked/false === "requires_approval_default"  true

samuel-br:
fixes #201 
### Solution
Add default value `true` to attribute `requires_approval_default` what cause that when creating project policy the outcome will be the same as create the project with Env0 UI


samuel-br:
@RLRabinowitz  does it close the issue? the problem with the wrong wording isn't fix here, just make sure the behavior the same.
also I don't think its need a UT, correct me if I wrong.

yaronya (commented):
@samuel-br please also add some UT around this (that asserts the default value of `true` being sent on API request

samuel-br:
> @samuel-br please also add some UT around this (that asserts the default value of `true` being sent on API request

6ce7065ada02f747f9437a1f1c4c7b7ae2e78b72

yaronya (approved):

PR #293, opened 2022-03-27
Fix - Add validation for case when isReadOnly and isRequired are true and value empty in config variable

### LINKED ISSUE #217: Should not allow for readonly+required+no value
The bug is -
1. In TF Provider, create a variable with no value, `is_read_only = true`, `is_required = true`.
2. Apply - Successful (Shouldn't be)
3. Go to the variables page in env0, see your variable there. Try to add a new one and save
4. Failed, validation failed since var can't be readonly+required, and have no value

What needs to be done here is to add validation that throws exception when both are set to `true`

samuel-br:
fixes #217 
### Solution
Add validation func that throw error when `is_required` and `is_read_only` are `true` and `value ` is empty


eranelbaz (commented):
What about using `ConflictsWith`?

samuel-br:
> What about using `ConflictsWith`?

@eranelbaz 
`ConflictsWith` is work as 'or' operator and we need 'and'

samuel-br:
also I did try to find a way to check it in the schema its self but it seems to be known issue on TF that there is no validation relating to more then one key

eranelbaz:
> > What about using `ConflictsWith`?
> 
> @eranelbaz `ConflictsWith` is work as 'or' operator and we need 'and'

Yes, it will say either a or b
and that is what we want, no?

samuel-br:
> > > What about using `ConflictsWith`?
> > 
> > 
> > @eranelbaz `ConflictsWith` is work as 'or' operator and we need 'and'
> 
> Yes, it will say either a or b and that is what we want, no?

no key is conflict with just one of the other keyes

isRequired conflict with isReadOnly **and** empty value
isReadOnly conflict with isRequired **and** empty value
empty value conflict with isRequired **and** isReadOnly 

so you cant ensure that with `ConflictsWith`

env0/resource_configuration_variable_test.go
```diff
@@ -602,4 +602,34 @@ resource "%s" "test" {
 			mock.EXPECT().ConfigurationVariableDelete(configVarImport.Id).Times(1).Return(nil)
 		})
 	})
+
+	t.Run("cant be empty value when isRequired and isReadOnly are true", func(t *testing.T) {
```
eranelbaz, line 606: add another case for update
where you update to empty value when `isRequired==isReadonly==true`

samuel-br:
@eranelbaz 
https://github.com/hashicorp/terraform-plugin-sdk/issues/233

env0/resource_configuration_variable_test.go
samuel-br: f3606c7ad0fdc72a78fc51bee819696706761ca8

env0/resource_configuration_variable_test.go
```diff
@@ -602,4 +602,101 @@ resource "%s" "test" {
...
+			updateParams.Format = client.HCL
+			updateParams.IsReadonly = *newConfigVar.IsReadonly
+			updateParams.IsRequired = *newConfigVar.IsRequired
+
+			mock.EXPECT().ConfigurationVariableCreate(createParams).Times(1).Return(newConfigVar, nil)
+			gomock.InOrder(
+				mock.EXPECT().ConfigurationVariablesById(newConfigVar.Id).Times(2).Return(newConfigVar, nil),
+				//mock.EXPECT().ConfigurationVariablesById(newConfigVar.Id).Return(newConfigVar, nil),
+			)
+			//	mock.EXPECT().ConfigurationVariableUpdate(client.ConfigurationVariableUpdateParams{CommonParams: updateParams, Id: newConfigVar.Id}).Times(1).Return(configVar, nil)
```
eranelbaz, line 698: please remove commented code

eranelbaz (approved):
LGTM

PR #294, opened 2022-03-27
Feat - Add resource and data  for aws cost credentials

samuel-br:
part of the solution for  #218
### Solution
Add cost credentials resource and data for aws credentials
**this PR is not resolving #218 by itself as it lacks cost cred for the rest of the platforms**
Part of #218 with #295,#296,#297


env0/data_aws_cost_credentials.go
```diff
@@ -0,0 +1,92 @@
...
+		ReadContext: dataAwsCostCredentialsRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Description:  "the name of the credential",
+				Optional:     true,
+				ExactlyOneOf: []string{"name", "id"},
+			},
+			"id": {
```
yaronya, line 22: Why not adding the role arn as well here? one might wanna use the arn of one cost creds in other creds

yaronya (commented):
Looks good overall
Left a couple of comments

env0/data_aws_cost_credentials.go
samuel-br: I copied the functionality from the general cred resource, if its needed Ill add it with the changes I am working on

Issue #281, opened 2022-03-21
Add support for API key resource


yaronya:
Should support both resource and data resource.
Available via env0 UI: `Organization Settings -> API keys`

PR #301, opened 2022-03-27
Feat: add support for API key resource

TomerHeber:
### Solution

Implement api-keys API client calls.

### DIFF
--- a/client/api_client.go
+++ b/client/api_client.go
@@ -88,6 +88,9 @@ type ApiClientInterface interface {
 	GitTokens() ([]GitToken, error)
 	GitTokenCreate(payload GitTokenCreatePayload) (*GitToken, error)
 	GitTokenDelete(id string) error
+	APIKeyCreate(payload APIKeyCreatePayload) (*APIKey, error)
+	APIKeyDelete(id string) error
+	APIKeys() ([]APIKey, error)
 }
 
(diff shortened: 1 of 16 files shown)


TomerHeber:
Part of #281

client/api_client.go
```diff
@@ -84,6 +84,9 @@ type ApiClientInterface interface {
 	ModuleDelete(id string) error
 	ModuleUpdate(id string, payload ModuleUpdatePayload) (*Module, error)
 	Modules() ([]Module, error)
+	APIKeyCreate(payload APIKeyCreatePayload) (*APIKey, error)
```
yaronya, line 87: Would be best to keep it as pascal case (i.e `ApiKeyCreate`)

yaronya (commented):
Overall looks great
Left a small comment about naming

client/api_client.go
TomerHeber: @yaronya 
arghhh! it looks like ApiKey was already used.
In any case, I renamed lots of stuff. Hopefully it makes more sense now.

https://github.com/env0/terraform-provider-env0/pull/301/commits/29d5364208e8dfde091ebde87dffcbfe75835869

TomerHeber: 4214ea271728fcf7965e3c3e4f10f5baa70ccf9e

yaronya (approved):

Issue #284, opened 2022-03-21
Handle drift in all other resources


yaronya:
We've had several issues about resources not handling drifts properly. This issue settles all other resources that aren't covered or tracked. 
Issue is that when one of the following resource types is not found we throw an error instead of simply trying to recreate them (the TF way of handling drifts):
- `env0_ssh_key`
- `env0_aws_credentials`
- `env0_gcp_credentials`
- `env0_azure_credentials`
- `env0_team`
- `env0_notification`

Note that this GH issue shouldn't get closed unless all resource types are covered by a proper drift detection logic.

PR #303, opened 2022-03-28
Fix: handle drift in all resources

TomerHeber:
### Solution

Added drift handling for AWS, Azure, and GCP credentials resources.
Part of #284.


yaronya (approved):
👌

PR #306, opened 2022-03-29
Fix: handle drift in all resources

### LINKED ISSUE #284: Handle drift in all other resources

TomerHeber:
fixes #284 

### Solution
Some refactoring in ssh key resource, data and client.
Detects drift for sshkey and resets the id, instead of returning an error.

### DIFF
--- a/client/sshkey.go
+++ b/client/sshkey.go
@@ -1,31 +1,17 @@
 package client
 
-import (
-	"errors"
-)
-
-func (self *ApiClient) SshKeyCreate(payload SshKeyCreatePayload) (SshKey, error) {
-	if payload.Name == "" {
-		return SshKey{}, errors.New("Must specify ssh key name on creation")
-	}
-	if payload.Value == "" {
-		return SshKey{}, errors.New("Must specify ssh key value (private key in PEM format) on creation")
-	}
-	if payload.OrganizationId != "" {
-		return SshKey{}, errors.New("Must not specify organizationId")
-	}
+func (self *ApiClient) SshKeyCreate(payload SshKeyCreatePayload) (*SshKey, error) {
 	organizationId, err := self.organizationId()
 	if err != nil {
-		return SshKey{}, nil
+		return nil, err
 	}
 	payload.OrganizationId = organizationId
 
 	var result SshKey
-	err = self.http.Post("/ssh-keys", payload, &result)
-	if err != nil {
-		return SshKey{}, err
+	if err := self.http.Post("/ssh-keys", payload, &result); err != nil {
+		return nil, err
 	}
-	return result, nil
+	return &result, nil
 }
 
 func (self *ApiClient) SshKeyDelete(id string) error {
--- a/env0/data_sshkey.go
+++ b/env0/data_sshkey.go
@@ -2,6 +2,7 @@ package env0
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/env0/terraform-provider-env0/client"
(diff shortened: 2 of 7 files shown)


client/sshkey.go
```diff
@@ -1,31 +1,17 @@
 package client
 
-import (
-	"errors"
-)
-
-func (self *ApiClient) SshKeyCreate(payload SshKeyCreatePayload) (SshKey, error) {
-	if payload.Name == "" {
```
TomerHeber, line 8: Terraform schema prevents this from happening. These checks are not required.

env0/resource_sshkey.go
```diff
@@ -54,9 +54,13 @@ func resourceSshKeyCreate(ctx context.Context, d *schema.ResourceData, meta inte
 }
 
 func resourceSshKeyRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-	_, err := getSshKeyById(d.Id(), meta)
+	sshKey, err := getSshKeyById(d.Id(), meta)
 	if err != nil {
-		return err
+		return diag.Errorf("could not get ssh key: %v", err)
+	}
+	if sshKey == nil {
```
TomerHeber, line 61: This handles the drift.

env0/data_sshkey.go
```diff
@@ -31,23 +32,20 @@ func dataSshKey() *schema.Resource {
...
-		if err != nil {
-			return err
-		}
+	}
 
+	if err != nil {
+		return diag.Errorf("could not read ssh key: %v", err)
 	}
 
 	d.SetId(sshKey.Id)
```
yaronya, line 51: `sshKey` can be `nil` now, isn't it?

yaronya (commented):
Well done on asserting best practices 💪 
Left one question about `data_sshkey`

env0/data_sshkey.go
TomerHeber: yes! good catch!
I'll need to fix that.

TomerHeber: And another small one (wanted output to remain consistent):
https://github.com/env0/terraform-provider-env0/pull/306/commits/476871a0dbc418e34069d89bde0652b93fcdf7ad

env0/data_sshkey.go
```diff
@@ -31,23 +32,23 @@ func dataSshKey() *schema.Resource {
...
 	} else {
 		id, idSpecified := d.GetOk("id")
 		if !idSpecified {
 			return diag.Errorf("At lease one of 'id', 'name' must be specified")
 		}
 		sshKey, err = getSshKeyById(id, meta)
-		if err != nil {
-			return err
+		if err == nil && sshKey == nil {
+			err = fmt.Errorf("id %s not found", id)
```
yaronya, line 46: Can you please add some context to the error text? (`ssh key with id X not found`)

env0/data_sshkey.go
TomerHeber: @yaronya 
3cc1e9fe8c4e40f54c46a08b503e31c528ac4357

PR #308, opened 2022-03-30
Feat: add the api key resource

TomerHeber:
### Solution

Implemented the resource and added unit tests.
Part of #281


env0/resource_notification.go
```diff
@@ -170,6 +170,9 @@ func resourceNotificationImport(ctx context.Context, d *schema.ResourceData, met
 	if err != nil {
 		return nil, err
 	}
+	if notification == nil {
```
TomerHeber, line 173: unrelated... but it's bug/use-case I noticed I missed.

yaronya (approved):

PR #309, opened 2022-03-30
Feat: add git token resource

### LINKED ISSUE #148: Add support for git token resource
I would like to use github token_ids in my templates instead of just ssh keys.  There doesn't seem to be a data source to look up the token_id of a github token in env0 like there is for ssh keys.  Can that get added please?

TomerHeber:
resolves #148

### Solution
Last PR.
Added git token data.
Added acceptance tests.
Added integration tests.
Added documentation.


env0/data_git_token.go
```diff
@@ -0,0 +1,58 @@
...
+				Optional:     true,
+				ExactlyOneOf: []string{"name", "id"},
+			},
+			"id": {
+				Type:         schema.TypeString,
+				Description:  "id of the git token",
+				Optional:     true,
+				ExactlyOneOf: []string{"name", "id"},
+			},
+			"value": {
```
yaronya, line 28: If I'm not mistaken, API doesn't return this value anyway (or returns a masked value) so there is no value in having it in the data resource schema. 
The value of having this data resource in the first place is for users to query by name so they can re-use the git token's id in other templates, for example.

examples/data-sources/env0_git_token/data-source.tf
```diff
@@ -0,0 +1,12 @@
+resource "env0_git_token" "git_token_sample" {
+  name  = "name"
+  value = "value"
+}
+
+data "env0_git_token" "git_token_sample" {
```
yaronya, line 6: Can you please also add an example for fetching this data resource by name?

yaronya (commented):
Nicely done 👍 
Left a couple of comments

env0/data_git_token.go
TomerHeber: @yaronya 
078cc22530fccc8e51a7043f84eb7b6883d15a96

examples/data-sources/env0_git_token/data-source.tf
TomerHeber: @yaronya 
078cc22530fccc8e51a7043f84eb7b6883d15a96

yaronya (approved):

Issue #283, opened 2022-03-21
Add support for organization level policies


yaronya:
Should add an `env0_organization_policy` resource and data resource

eranelbaz:
Relevant API - https://developer.env0.com/docs/api/b3A6MTkyNjc1OTA-update-policy

default values should be the same as for new organization
Add description for TTL with the allowed values - you can get them from our API

PR #313, opened 2022-03-31
Feat: add support for organization level policies

TomerHeber:
### Solution
Added the required API.
Part of #283


client/organization.go
```diff
@@ -4,26 +4,40 @@ import (
 	"errors"
 )
 
-func (self *ApiClient) Organization() (Organization, error) {
+func (ac *ApiClient) Organization() (Organization, error) {
```
TomerHeber, line 7: gostaticcheck complains about "self".
I was working in this file, so changed it.

client/organization.go
```diff
@@ -4,26 +4,40 @@ import (
...
 	var result []Organization
-	err := self.http.Get("/organizations", nil, &result)
+	err := ac.http.Get("/organizations", nil, &result)
 	if err != nil {
 		return Organization{}, err
 	}
 	if len(result) != 1 {
-		return Organization{}, errors.New("Server responded with too many organizations")
+		return Organization{}, errors.New("server responded with too many organizations")
```
TomerHeber, line 14: errors should begin with lower case (another complaint in my IDE).

client/model.go
```diff
@@ -3,11 +3,27 @@ package client
...
+	CreatedBy                           string  `json:"createdBy"`
+	CreatedAt                           string  `json:"createdAt"`
+	UpdatedAt                           string  `json:"updatedAt"`
+	Role                                string  `json:"role"`
+	IsSelfHosted                        bool    `json:"isSelfHosted"`
+	IsSelfHostedK8s                     bool    `json:"isSelfHostedK8s"`
+}
+
+type OrganizationPolicyUpdatePayload struct {
+	MaxTtl                              *string `json:"maxTtl"`
```
TomerHeber, line 23: null is actually meaningful in this case (infinite).
*string without omitempty is on purpose. (These fields are "required").

liranfarage89 (approved):
Great work 🥇

sagilaufer1992 (commented):
Looks great. small request and small question. refer and I'll approve

client/organization.go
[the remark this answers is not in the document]
TomerHeber: That's the general naming convention in golang (in most cases it would even be one letter).
E.g.:
https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/terraform

But yeah, I can see how it causes a little bit of confusion.

(Also my IDE is complaining about the usage of "self").

PR #314, opened 2022-04-01
Chore: run acceptance tests in parallel

### LINKED ISSUE #96: Unit tests take a minute
On Mac :

```
ok  	github.com/env0/terraform-provider-env0/client/http	(cached)
=== RUN   TestUnitOrganizationData
--- PASS: TestUnitOrganizationData (16.59s)
=== RUN   TestUnitProjectData
--- PASS: TestUnitProjectData (16.04s)
=== RUN   TestProvider
--- PASS: TestProvider (0.00s)
=== RUN   TestMissingConfigurations
--- PASS: TestMissingConfigurations (0.00s)
=== RUN   TestUnitProjectResource
--- PASS: TestUnitProjectResource (27.87s)
PASS
ok  	github.com/env0/terraform-provider-env0/env0	61.365s
```

<div><br/>via <a href="https://giphy.com/gifs/aint-nobody-got-time-for-that-gif-bWM2eWYfN3r20">GIPHY</a></div>

TomerHeber:
resolves #96 

### Solution
Modified code to use ParallelTest instead of Test.
The changes unmasked some bugs. Fixed them as well.


env0/provider.go
```diff
@@ -13,6 +13,14 @@ import (
 
 func Provider(version string) plugin.ProviderFunc {
 	return func() *schema.Provider {
+		apiKeyEnv := "ENV0_API_KEY"
```
TomerHeber, line 16: This is required when running the provider unit test (which modified the env).
Unfortunately, there's is no good way to configure env variables when running tests in parallel.
E.g. the new t.SetEnv(...) method is not supported when running tests in parallel (maybe future golang versions will have better support for it).

env0/provider_test.go
```diff
@@ -3,55 +3,48 @@ package env0
...
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
-	"os"
-	"strings"
-	"testing"
 )
 
-var (
```
TomerHeber, line 18: These two vars should not be global but per test. (Otherwise mocking doesn't work properly).

env0/provider_test.go
```diff
@@ -3,55 +3,48 @@ package env0
...
-		provider := Provider("")()
-		provider.ConfigureContextFunc = func(ctx context.Context, d *schema.ResourceData) (interface{}, diag.Diagnostics) {
-			return apiClientMock, nil
-		}
-		return provider, nil
-	},
-}
-
 func runUnitTest(t *testing.T, testCase resource.TestCase, mockFunc func(mockFunc *client.MockApiClientInterface)) {
+	os.Setenv("TF_ACC", "1")
```
TomerHeber, line 20: `TF_ACC` is required when running terraform acceptance tests.
(Will error if not set).
In addition, this actually uncovered several bugs. When this env variable is set, some errors turn into panics.

env0/provider_test.go
```diff
@@ -3,55 +3,48 @@ package env0
...
+	os.Setenv("ENV0_API_KEY", "value")
```
TomerHeber, line 21: the secret and key env variables must not be unset. (Specifically when running tests in parallel).

env0/provider_test.go
```diff
@@ -3,55 +3,48 @@ package env0
...
+			provider := Provider("")()
+			provider.ConfigureContextFunc = func(ctx context.Context, d *schema.ResourceData) (interface{}, diag.Diagnostics) {
+				return apiClientMock, nil
+			}
+			return provider, nil
+		},
+	}
 	testCase.PreventPostDestroyRefresh = true
-	resource.UnitTest(&testReporter, testCase)
+	resource.ParallelTest(&testReporter, testCase)
```
TomerHeber, line 47: This makes it all run in Parallel.

env0/resource_azure_credentials_test.go
```diff
@@ -135,15 +135,18 @@ func TestUnitAzureCredentialsResource(t *testing.T) {
...
+			"client_secret",
+			"subscription_id",
+			"tenant_id",
+			"name",
 		}
-		for _, testCase := range missingArgumentsTestCases {
-			runUnitTest(t, testCase, func(mock *client.MockApiClientInterface) {
+
+		for _, argument := range arguments {
+			tc := missingArgumentTestCase(resourceType, resourceName, map[string]interface{}{}, argument)
```
TomerHeber, line 147: There are several places with similar fixes. Doing it in the old manner causes a parallel test inside a parallel, which is not allowed in Golang.

env0/resource_environment.go
```diff
@@ -185,10 +185,10 @@ func setEnvironmentSchema(d *schema.ResourceData, environment client.Environment
 	d.Set("id", environment.Id)
 	d.Set("name", environment.Name)
 	d.Set("project_id", environment.ProjectId)
-	d.Set("workspace", environment.WorkspaceName)
-	d.Set("auto_deploy_by_custom_glob", environment.AutoDeployByCustomGlob)
-	d.Set("ttl", environment.LifespanEndAt)
-	d.Set("terragrunt_working_directory", environment.TerragruntWorkingDirectory)
+	safeSet(d, "workspace", environment.WorkspaceName)
```
TomerHeber, line 188: Minor bug. These calls return errors (fields are not in the schema).
With TF_ACC it panics.

safeSet first checks if the field is in the schema.

env0/resource_environment.go
```diff
@@ -209,11 +209,17 @@ func setEnvironmentSchema(d *schema.ResourceData, environment client.Environment
 }
 
 func setEnvironmentConfigurationSchema(d *schema.ResourceData, configurationVariables []client.ConfigurationVariable) {
-	for index, configurationVariable := range configurationVariables {
+	var variables []interface{}
```
TomerHeber, line 212: I have no idea how this worked and how it had passed the acceptance tests.
Took my a while to figure out what was wrong and how to fix it.

(With TF_ACC it panics).

TomerHeber:
On my Mac:
75 seconds (after-parallel) vs 175 seconds (before-serialized).

In the pipeline, it seems more-or-less the same. Perhaps it needs a better CPU and more cores.

yaronya:
> In the pipeline, it seems more-or-less the same. Perhaps it needs a better CPU and more cores.

@TomerHeber that's because GH actions uses a single-core container 😞

env0/provider.go
yaronya: It'd be helpful if you can add some comment around that so any other person that reaches this piece of logic will understand what's going on

yaronya (approved):

PR #307, opened 2022-03-29
Feat cloud cost credentials - 218

samuel-br:
### Issue & Steps to Reproduce / Feature Request
[//]: <> (choose “resolves” “fixes” or “closes” and put link for the issue)
### Solution

- Add one schema for all the cost resources that uses the `ExactlyOneOf` and `ConflictsWith` to keep the validation and logic
- Add one file for tests that check all the cost resources and on the way the schema validation 

resolves #218

PR #317, opened 2022-04-03
Feat cloud cost credentials data - 218

samuel-br:
#218 
### Solution

- add data resource for each cloud cost credentials
- add UT for each cloud cost credentials

**part of solution for #218, along with #252 **


env0/data_aws_cost_credentials.go
```diff
@@ -0,0 +1,92 @@
...
+			return diag.Errorf("Either 'name' or 'id' must be specified")
+		}
+		credentials, err = getAwsCostCredentialsByName(name.(string), meta)
+		if err != nil {
+			return err
+		}
+	}
+
+	d.SetId(credentials.Id)
+	d.Set("name", credentials.Name)
```
yaronya, line 54: Please use the new common `writeResourceData` instead. Have a look at existing usages of it to get the idea.

Note: this comment is relevant to all other resources

env0/data_google_cost_credentials.go
```diff
@@ -0,0 +1,93 @@
...
+
+	credentialsByNameAndType := make([]client.Credentials, 0)
+	for _, candidate := range credentialsList {
+		if candidate.Name == name.(string) && isValidGoogleCostCredentialsType(candidate.Type) {
+			credentialsByNameAndType = append(credentialsByNameAndType, candidate)
+		}
+	}
+
+	if len(credentialsByNameAndType) > 1 {
+		return client.Credentials{}, diag.Errorf("Found multiple Google cost Credentials for name: %s", name)
```
yaronya, line 71: Please align this error message with all other error messages in this file...

yaronya (commented):
Looking good overall
Left a few comments

env0/data_aws_cost_credentials.go
samuel-br: Good point ill do it, just a little note, the function you mention based on the name of the field in schema against the field name in the struct , in aws creds we have 'arn' in schema but the struct field is '**RoleArn** so in some places like that we cant use those functions, or we can just change the name of the field in struct/schema

samuel-br: 4eec18f9488a26f0acf5fe31b46a44b23cc303cd

env0/data_google_cost_credentials.go
samuel-br: 4eec18f9488a26f0acf5fe31b46a44b23cc303cd

eranelbaz (changes requested):
Why for `data` we have 3 different functions for that?
the only difference is the `credentials.Type`
please DRY that

env0/data_aws_cost_credentials.go
yaronya: So how did you solve that arn -> RoleArn tranform?

eranelbaz:
@samuel-br let's put this PR on hold until we finish with #307 as I see similar typos here and there

Also, did you see 
> Why for `data` we have 3 different functions for that?
the only difference is the `credentials.Type`
please DRY that

env0/data_aws_cost_credentials.go
samuel-br: In this specific code there is no `arn` attribute so I do used the function you suggested , but in resource(no data resource) for example I cant use it, another problem in this function is when the value of the filed in the struct is struct itself like `value` in `awscredentialscreate` but for this I have a solution, anyway its not enough because the problem we talked above, unless we change the schema field name to `role_arn`

yaronya: Ok understood. 
Regarding your last sentence - I do think we should change it to `role_arn` so:
1. api and schema are aligned 
2. We can use this util

eranelbaz (approved):
LGTM

PR #320, opened 2022-04-04
Chore: remove model.go and split its contents to api files

### LINKED ISSUE #6: Split up models file
https://github.com/env0/terraform-provider-env0/blob/main/client/model.go is going to get bigger and bigger, should be split up per class, preferably to go along the API methods themselves.

TomerHeber:
resolves #6 

### Solution
Removed model.py.
All the content has been split into the API files.


TomerHeber:
It seems like there are many changes here. But what was done is just move struct from one file to others.
As long as unit tests are passing this should be fine.

client/configuration_variable_test.go
```diff
@@ -35,6 +37,33 @@ var _ = Describe("Configuration Variable", func() {
 		IsRequired:     &isRequired,
 	}
 
+	Describe("ConfigurationVariable", func() {
```
TomerHeber, line 40: This was taken from model_test.go

client/common_models.go
```diff
@@ -0,0 +1,13 @@
+package client
+
+type User struct {
```
TomerHeber, line 3: This is the only struct (I could find) that is common across many files.
Decided to create a common_models.go file for this use-case. (It shouldn't bloat as model.go did).

PR #322, opened 2022-04-06
Fix: Remove flaky template tests

RLRabinowitz:
The template tests in harness are flaky, since using `data` right after creating a resource is eventually consistent.
This isn't a real use case
### Solution
We already have other `data`s of templates that should exist as a prerequisite. Only use those in the template tests


tests/integration/004_template/expected_outputs.json
```diff
@@ -1,8 +1,9 @@
...
-    "tested2_template_name": "tested1-",
-    "tested2_template_repository": "https://github.com/env0/templates",
-    "tested1_template_repository": "https://gitlab.com/env0/gitlab-vcs-integration-tests.git",
-    "tested2_template_path": "second",
-    "tg_tg_version" : "0.35.0"
+    "tested1_template_type": "terraform",
+    "tested1_template_name": "tested1-",
+    "tested1_template_repository": "https://github.com/env0/templates",
+    "tested2_template_repository": "https://gitlab.com/env0/gitlab-vcs-integration-tests.git",
+    "tested1_template_path": "second",
```
RLRabinowitz, line 6: The `data`s were mixed up 🤦🏻 
`tested1` pointed at `tested2` and vice-versa

tests/integration/004_template/main.tf
```diff
@@ -67,52 +67,28 @@ resource "env0_configuration_variable" "in_a_template2" {
...
-  name       = "tested1-${random_string.random.result}"
-}
-data "env0_template" "tested1" {
-  depends_on = [time_sleep.wait_for_all_templates]
-  name       = "GitLab Test-${random_string.random.result}"
-}
-data "env0_template" "template_tg" {
-  depends_on = [time_sleep.wait_for_all_templates]
-  name       = "Template for environment resource - tg-${random_string.random.result}"
-}
```
RLRabinowitz, line 91: No need for those. We already test `data`s here

tests/integration/004_template/expected_outputs.json
```diff
@@ -1,8 +1,9 @@
...
+    "tg_tg_version" : "0.35.0",
+    "github_template_type": "terraform"
```
RLRabinowitz, line 8: This is the only output I've added

tests/integration/004_template/main.tf
```diff
@@ -67,52 +67,28 @@ resource "env0_configuration_variable" "in_a_template2" {
...
 
-# Temporary - Sleep to avoid eventual consistency issues with the data sources
-resource "time_sleep" "wait_for_all_templates" {
-  depends_on      = [env0_template.tested1, env0_template.tested2, env0_template.template_tg]
-  triggers = {
-      # Only using depends_on doesn't work on re-apply. This makes sure the sleep happens on re-apply
-      second_run = var.second_run
-    }
-  create_duration = "5s"
+output "tested1_template_id" {
```
avnerenv0, line 70: IMO this isn't used

tests/integration/004_template/expected_outputs.json
avnerenv0: Maybe change the names to something meaningful instead?
Will probably be less confusing than 1 and 2

avnerenv0 (approved):
minor comments

tests/integration/004_template/expected_outputs.json
sagilaufer1992: thank God (and you of course), it wasn't just in my head... because it passed before I believed it was intentional.

tests/integration/004_template/main.tf
RLRabinowitz: Yeah, I think this is only to make sure the data exists.
We can't really validate it

tests/integration/004_template/expected_outputs.json
RLRabinowitz: https://github.com/env0/terraform-provider-env0/pull/322/commits/70d3fe260af5383ccc86f38cbb04d3809a4bbb92

liranfarage89 (approved):

PR #326, opened 2022-04-07
Bug: environment resource configuration field is affected by global s…

### LINKED ISSUE #318: Environment resource `configuration` field is affected by global scope variables
Consider the following TF:
```
resource "env0_environment" "env" {
  name = "name"
  project_id = "my-project-id"
  template_id = "my-template-id"
  force_destroy = true
  configuration {
    name = "myvar"
    value = "value"
  }
}

resource "env0_configuration_variable" "global_variable" {
  name  = "anothervar"
  value = "anothervalue"
}
```

On first apply, this works as intended.
On a second apply, the `env0_environment`'s plan get's messed up, and affected by the global variable (variable defined at organization level). The plan would be something like this:

```
  # env0_environment.envy has been changed
  ~ resource "env0_environment" "envy" {
        id                               = "ID"
        name                         = "name"
        # (10 unchanged attributes hidden)

      ~ configuration {
          ~ name         = "myvar" -> "anothervar"
          ~ value         = "new" -> "anothervalue"
            # (4 unchanged attributes hidden)
        }
      + configuration {
          + name         = "myvar"
          + type           = "environment"
          + value          = "value"
        }
    }
```

The reason is that the `/configuration` API is getting both `organizationId` and `environmentId` in the query parameters, so it returns a merged cofiguration of both organization-level variables and environment-level variables

We should not send `organizationId` to that API when considering variables in `configuration` field in `env0_environment`

**Note:** the behaviour is the same for `data env0_configuration_variable` and should be changed there as well. There, we only allow for a single specific scope (`project_id`/`template_id`/...), and we merge the configuration with organization variables. Let's remove this configuration merge (as-in, let's stop putting `organizationId` in the request when scope is not `ScopeGlobal`)

TomerHeber:
Environment resource configuration field is affected by global scope variables

fixes #318 

### Solution
In the environment resource: filter out any configuration that is not within the environment scope.

### DIFF
--- a/client/configuration_variable.go
+++ b/client/configuration_variable.go
@@ -100,6 +100,18 @@ func (client *ApiClient) ConfigurationVariablesByScope(scope Scope, scopeId stri
 	if err != nil {
 		return []ConfigurationVariable{}, err
 	}
+
+	if scope != ScopeGlobal {
+		// Filter out global scopes. If a non global scope is requested.
+		filteredResult := []ConfigurationVariable{}
+		for _, variable := range result {
+			if variable.Scope != ScopeGlobal {
+				filteredResult = append(filteredResult, variable)
+			}
+		}
+		return filteredResult, nil
+	}
+
 	return result, nil
 }
 
--- a/client/configuration_variable_test.go
+++ b/client/configuration_variable_test.go
@@ -37,6 +37,22 @@ var _ = Describe("Configuration Variable", func() {
 		IsRequired:     &isRequired,
 	}
 
+	mockGlobalConfigurationVariable := ConfigurationVariable{
+		Id:             "config-var-id-789",
+		Name:           "configName",
+		Description:    "configDescription",
+		Value:          "configValue",
+		OrganizationId: organizationId,
+		IsSensitive:    &isSensitive,
+		Scope:          ScopeGlobal,
+		Type:           &varType,
+		ScopeId:        "project-123",
+		UserId:         "user|123",
+		Schema:         &schema,
+		IsReadonly:     &isReadonly,
+		IsRequired:     &isRequired,
+	}
+
 	Describe("ConfigurationVariable", func() {
 		Describe("Schema", func() {
 			It("On schema type is free text, enum should be nil", func() {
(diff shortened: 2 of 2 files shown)


env0/resource_environment.go
```diff
@@ -229,6 +229,10 @@ func setEnvironmentConfigurationSchema(d *schema.ResourceData, configurationVari
 	var variables []interface{}
 
 	for _, configurationVariable := range configurationVariables {
+		if configurationVariable.Scope != client.ScopeEnvironment {
```
TomerHeber, line 232: Decided to keep the API as-is.
IMHO it makes more sense for each "consumer" of the API to filter/add whatever it needs.

RLRabinowitz:
Couple of questions:

1. How did [this code change](https://github.com/env0/terraform-provider-env0/pull/314/files#diff-5dfeada10c7bdbcd4234860009b1a414cb4d99b87791eb4cb6ad6c3b7c3f703aR212) change the behaviour here? It's a bit unclear to me
2. Maybe instead of filtering out the scope, let the API only ask for variables in environment scope? As-in - have the API not send the `organizationId` on its request

RLRabinowitz:
We talked. I was wrong about the API, it must receive `organizationId`.
We're also not sure how this code used to work correctly in 0.2.31.

For now, we'll go with this approach

@TomerHeber as discussed, this only needs some tests

client/configuration_variable.go
```diff
@@ -100,6 +100,18 @@ func (client *ApiClient) ConfigurationVariablesByScope(scope Scope, scopeId stri
 	if err != nil {
 		return []ConfigurationVariable{}, err
 	}
+
+	if scope != ScopeGlobal {
```
TomerHeber, line 104: @RLRabinowitz - decided to fix the issue here. This covers much more use-cases.
I hope it makes sense: I filter out the global scope if the requested scope isn't global.

client/configuration_variable_test.go
```diff
@@ -247,7 +263,7 @@ var _ = Describe("Configuration Variable", func() {
 
 	Describe("ConfigurationVariablesByScope", func() {
 		var returnedVariables []ConfigurationVariable
-		mockVariables := []ConfigurationVariable{mockConfigurationVariable}
+		mockVariables := []ConfigurationVariable{mockConfigurationVariable, mockGlobalConfigurationVariable}
```
TomerHeber, line 266: As discussed. Added the use-case to the unit tests.

TomerHeber:
Note: I was unable to get the acceptance test working due to timeout.
See: https://env0.slack.com/archives/C01QBBEGBEH/p1649617036221799

env0/resource_environment.go
TomerHeber: no longer correct. Closing this conversation.

RLRabinowitz (approved):
Looks good 👍🏻

PR #330, opened 2022-04-12
Feat: add support for agent settings

### LINKED ISSUE #316: Add support for agent settings
Specifically for [assigning an agent to a project](https://developer.env0.com/docs/api/b3A6MzY2MjQwMjM-assign-agents-to-projects)
Supported resources:
1. `env0_agent_project_assignment` resource - should receive `project_id` and `agent_key`. Body is a map of project id to agent key FYI
2. `env0_agents` data resource - for receiving a list of all the agents

TomerHeber:
resolves #316 

### Solution
Added API call.
Added data-agents source.
Added unit tests.
Added documentation.


examples/resources/env0_agent_project_assignment/resource.tf
```diff
@@ -0,0 +1,11 @@
+resource "env0_project" "example" {
```
TomerHeber, line 1: Note/Question:

Is there a way to test this in harness?
I don't believe I have a way to create agents. Open to suggestions.

At the moment there are no added integration tests.

env0/data_agents.go
```diff
@@ -0,0 +1,54 @@
...
+	data := struct {
+		Agents []client.Agent
+	}{agents}
+
+	if err := writeResourceData(&data, d); err != nil {
+		return diag.Errorf("schema resource data serialization failed: %v", err)
+	}
+
+	// Not really needed. But required by Terraform SDK - https://github.com/hashicorp/terraform-plugin-sdk/issues/541
+	d.SetId("1")
```
yaronya, line 51: Does that mean that a user cannot have more than a single data resource of agents in his TF code?

examples/resources/env0_agent_project_assignment/resource.tf
yaronya: That wouldn't be ROI positive at the moment so I'd leave it without harness tests

yaronya (commented):
Left one comment

env0/data_agents.go
TomerHeber: @yaronya ,

I can make it random if preferred.
However, why would using "1" as the ID prevent it from being used in multiple places? 

See this bug fix in the google provider:
https://github.com/hashicorp/terraform-provider-google/pull/7375/files

yaronya: > However, why would using "1" as the ID prevent it from being used in multiple places?

I think that the id has to be unique across the TF state, isn't it? 
Random is also not so great, but preferable than `"1"`.
I'd try and see if we can get from the schema/context the resource name and set it based on it (since it has to be unique anyways on your TF stack).
WDYT?

> See this bug fix in the google provider

The example you just sent has the id based on some dynamic parameters such as project id and zone id

TomerHeber: @yaronya - regardless wouldn't it potentially cause duplication of ID (the example I sent).

yaronya: > @yaronya - regardless wouldn't it potentially cause duplication of ID (the example I sent).

@TomerHeber 
Yes, I believe it may create such duplication, but that example has somewhat lower chance to do so as it depends on **at least some** parameters
I'm assuming it'd break stuff as the id has to be unique, but maybe I'm wrong.

yaronya (approved):

Issue #324, opened 2022-04-07
Add proper release drafts 


yaronya:
We currently have an automatic go releaser (that builds and ships the provider to the TF provider registry) but our releases aren't descriptive. 
We need to have some sort of descriptive changelog for our releases that will comply with the best practices of TF provider and golang repos

PR #334, opened 2022-04-13
Chore: add proper release drafts

TomerHeber:
### Solution

Verifies that the PR title has Chore/Fix/Feat.

Part of #324.


.github/workflows/pr-commit-message-check.yml
```diff
@@ -0,0 +1,19 @@
+name: 'Commit Message Check'
```
yaronya, line 1: So eventually it looks like this on GH UI:
image

I'd suggest to rename it in a way that workflow name would be `Validations` or similar and the actual job name would be `PR Title`

TomerHeber: @yaronya is this what you meant? (see the PR checks).

yaronya: @TomerHeber I can't see it in the status checks now...

TomerHeber: @yaronya - you just need to change the title.
In any case, just did it. So you should see it now.

yaronya (dismissed):

.github/workflows/pr-commit-message-check.yml
yaronya: @TomerHeber the name of the status check can't be dynamic since this way you cannot have it as required for the PR to pass.
I'd suggest to have it as `Validations / PR Title` since we might add more validations in the future

TomerHeber: @yaronya - modified.

yaronya: Thanks. 
It seems to fail now 🤔
Can you validate it please?

TomerHeber: @yaronya - where do you see a failure?

yaronya: IDK, but I don't see it now so 👍

yaronya (approved):

.github/workflows/pr-commit-message-check.yml
TomerHeber: You probably saw it when I was testing (:

PR #337, opened 2022-04-14
Fix: changing variable scope

### LINKED ISSUE #332: Bug report : Changing variable scope
> I had a variable set to a template scope, then changed it to global, and it didn't show up in the global variables section. I ran terraform taint on it and it showed up and worked as expected, but the modify by itself didn't do that for some reason.

TomerHeber:
fixes #332 

### Solution
Added ForceNew to project_id, template_id, and environment_id.


sabasabo (approved):
LGTM

env0/resource_configuration_variable.go
```diff
@@ -51,18 +51,21 @@ func resourceConfigurationVariable() *schema.Resource {
...
 				Optional:      true,
 				ConflictsWith: []string{"project_id", "environment_id"},
+				ForceNew:      true,
 			},
 			"environment_id": {
 				Type:          schema.TypeString,
 				Description:   "create the variable under this environment, not globally",
 				Optional:      true,
 				ConflictsWith: []string{"template_id", "project_id", "is_required", "is_read_only"},
+				ForceNew:      true,
```
yaronya, line 68: Please also apply this to the `type` attribute of the schema. API doesn't support changing an existing variable's type

TomerHeber: added.

PR #338, opened 2022-04-14
Feat: upgrade to go version 1.18

### LINKED ISSUE #336: Upgrade to go version 1.18

TomerHeber:
resolves #336 

### Solution
Located all the 1.16 locations and modified them to 1.18.
Ran: `go mod tidy`.
(Unrelated) Formatted some terraform file (`terraform fmt`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ env:
   ENV0_API_ENDPOINT: ${{ secrets.ENV0_API_ENDPOINT }}
   ENV0_API_KEY: ${{ secrets.TF_PROVIDER_INTEGRATION_TEST_API_KEY }} # API Key for organization 'TF-provider-integration-tests' @ dev
   ENV0_API_SECRET: ${{ secrets.TF_PROVIDER_INTEGRATION_TEST_API_SECRET }}
-  GO_VERSION: 1.16
+  GO_VERSION: 1.18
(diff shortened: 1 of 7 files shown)


TomerHeber:
@yaronya - there's a comment in the code `// please change also in repo secrets`. Do you know what's it about?

yaronya:
> @yaronya - there's a comment in the code `// please change also in repo secrets`. Do you know what's it about?

@TomerHeber I think we used to fetch the version as a secret from GH (since that's the only way to share a common value between multiple GH actions) but we changed it back and only comment was left. You can ignore it probably.

.github/workflows/ci.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       - name: Go staticcheck
         uses: dominikh/staticcheck-action@v1.2.0
         with:
-          version: "2021.1.2"
+          version: "2022.1"
```
TomerHeber, line 39: Go 1.18 requires a newer staticcheck.

razbensimon (approved):
LGTM

razbensimon:
> > @yaronya - there's a comment in the code `// please change also in repo secrets`. Do you know what's it about?
> 
> @TomerHeber I think we used to fetch the version as a secret from GH (since that's the only way to share a common value between multiple GH actions) but we changed it back and only comment was left. You can ignore it probably.

So maybe let's remove that comment? @yaronya 
@TomerHeber

PR #343, opened 2022-04-24
Feat: add data source for teams

### LINKED ISSUE #340: Add data source for teams
Add a new data source to list all teams in the org.
This should provide a list of all the teams names in a given Org.

This list should be able to be used to create a team data source based on the names returned.

Usage example:

data "env0_teams" "all_teams" {
    organization = "1234"                       # The organization is normally part of the API key, just placed this here as an example
    search          = "Some search text "  # This is not part of the current ticket requirement, but an example of how this can be used in the future....
}

data "env0_team" "teams" {
    for_each = tomap(data.env0_teams.all_teams.names)
    name      = each.value
}

Relevant API:
  - https://developer.env0.com/docs/api/b3A6MTkyNjc2MDg-list-teams

TomerHeber:
resolves #340 

### Solution
Added the new data source.
Added unit tests.
Added integration tests.
Added examples.

### DIFF
--- a/env0/data_teams.go
+++ b/env0/data_teams.go
@@ -42,7 +42,7 @@ func dataTeamsRead(ctx context.Context, d *schema.ResourceData, meta interface{}
 	d.Set("names", data)
 
 	// Not really needed. But required by Terraform SDK - https://github.com/hashicorp/terraform-plugin-sdk/issues/541
-	d.SetId("1")
+	d.SetId("all_teams_names")
 
 	return nil
 }
--- a/env0/data_teams.go
+++ b/env0/data_teams.go
@@ -1,0 +1,48 @@
+package env0
+
+import (
+	"context"
+
+	"github.com/env0/terraform-provider-env0/client"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataTeams() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataTeamsRead,
+
+		Schema: map[string]*schema.Schema{
+			"names": {
+				Type:        schema.TypeList,
+				Description: "list of all teams (by name)",
+				Computed:    true,
+				Elem: &schema.Schema{
+					Type:        schema.TypeString,
+					Description: "the team name",
+				},
+			},
+		},
+	}
+}
+
+func dataTeamsRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	apiClient := meta.(client.ApiClientInterface)
+	teams, err := apiClient.Teams()
+	if err != nil {
+		return diag.Errorf("Could not get teams: %v", err)
+	}
+
+	data := []string{}
+
+	for _, team := range teams {
+		data = append(data, team.Name)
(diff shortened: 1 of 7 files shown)


env0/data_teams.go
```diff
@@ -0,0 +1,48 @@
...
+func dataTeamsRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	apiClient := meta.(client.ApiClientInterface)
+	teams, err := apiClient.Teams()
+	if err != nil {
+		return diag.Errorf("Could not get teams: %v", err)
+	}
+
+	data := []string{}
+
+	for _, team := range teams {
```
TomerHeber, line 38: I don't test if there are duplicate names. Wasn't sure it will be useful.

env0/data_teams.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataTeams() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataTeamsRead,
+
+		Schema: map[string]*schema.Schema{
+			"names": {
```
TomerHeber, line 16: could have added "ids" as well. wasn't sure it will be useful.

tests/integration/021_teams/main.tf
```diff
@@ -0,0 +1,22 @@
...
+  name = "team1"
+}
+
+resource "env0_team" "team_resource2" {
+  name = "team2"
+}
+
+data "env0_teams" "all_teams" {}
+
+data "env0_team" "teams" {
```
TomerHeber, line 11: See this integration test to understand how to use it. Let me know if something here doesn't seem right.

env0/data_teams.go
GiliFaroEnv0: not sure it will help, but we can always add it in the future

GiliFaroEnv0: not sure it will help, but we can always add it in the future

env0/data_teams.go
GiliFaroEnv0: Don't think it's relevant when reading from server

env0/data_teams.go
```diff
@@ -0,0 +1,48 @@
...
+	data := []string{}
+
+	for _, team := range teams {
+		data = append(data, team.Name)
+	}
+
+	d.Set("names", data)
+
+	// Not really needed. But required by Terraform SDK - https://github.com/hashicorp/terraform-plugin-sdk/issues/541
+	d.SetId("1")
```
GiliFaroEnv0, line 45: (I don't remember if must be a number,)
but if it is a must since it used in some "complex" behind the scenes try to make it unique so it be easy to identify 
```suggestion
	d.SetId("all_teams_names")
```

tests/integration/021_teams/main.tf
GiliFaroEnv0: LGTM

GiliFaroEnv0 (commented):
GREAT JOB!
Left a minor comment

GiliFaroEnv0 (approved):
LGTM!
as usual great job!

PR #345, opened 2022-04-27
Feat: add data source for templates

### LINKED ISSUE #341: Add data source for templates
Add a new data source to list all templates in the org.
This should provide a list of all the templates names in a given Org.

This list should be able to be used to create a template data source based on the names returned.

Usage example:

data "env0_templates" "all_templates" {
organization = "1234" # The organization is normally part of the API key, just placed this here as an example
search = "Some search text " # This is not part of the current ticket requirement, but an example of how this can be used in the future....
}

data "env0_template" "templates" {
for_each = tomap(data.env0_templates.all_templates.names)
name = each.value
}

Relevant API:
  - https://developer.env0.com/docs/api/b3A6NTg5Mjk3MQ-list-templates

TomerHeber:
resolves #341 

### Solution
Added the new data source.
Added unit tests.
Added integration tests.
Added examples.

### DIFF
--- a/tests/integration/022_templates/main.tf
+++ b/tests/integration/022_templates/main.tf
@@ -1,9 +1,17 @@
+provider "random" {}
+
+resource "random_string" "random" {
+  length    = 8
+  special   = false
+  min_lower = 8
+}
+
 data "env0_template" "github_template" {
   name = "Github Integrated Template"
 }
 
 resource "env0_template" "github_template1" {
-  name                                    = "Github Test-111"
+  name                                    = "Github Test ${random_string.random.result}-1"
   description                             = "Template description - GitHub"
   type                                    = "terraform"
   repository                              = data.env0_template.github_template.repository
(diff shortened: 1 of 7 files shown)


tests/integration/022_templates/main.tf
```diff
@@ -0,0 +1,44 @@
+data "env0_template" "github_template" {
+  name = "Github Integrated Template"
+}
+
+resource "env0_template" "github_template1" {
+  name                                    = "Github Test-111"
```
yaronya, line 6: Please make sure to set the name randomly (like we do in order tests) - relevant for below resources as well.
Otherwise the integration tests would fail on consecutive runs

TomerHeber: good point.... /:
I guess I'll remove the output then. Can't it have it both ways I guess....

yaronya (approved):

PR #356, opened 2022-05-01
Fix: support fetching environments with the limit+offset feature

### LINKED ISSUE #342: Support fetching environments with the limit+offset feature
Recently the api for getting environments added `limit` and `offset` to the query string parameters. The api also is limited by 100 results per query for environments. 
The provider should support the new parameters and allow fetching all the relevant environments for a user.
[https://developer.env0.com/docs/api/b3A6MTkzNDg1MjA-list-environments](relevant api doc)

TomerHeber:
fixes #342

### Solution

Extracts environments in a for loop until there are fewer elements than the limit (100).
Uses the new golang generics feature.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,7 +12,6 @@ repos:
     hooks:
       - id: go-fmt
       - id: go-vet
-      - id: go-imports
       - id: go-mod-tidy
   - repo: https://github.com/TekWizely/pre-commit-golang
     rev: v1.0.0-beta.5
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,7 @@ repos:
     hooks:
       - id: go-fmt
       - id: go-vet
+      - id: go-imports
       - id: go-mod-tidy
   - repo: https://github.com/TekWizely/pre-commit-golang
     rev: v1.0.0-beta.5
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -21,6 +21,7 @@ var (
 	mockHttpClient     *http.MockHttpClientInterface
 	apiClient          ApiClientInterface
 	httpCall           *gomock.Call
+	httpCall2          *gomock.Call
 	organizationIdCall *gomock.Call
 )
 
--- a/client/environment.go
+++ b/client/environment.go
@@ -92,24 +92,16 @@ type EnvironmentDeployResponse struct {
 	Id string `json:"id"`
 }
 
+func (Environment) getEndpoint() string {
+	return "/environments"
+}
+
 func (client *ApiClient) Environments() ([]Environment, error) {
-	var result []Environment
-	err := client.http.Get("/environments", nil, &result)
-	if err != nil {
-		return []Environment{}, err
-	}
-	return result, nil
+	return getAll(client, nil)
 }
 
 func (client *ApiClient) ProjectEnvironments(projectId string) ([]Environment, error) {
-
-	var result []Environment
-	err := client.http.Get("/environments", map[string]string{"projectId": projectId}, &result)
-
-	if err != nil {
-		return []Environment{}, err
-	}
-	return result, nil
+	return getAll(client, map[string]string{"projectId": projectId})
 }
 
 func (client *ApiClient) Environment(id string) (Environment, error) {
--- a/client/environment_test.go
+++ b/client/environment_test.go
@@ -10,6 +10,9 @@ import (
 	. "github.com/onsi/gomega"
 )
 
+const full_page = 100
+const partial_page = 33
+
 var _ = Describe("Environment Client", func() {
 	const (
 		environmentId = "env-id"
@@ -47,11 +50,11 @@ var _ = Describe("Environment Client", func() {
 
 		Describe("SuccessMultiPages", func() {
 			var environmentsP1, environmentsP2 []Environment
-			for i := 0; i < 100; i++ {
+			for i := 0; i < full_page; i++ {
 				environmentsP1 = append(environmentsP1, mockEnvironment)
 			}
 
-			for i := 0; i < 33; i++ {
+			for i := 0; i < partial_page; i++ {
 				environmentsP2 = append(environmentsP2, mockEnvironment)
 			}
 
@@ -85,11 +88,11 @@ var _ = Describe("Environment Client", func() {
 		Describe("SuccessMultiPagesWithProject", func() {
 			projectId := "proj123"
 			var environmentsP1, environmentsP2 []Environment
-			for i := 0; i < 100; i++ {
+			for i := 0; i < full_page; i++ {
 				environmentsP1 = append(environmentsP1, mockEnvironment)
 			}
 
-			for i := 0; i < 33; i++ {
+			for i := 0; i < partial_page; i++ {
 				environmentsP2 = append(environmentsP2, mockEnvironment)
 			}
 
--- a/client/environment_test.go
+++ b/client/environment_test.go
@@ -28,7 +28,7 @@ var _ = Describe("Environment Client", func() {
 		Describe("Success", func() {
 			BeforeEach(func() {
 				httpCall = mockHttpClient.EXPECT().
-					Get("/environments", nil, gomock.Any()).
+					Get("/environments", gomock.Any(), gomock.Any()).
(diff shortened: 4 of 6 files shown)


.pre-commit-config.yaml
```diff
@@ -12,7 +12,6 @@ repos:
     hooks:
       - id: go-fmt
       - id: go-vet
-      - id: go-imports
```
TomerHeber, line 15: go-imports does not fully support go 1.18.
removing for now. Go fmt is good enough.

client/client_test.go
```diff
@@ -21,6 +21,7 @@ var (
 	mockHttpClient     *http.MockHttpClientInterface
 	apiClient          ApiClientInterface
 	httpCall           *gomock.Call
+	httpCall2          *gomock.Call
```
TomerHeber, line 24: added this to test the new use-case (which is two API calls).

client/environment.go
```diff
@@ -92,24 +92,16 @@ type EnvironmentDeployResponse struct {
 	Id string `json:"id"`
 }
 
+func (Environment) getEndpoint() string {
```
TomerHeber, line 95: required by the interface (see pagination.go).

client/pagination.go
```diff
@@ -0,0 +1,60 @@
+package client
```
TomerHeber, line 1: This uses golang generics.
Please reach out if comments are missing or something is unclear.

With generics, we can easily expand it in the future to the rest of the API calls.

client/pagination.go
```diff
@@ -0,0 +1,60 @@
...
+
+import "strconv"
+
+const limit = 100
+
+type Pagination struct {
+	offset int
+	params map[string]string
```
TomerHeber, line 9: extra params to be passed each time. (E.g. "projectId": "the-project-id").

client/pagination.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+type Paginated interface {
+	Environment
```
TomerHeber, line 13: Right now the generics is just for Environment but can be expanded in the future for other types.

client/pagination.go
```diff
@@ -0,0 +1,60 @@
...
+	return map[string]string{
+		"limit":  strconv.Itoa(limit),
+		"offset": strconv.Itoa(p.offset),
+	}
+}
+
+// returns true if there is more data.
+func (p *Pagination) next(currentPageSize int) bool {
+	p.offset += currentPageSize
+	return currentPageSize == limit
```
TomerHeber, line 27: Assume there is more data if the current page is the length of limit.

client/pagination.go
```diff
@@ -0,0 +1,60 @@
...
+	p := Pagination{
+		offset: 0,
+		params: params,
+	}
+
+	var allResults []P
+
+	for {
+		pageParams := p.getParams()
+		for k, v := range params {
```
TomerHeber, line 41: merge the offset, limit params with the other params (may be nil if there are no other params).

client/pagination.go
```diff
@@ -0,0 +1,60 @@
...
+			pageParams[k] = v
+		}
+
+		var pageResults []P
+
+		err := client.http.Get(P{}.getEndpoint(), pageParams, &pageResults)
```
TomerHeber, line 47: hahaha can't believe `P{}.getEndpoint()` works.

client/pagination.go
```diff
@@ -0,0 +1,60 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		allResults = append(allResults, pageResults...)
```
TomerHeber, line 52: accumulate the results.
Let's hope it never reaches a stage where it's OOM.

client/pagination.go
TomerHeber: correction. I had to comment out generics due to harness using go 1.17.

client/pagination.go
```diff
@@ -0,0 +1,67 @@
...
+import "strconv"
+
+const limit = 100
+
+type Pagination struct {
+	offset int
+	params map[string]string
+}
+
+// Note: all the FIXME should be uncommented when Alpine adds support for go 1.18.
```
TomerHeber, line 12: Harnes uses terraform docker.
Alpine 3.15 (which terraform docker uses) supports golang 3.17.
Will have to wait for Alpine 3.16 which runs golang 3.18.

Once this happens we can switch back to generics (which is a new feature in golang 3.18).

Note: the env0-terraform library builds via go1.18, the issue is strictly related to integration tests (Harnes).

client/pagination.go
```diff
@@ -0,0 +1,67 @@
...
+// This will enable generics.
+
+type Paginated interface {
+	// FIXME: Environment
```
TomerHeber, line 16: Once generic is used, we can add all the "elements" that support pagination.
E.g. `Environment | Module | Project`

TomerHeber:
Note: initially was written with go generics. However, Harnes is still using go1.17. (More details in the comments).

client/pagination.go
```diff
@@ -0,0 +1,67 @@
...
+	getEndpoint() string
+}
+
+func (p *Pagination) getParams() map[string]string {
+	return map[string]string{
+		"limit":  strconv.Itoa(limit),
```
sagilaufer1992, line 22: if no limit is passed in the params, what will happen? we didn't define default  a limit for the provider, didn't see one defined in this PR

client/environment_test.go
```diff
@@ -45,11 +45,88 @@ var _ = Describe("Environment Client", func() {
...
+
+			It("Should return the environments", func() {
+				Expect(environments).To(Equal(append(environmentsP1, environmentsP2...)))
+			})
+		})
+
+		Describe("SuccessMultiPagesWithProject", func() {
+			projectId := "proj123"
+			var environmentsP1, environmentsP2 []Environment
+			for i := 0; i < 100; i++ {
```
sagilaufer1992, line 88: 100, 33 - magic numbers. please name those.

client/environment_test.go
```diff
@@ -45,11 +45,88 @@ var _ = Describe("Environment Client", func() {
...
+
+				environments, err = apiClient.Environments()
+			})
+
+			It("Should return the environments", func() {
+				Expect(environments).To(Equal(append(environmentsP1, environmentsP2...)))
+			})
+		})
+
+		Describe("SuccessMultiPagesWithProject", func() {
```
sagilaufer1992, line 85: should have one with organization as well, and one that should FAIL when both are passed

client/environment_test.go
```diff
@@ -45,11 +45,88 @@ var _ = Describe("Environment Client", func() {
 			})
 		})
 
+		Describe("SuccessMultiPages", func() {
```
sagilaufer1992, line 48: without project or organization the http call should fail

sagilaufer1992 (commented):
looks good, left some comments

client/pagination.go
TomerHeber: If no limit is passed it will use the default as defined by the API (in the website) (though currently missing from the website due to an issue). Regardless, it's better to be explicit and always pass a limit (in case the default changes in the backend). 

Sorry. I'm not sure I understand.
What do you mean to define a limit? where?

There's a `const limit = 100` in the code if that's what you meant...

client/environment_test.go
TomerHeber: It doesn't fail (this applies to the code before my changes).
I guess it uses organization by default if nothing is passed.

client/environment_test.go
TomerHeber: This is not something that is enforced in the code. Therefore, not really something we would want to test.
The backend API will reject it.

client/environment_test.go
TomerHeber: 1c3546254247fe7a2076a442fe3eb759f09bd16f

client/pagination.go
sagilaufer1992: missed it, my bad

sagilaufer1992 (approved):

client/pagination.go
TomerHeber: @sagilaufer1992 - no worries. I just merged the code.
Regardless, if you think there are things that need to be improved (in general for environments), let me know and I will open a new issue.

PR #361, opened 2022-05-04
Feat: add data source user

### LINKED ISSUE #298: Add data source for user
Add `data env0_user`
the query will be by email - and return user details from https://developer.env0.com/docs/api/b3A6MTkyNjc1OTE-list-users

need to expose only user id (can be by `id`)

TomerHeber:
closes #298 

### Solution

Added the required data source, client API, unit tests and documentation.
Note: I tested integration manually. Since the user is not known in advance, I decided not to add an integration test for it.
Note2: as requested - I'm exposing just the user id, if more fields are required please let me know.

### DIFF
--- a/client/user_test.go
+++ b/client/user_test.go
@@ -1,0 +1,57 @@
+package client_test
+
+import (
+	"errors"
+
+	. "github.com/env0/terraform-provider-env0/client"
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Module Client", func() {
+	mockUser := OrganizationUser{
+		User: User{
+			Email:  "a@b.com",
+			UserId: "1",
+		},
+	}
+
+	Describe("Users", func() {
+		var users []OrganizationUser
+		mockUsers := []OrganizationUser{mockUser}
+		var err error
+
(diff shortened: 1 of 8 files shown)


env0/data_user.go
```diff
@@ -0,0 +1,64 @@
...
+	return &schema.Resource{
+		ReadContext: dataUserRead,
+
+		Schema: map[string]*schema.Schema{
+			"email": {
+				Type:        schema.TypeString,
+				Description: "the email of the user",
+				Required:    true,
+			},
+			"id": {
```
TomerHeber, line 21: As requested. Just "id" is exposed.

client/user_test.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"errors"
+
+	. "github.com/env0/terraform-provider-env0/client"
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Module Client", func() {
```
yaronya, line 12: `Module Client` ? 🤔

yaronya (approved):

client/user_test.go
TomerHeber: ooops! fixing!

PR #369, opened 2022-05-09
Fix: env0_cost_credentials_project_assignment not working

### LINKED ISSUE #366: env0_cost_credentials_project_assignment Not Working - "Bad Request: /body must be object"
When trying to assign Cost Credentials to a Project, I get the following error:
`Error: could not assign cost credentials to project: 400 Bad Request: /body must be object`

My code is:

```
resource "env0_aws_cost_credentials" "cost" {
  name        = "${var.account_name}-cost"
  arn         = aws_iam_role.env0_cost.arn
  external_id = "$${ssm:env0/${var.org_name}/cost-external-id}"
}
resource "env0_cost_credentials_project_assignment" "cost" {
  credential_id = env0_aws_cost_credentials.cost.id
  project_id    = env0_project.main.id
}
```

Provider version `1.0.2` in use.
Have tested with both TF version `0.13.7` and `1.0.8` just to rule it out.

I'm probably wrong but looks like the issue might be here:
https://github.com/env0/terraform-provider-env0/blob/ef9800143cba13c7a1dccbd0f79da7a529279ab3/client/cost_credentials_project_assignment.go#L9

According to the API, it expects the body as a json object:

```
'{
  "credentialsId": "string"
}'
```

If I am not mistaken, looks like it is just be being passed in as a string of the ID itself but my GoLang is not that strong so I could be well off.

TomerHeber:
fixes #366

### Solution
The body argument was incorrect.
The URL path was incorrect.
Added an integration test.


client/cost_credentials_project_assignment.go
```diff
@@ -22,7 +24,7 @@ func (client *ApiClient) RemoveCostCredentialsFromProject(projectId string, cred
 
 func (client *ApiClient) CostCredentialIdsInProject(projectId string) ([]CostCredentialProjectAssignment, error) {
 	var result []CostCredentialProjectAssignment
-	err := client.http.Get("/costs/project/"+projectId, nil, &result)
+	err := client.http.Get("/costs/project/"+projectId+"/credentials", nil, &result)
```
TomerHeber, line 27: url path was incorrect.

PR #371, opened 2022-05-10
Fix: env0_aws_credentials - apply after import will delete+create

### LINKED ISSUE #99: env0_aws_credentials - apply after import will delete+create
1. Create a Cloud Credential
2. Create the TF config for it
3. Do a `terraform import` on it
4. Apply- you will be prompted that the credential will be replaced

This is probably because we don't get the full state of the credential on the import.

TomerHeber:
I'm keeping open, to adding support for other credential types.

### Solution
Added import.
Removed a schema requirements and instead enforce during read.
Added acceptance tests.
Added a "resource" tag.
Added import examples.
Fixed some other issues I noticed.

### DIFF
--- a/client/cloud_credentials.go
+++ b/client/cloud_credentials.go
@@ -33,7 +33,7 @@ type AwsCredentialsCreatePayload struct {
 }
 
 type AwsCredentialsValuePayload struct {
-	RoleArn         string `json:"roleArn" resource:"arn"`
+	RoleArn         string `json:"roleArn" tfschema:"arn"`
 	ExternalId      string `json:"externalId"`
 	AccessKeyId     string `json:"accessKeyId"`
 	SecretAccessKey string `json:"secretAccessKey"`
--- a/client/cloud_credentials.go
+++ b/client/cloud_credentials.go
@@ -1,9 +1,5 @@
 package client
 
-import (
-	"fmt"
-)
-
 type AwsCredentialsType string
 type GcpCredentialsType string
 type AzureCredentialsType string
@@ -37,7 +33,7 @@ type AwsCredentialsCreatePayload struct {
 }
 
 type AwsCredentialsValuePayload struct {
-	RoleArn         string `json:"roleArn"`
+	RoleArn         string `json:"roleArn" resource:"arn"`
 	ExternalId      string `json:"externalId"`
 	AccessKeyId     string `json:"accessKeyId"`
 	SecretAccessKey string `json:"secretAccessKey"`
@@ -89,7 +85,7 @@ func (client *ApiClient) CloudCredentials(id string) (Credentials, error) {
 		}
 	}
 
-	return Credentials{}, fmt.Errorf("CloudCredentials: [%s] not found ", id)
+	return Credentials{}, &NotFoundError{}
 }
 
 func (client *ApiClient) CloudCredentialsList() ([]Credentials, error) {
--- a/env0/credentials.go
+++ b/env0/credentials.go
@@ -1,0 +1,47 @@
+package env0
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/env0/terraform-provider-env0/client"
+	"github.com/google/uuid"
+)
+
+func getCredentialsByName(name string, prefix string, meta interface{}) (client.Credentials, error) {
+	apiClient := meta.(client.ApiClientInterface)
+
+	credentialsList, err := apiClient.CloudCredentialsList()
+	if err != nil {
+		return client.Credentials{}, err
+	}
+
+	var foundCredentials []client.Credentials
+	for _, credentials := range credentialsList {
+		if credentials.Name == name && strings.HasPrefix(credentials.Type, prefix) {
+			foundCredentials = append(foundCredentials, credentials)
+		}
+	}
+
+	if len(foundCredentials) == 0 {
+		return client.Credentials{}, fmt.Errorf("credentials with name %v not found", name)
+	}
+
+	if len(foundCredentials) > 1 {
+		return client.Credentials{}, fmt.Errorf("found multiple credentials with name: %s. Use id instead or make sure credential names are unique %v", name, foundCredentials)
+	}
+
+	return foundCredentials[0], nil
+}
+
+func getCredentials(id string, prefix string, meta interface{}) (client.Credentials, error) {
+	_, err := uuid.Parse(id)
+	if err == nil {
+		log.Println("[INFO] Resolving credentials by id: ", id)
+		return meta.(client.ApiClientInterface).CloudCredentials(id)
+	} else {
+		log.Println("[INFO] Resolving credentials by name: ", id)
+		return getCredentialsByName(id, prefix, meta)
+	}
+}
--- a/env0/data_notification.go
+++ b/env0/data_notification.go
@@ -61,7 +61,7 @@ func dataNotificationRead(ctx context.Context, d *schema.ResourceData, meta inte
 	}
 
 	if err := writeResourceData(notification, d); err != nil {
-		diag.Errorf("schema resource data serialization failed: %v", err)
+		return diag.Errorf("schema resource data serialization failed: %v", err)
 	}
 
 	return nil
--- a/env0/resource_aws_credentials.go
+++ b/env0/resource_aws_credentials.go
@@ -2,6 +2,7 @@ package env0
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/env0/terraform-provider-env0/client"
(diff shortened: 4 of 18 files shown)


client/cloud_credentials.go
```diff
@@ -37,7 +33,7 @@ type AwsCredentialsCreatePayload struct {
 }
 
 type AwsCredentialsValuePayload struct {
-	RoleArn         string `json:"roleArn"`
+	RoleArn         string `json:"roleArn" resource:"arn"`
```
TomerHeber, line 36: new tag type for serialization. See usage below.
(Basically in terraform the resource is "arn" and not "role_arn" - so needed to resolve this).

client/cloud_credentials.go
```diff
@@ -89,7 +85,7 @@ func (client *ApiClient) CloudCredentials(id string) (Credentials, error) {
 		}
 	}
 
-	return Credentials{}, fmt.Errorf("CloudCredentials: [%s] not found ", id)
+	return Credentials{}, &NotFoundError{}
```
TomerHeber, line 88: Used to handle drift later.

env0/credentials.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/env0/terraform-provider-env0/client"
+	"github.com/google/uuid"
+)
+
+func getCredentialsByName(name string, prefix string, meta interface{}) (client.Credentials, error) {
```
TomerHeber, line 12: Common helper functions for credentials (can be used by GCP and Azure in the future). (In case import is added to them as well).

env0/data_notification.go
```diff
@@ -61,7 +61,7 @@ func dataNotificationRead(ctx context.Context, d *schema.ResourceData, meta inte
 	}
 
 	if err := writeResourceData(notification, d); err != nil {
-		diag.Errorf("schema resource data serialization failed: %v", err)
+		return diag.Errorf("schema resource data serialization failed: %v", err)
```
TomerHeber, line 64: bug.

TomerHeber: there are a few places with the same bug...

env0/resource_aws_credentials.go
```diff
@@ -27,7 +30,6 @@ func resourceAwsCredentials() *schema.Resource {
 				Optional:      true,
 				ForceNew:      true,
 				ConflictsWith: []string{"access_key_id"},
-				ExactlyOneOf:  []string{"access_key_id"},
```
TomerHeber, line 30: The "ExactlyOneOf" cannot be inforced in the schema level. (Will fail on import).

env0/resource_aws_credentials.go
```diff
@@ -62,29 +63,32 @@ func resourceAwsCredentials() *schema.Resource {
 }
 
 func resourceAwsCredentialsCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	_, accessKeyExist := d.GetOk("access_key_id")
```
TomerHeber, line 66: Enforcing the exactlyOnce in the code level only for creation (not for read/import).

env0/resource_aws_credentials.go
```diff
@@ -62,29 +63,32 @@ func resourceAwsCredentials() *schema.Resource {
...
 
 	value := client.AwsCredentialsValuePayload{}
-	requestType := client.AwsAssumedRoleCredentialsType
-	if arn, ok := d.GetOk("arn"); ok {
-		value.RoleArn = arn.(string)
-		requestType = client.AwsAssumedRoleCredentialsType
-	}
-	if externalId, ok := d.GetOk("external_id"); ok {
-		value.ExternalId = externalId.(string)
+	if err := readResourceData(&value, d); err != nil {
```
TomerHeber, line 77: some small refactoring (less lines of code).

env0/utils.go
```diff
@@ -42,7 +42,13 @@ func readResourceData(i interface{}, d *schema.ResourceData) error {
 	for i := 0; i < val.NumField(); i++ {
 		fieldName := val.Type().Field(i).Name
 		// Assumes golang is CamalCase and Terraform is snake_case.
+		// This behavior can be overrided be used in the 'resource' tag.
 		fieldNameSC := toSnakeCase(fieldName)
+		if resFieldName, ok := val.Type().Field(i).Tag.Lookup("resource"); ok {
```
TomerHeber, line 47: This is for the new tag I created (resource). Ping me if this is still unclear.

env0/resource_aws_credentials_test.go
```diff
@@ -108,7 +116,7 @@ func TestUnitAwsCredentialsResource(t *testing.T) {
 		Steps: []resource.TestStep{
 			{
 				Config:      resourceConfigCreate(resourceType, resourceName, mutuallyExclusiveErrorResource),
-				ExpectError: regexp.MustCompile("only one of `access_key_id,arn` can be specified"),
+				ExpectError: regexp.MustCompile(`"external_id": conflicts with access_key_id`),
```
TomerHeber, line 119: There are multiple errors., the following just come before.

env0/resource_aws_credentials.go
```diff
@@ -98,11 +102,14 @@ func resourceAwsCredentialsCreate(ctx context.Context, d *schema.ResourceData, m
...
 
-	id := d.Id()
-	_, err := apiClient.CloudCredentials(id)
+	credentials, err := apiClient.CloudCredentials(d.Id())
 	if err != nil {
 		return ResourceGetFailure("aws credentials", d, err)
 	}
+
+	d.Set("name", credentials.Name)
+	d.SetId(credentials.Id)
```
yaronya, line 111: Can we use the seralization logic here instead?

env0/resource_aws_credentials.go
```diff
@@ -116,3 +123,18 @@ func resourceAwsCredentialsDelete(ctx context.Context, d *schema.ResourceData, m
...
+	credentials, err := getCredentials(d.Id(), "AWS_", meta)
+	if err != nil {
+		if _, ok := err.(*client.NotFoundError); ok {
+			return nil, fmt.Errorf("aws credentials resource with id %v not found", d.Id())
+		}
+		return nil, err
+	}
+
+	d.Set("name", credentials.Name)
+	d.SetId(credentials.Id)
```
yaronya, line 137: Same question here

env0/utils.go
yaronya: Very cool 👏 
Just wondering together with you whether `resourcee` is a decent tag name or we can come up with a better name. Maybe something that indicates that this is related to the TF schema would do better.

Also - WDYT about adding some note regarding it in `DEVELOPMENT.md`?

yaronya (commented):
Nice work!
Left some few minor comments

env0/utils.go
TomerHeber: yes and yes.
How about tfschema or tfresource ? (I'm bad with names).

yaronya: Both are better. If you want me to choose then I'd go with `tschema` 😅

env0/resource_aws_credentials.go
TomerHeber: yeah sure. In general, I use it when there a multiple fields.
Either way is fine by me. I'll fix it.

env0/resource_aws_credentials.go
TomerHeber: same answer.

env0/utils.go
TomerHeber: we have a winner (:

env0/resource_aws_credentials.go
```diff
@@ -14,6 +15,8 @@ func resourceAwsCredentials() *schema.Resource {
 		ReadContext:   resourceAwsCredentialsRead,
 		DeleteContext: resourceAwsCredentialsDelete,
 
+		Importer: &schema.ResourceImporter{StateContext: resourceAwsCredentialsImport},
```
yaronya, line 18: @TomerHeber if we're already at it, can we perhaps also support import for all other credentials type? I'd do it in a separate PR tho

env0/resource_aws_credentials.go
yaronya: I think always using it makes a lot of sense since we wouldn't need to change anything when we'll add a new field 💪

env0/resource_aws_credentials.go
TomerHeber: keeping the issue open.

env0/resource_aws_credentials.go
TomerHeber: done!

PR #372, opened 2022-05-11
Chore: Add random string to gcp creds

eranelbaz:
We have issues where our E2E breaks since we are trying to create another resource with the same name

### Solution
use random string


chpl (approved):
LGTM

PR #377, opened 2022-05-13
Fix: env0_environment resource and env0_configuration_variable config…

### LINKED ISSUE #347: env0_environment resource and env0_configuration_variable `configuration` block
**Summary**:
Users reports that the `env0_environment` has an issue that reports that the `configuration` block is missing and tries to delete all `env0_configuration_variable` where environment scope is defined.

**Repro steps**:
1. define `env0_environment` resource
2. define `env0_configuration_variable` that uses `env0_environment` id
3. run tf refresh / apply on `env0_environment` 
4. expect to see no changes
5. actual behavior: env0 deletes `env0_configuration_variables`

** Potential workaround  **
lifecycle ignore_changes on configuration{}

** Let's try to understand why there is this issue and share your findings

TomerHeber:
…uration block

fixes #347 

### Solution
Removed the field.

### DIFF
--- a/env0/resource_configuration_variable.go
+++ b/env0/resource_configuration_variable.go
@@ -50,14 +50,14 @@ func resourceConfigurationVariable() *schema.Resource {
 				Type:          schema.TypeString,
 				Description:   "create the variable under this project, not globally",
 				Optional:      true,
-				ConflictsWith: []string{"template_id", "environment_id"},
+				ConflictsWith: []string{"template_id"},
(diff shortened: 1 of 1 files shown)


yaronya:
@TomerHeber there is a discussion in Slack about halting this solution (I tagged you in it). 
We would like to figure first if there's another solution that's more suitable to our customers needs and usage. 
I'd suggest to close this PR until we decide it 🙏

PR #380, opened 2022-05-16
Fix: env0_gcp_credentials - apply after import will delete+create

### LINKED ISSUE #99: env0_aws_credentials - apply after import will delete+create

TomerHeber:
fixes #99 
The last part of the effort to add import to all credential types (Azure and AWS is complete).

### Solution
1. Added import.
2. Fixed some bugs.
3. Added acceptance tests.
4. Added examples.

### DIFF
--- a/env0/resource_gcp_credentials.go
+++ b/env0/resource_gcp_credentials.go
@@ -2,6 +2,7 @@ package env0
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/env0/terraform-provider-env0/client"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
@@ -13,6 +14,8 @@ func resourceGcpCredentials() *schema.Resource {
 		CreateContext: resourceGcpCredentialsCreate,
 		ReadContext:   resourceGcpCredentialsRead,
 		DeleteContext: resourceGcpCredentialsDelete,
+
+		Importer: &schema.ResourceImporter{StateContext: resourceGcpCredentialsImport},
 
 		Schema: map[string]*schema.Schema{
 			"name": {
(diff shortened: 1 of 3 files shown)


env0/resource_gcp_credentials_test.go
```diff
@@ -23,7 +26,7 @@ func TestUnitGcpCredentialsResource(t *testing.T) {
 	}
 
 	updateGcpCredentialResource := map[string]interface{}{
-		"name":                "testUpdate",
```
TomerHeber, line 26: bug in test.

env0/resource_gcp_credentials.go
```diff
@@ -67,10 +72,15 @@ func resourceGcpCredentialsRead(ctx context.Context, d *schema.ResourceData, met
...
 	id := d.Id()
-	_, err := apiClient.CloudCredentials(id)
+	credentials, err := apiClient.CloudCredentials(id)
 	if err != nil {
 		return ResourceGetFailure("gcp credentials", d, err)
 	}
+
+	if err := writeResourceData(&credentials, d); err != nil {
+		return diag.Errorf("schema resource data serialization failed: %v", err)
+	}
```
razbensimon, line 82: Just asking, what changed that we now need it than before?
Is `import` flow will not work without it?

razbensimon (approved):
👏🏼

env0/resource_gcp_credentials.go
TomerHeber: @razbensimon ,

The writeResourceData (and readReasourceData) is a utility function that reads all the fields in the interface and updates the schema via `set` calls.
This is future proof: in case you add more fields to the schema/api - no additional coding is needed (other than updating the schema).

There is a potential bug here that drifting is not detected (E.g., if someone changes the name in the UX). Once I added this code, it caught a bug in the tests as well (so I fixed the tests as well).
Import flow will work without it, and I just fixed bugs I encountered on the way.

PR #381, opened 2022-05-16

TomerHeber:
…ation logic

The code is going through refactoring. This is just one of many PRs. 

### Solution
Added "-" handling to tfschema (updated DEVELOPMENT.md).
Some refactoring of serialization. (Much more to do).


razbensimon (commented):
@TomerHeber generally looks good to me.
But I am not familiar with `tfschema` so I am not sure what the `-` actually means.
It is something supported by it? Or it is "customer syntax" you have added?
Because I see that on the read/write methods it needs a spacial reference.

Maybe can you explain more about it and the goal here?

razbensimon (approved):
Thanks for the explanation 👍

PR #388, opened 2022-05-19
Feat: support resolving terraform version from code for template reso…

### LINKED ISSUE #384: Support resolving terraform version from code for template resource
## Bug Description
After [this commit](https://github.com/env0/terraform-provider-env0/commit/633ce830908a61ed0227b6229a6dbca9009be98b) the functionality of the way Env0 sets the terraform_version was broken. Previously when the terraform version wasn't set, Env0 would grab the version to run the code from the required_version in the code template. The following were valid cases:

```terraform
resource "env0_template" "template" {
  ...
  terraform_version = ""
  ...
}

resource "env0_template" "template" {
  ...
  terraform_version = null
  ...
}

resource "env0_template" "template" {
  ...
  #no version set
  ...
}
```

Now, the default behavior is to set the terraform version to 0.15.1. This broke all env0 environments that relied on the required_version in the code template.

## Justification
Using the required_version in the code template makes it impossible for an engineer to accidently upgrade the Terraform version by running it locally.

By allowing required_version in a monorepo we can upgrade environments in a stepwise manner. 

## Acceptance Criteria
* As a user, I expect when terraform_version is passed `null` or `""` to have Env0 rely on the required_version in the code template
* As a user, I expect when terraform_version is not set to have Env0 rely on the required_version in the code template
* As a user, I expect that when there's a conflict between the template code's required version and a set version in the env0_template for Env0's value to take prescedence.

TomerHeber:
…urce

resolves #384

### Solution
1. RESOLVE_FROM_TERRAFORM_CODE is now a valid value.
2. Added the use-case to the integration tests.

### DIFF
--- a/env0/resource_template.go
+++ b/env0/resource_template.go
@@ -151,9 +151,9 @@ func resourceTemplate() *schema.Resource {
 			},
 			"terraform_version": {
 				Type:             schema.TypeString,
-				Description:      "the Terraform version to use (example: 0.15.1)",
+				Description:      "the Terraform version to use (example: 0.15.1). Setting to `RESOLVE_FROM_TERRAFORM_CODE` defaults to the version of `terraform.required_version` during run-time (resolve from terraform code).",
 				Optional:         true,
-				ValidateDiagFunc: NewRegexValidator(`^[0-9]\.[0-9]{1,2}\.[0-9]{1,2}$`),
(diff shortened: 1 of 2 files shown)


yaronya:
> @yaronya, while `RESOLVE_FROM_TERRAFORM_CODE` is longer, using it is a cleaner solution.
> 
> Let me know if you still prefer to use the `RESOLVE` keyword.

I'm fine with it either way 👍

yaronya (approved):
@TomerHeber please let's release a new version after merging this one 🙏

PR #389, opened 2022-05-20
Fix: data env0_templates includes deleted templates

### LINKED ISSUE #350: data `env0_templates` includes deleted templates
Using `data env0_templates` returns all templates, including deleted templates. The API call returns those as well.
`data env0_template` also uses that same API

We should filter out deleted templates (or not get them via the API to begin with)

TomerHeber:
fixes #350 

### Solution
1. Added an if condition to filter out deleted templates.
2. Uncommented integration test.
3. Updated acceptance test.


TomerHeber:
@RLRabinowitz - it looks like tests are still flaky.
Is it possible it takes time for the API to mark templates as isDeleted, even after selection?

RLRabinowitz (approved):
👍🏻

RLRabinowitz:
@TomerHeber Can you please make sure the templates in `012` are using a dynamic name? So that this test wouldn't be flaky

TomerHeber:
> @TomerHeber Can you please make sure the templates in `012` are using a dynamic name? So that this test wouldn't be flaky

@RLRabinowitz - sure thing. I'll fix that.

Issue #376, opened 2022-05-12
Refactor resources/data to use the new serialization/deserialization logic


yaronya:
We've recently introduced a new way of serialize/deserialize structs to/from API types and TF schema structure. 
This has been done to all resources moving forward from that point, but we still have plenty of resources and data resources that aren't using this utility. We should make an effort to align our codebase to use this utility.

Please make sure to split it to several pull requests as it might be a big effort.
Also - luckily all our resources are covered with UTs so if something would break - we'd know that on development phase 💪

yaronya:
@TomerHeber I see we've made a lot of progress with this effort. Can you share what's left? or how close are we to completing this?

TomerHeber:
@yaronya - I think it's almost done in all the places that make sense.
If there is something left, it's rather small.
I'll compose a list (if I find any left).

TomerHeber:
@yaronya - environment is a big one still left.
There are some challenges to it in general. I'm slowly refactoring.

PR #391, opened 2022-05-21

TomerHeber:
…ation logic

### Solution
Some minor refactoring.

Part of a larger effort #376.

### DIFF
--- a/env0/resource_project_policy.go
+++ b/env0/resource_project_policy.go
@@ -3,11 +3,8 @@ package env0
 import (
 	"context"
 	"errors"
-	"log"
-	"strings"
 
 	"github.com/env0/terraform-provider-env0/client"
-	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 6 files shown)


env0/resource_project_policy.go
```diff
@@ -28,39 +25,22 @@ func resourcePolicy() *schema.Resource {
...
-				ValidateDiagFunc: func(i interface{}, p cty.Path) diag.Diagnostics {
-					if strings.TrimSpace(i.(string)) == "" {
-						return diag.Errorf("project id must not be empty")
-					}
-					return nil
-				},
+				Type:             schema.TypeString,
+				Description:      "id of the project",
+				Required:         true,
+				ValidateDiagFunc: ValidateNotEmptyString,
```
yaronya, line 31: Do we really need to have this validation func when attribute is `Required`?

yaronya (approved):
LGTM
Left one question tho, please attend 🙏

env0/resource_project_policy.go
TomerHeber: So yeah... a weird thing...
It seems `project_id = ""` will path the `required` validation.
I was surprised as well...

PR #408, opened 2022-06-02
Chore: Add status field, which allows to wait for specific status

RLRabinowitz:
Currently, using `wait_for: FULLY_DEPLOYED` field on `env0_environment` resource, has a specific issue:

If environment fails to deploy on initial `terraform apply`, it'll get tainted. This means that on the next `terraform apply`, the environment will be destroyed and redeployed. If the user decides to `terraform untaint` the resource, then the environment resource would not detect any change on the next `terraform apply`, and therefor will not redeploy

Tainting happens by the terraform SDK, and this behaviour [cannot be changed](https://github.com/hashicorp/terraform-plugin-sdk/issues/330) as long as a multi-step create fails.

If we do not throw an error on deployment failure - then the TF code will continue deploying other dependant resources as though nothing's happened. Also, on subsequent `terraform apply`s, the environment resource will not redeploy
If we do not store this environment in state, then it'll just become unmanaged, and the next `terraform apply` will create another environment

### Solution

This PR proposes a partial solution for the issue: It helps TF realise that a redeploy is needed if a deployment has failed.
This PR introduces a new field to the environment resource: `status`. Users have to add `status: "ACTIVE"` to the environment resource, alongside `wait_for: "FULLY_DEPLOYED"`

In conjunction with auto-untainting using [env0 custom flows](https://docs.env0.com/docs/custom-flows), the behaviour here would be as expected: 
If initial environment deployment fails, the `terraform apply` will fail. The next `terraform apply` would trigger a redeploy


avnerenv0:
@RLRabinowitz since https://github.com/env0/terraform-provider-env0/issues/424
Can we close this one?

RLRabinowitz:
@avnerenv0 Yes, this is irrelevant either way. Closing

PR #413, opened 2022-06-08

TomerHeber:
### Solution

Some refactoring in resource configuration variable.
Part of a larger refactoring effort #376

### DIFF
--- a/env0/resource_configuration_variable.go
+++ b/env0/resource_configuration_variable.go
@@ -143,50 +143,26 @@ func whichScope(d *schema.ResourceData) (client.Scope, string) {
 }
 
 func resourceConfigurationVariableCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-	apiClient := meta.(client.ApiClientInterface)
 	scope, scopeId := whichScope(d)
-	name := d.Get("name").(string)
-	description := d.Get("description").(string)
-	value := d.Get("value").(string)
-	isSensitive := d.Get("is_sensitive").(bool)
-	typeAsString := d.Get("type").(string)
-	format := client.Format(d.Get("format").(string))
-	isReadOnly := d.Get("is_read_only").(bool)
-	isRequired := d.Get("is_required").(bool)
-	regex := d.Get("regex").(string)
+	params := client.ConfigurationVariableCreateParams{Scope: scope, ScopeId: scopeId}
+	if err := readResourceData(&params, d); err != nil {
+		return diag.Errorf("schema resource data deserialization failed: %v", err)
+	}
 
-	if err := validateNilValue(isReadOnly, isRequired, value); err != nil {
+	if err := validateNilValue(params.IsReadOnly, params.IsRequired, params.Value); err != nil {
 		return diag.Errorf(err.Error())
 	}
 
-	var type_ client.ConfigurationVariableType
-	switch typeAsString {
-	case "environment":
-		type_ = client.ConfigurationVariableTypeEnvironment
-	case "terraform":
-		type_ = client.ConfigurationVariableTypeTerraform
-	default:
-		return diag.Errorf("'type' can only receive either 'environment' or 'terraform': %s", typeAsString)
-	}
-	actualEnumValues, getEnumErr := getEnum(d, value)
+	actualEnumValues, getEnumErr := getEnum(d, params.Value)
 	if getEnumErr != nil {
 		return getEnumErr
 	}
 
-	configurationVariable, err := apiClient.ConfigurationVariableCreate(client.ConfigurationVariableCreateParams{
-		Name:        name,
-		Value:       value,
-		IsSensitive: isSensitive,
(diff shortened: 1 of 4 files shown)


env0/resource_configuration_variable.go
```diff
@@ -143,50 +143,26 @@ func whichScope(d *schema.ResourceData) (client.Scope, string) {
...
+	if err := readResourceData(&params, d); err != nil {
+		return diag.Errorf("schema resource data deserialization failed: %v", err)
+	}
 
-	if err := validateNilValue(isReadOnly, isRequired, value); err != nil {
+	if err := validateNilValue(params.IsReadOnly, params.IsRequired, params.Value); err != nil {
 		return diag.Errorf(err.Error())
 	}
 
-	var type_ client.ConfigurationVariableType
```
TomerHeber, line 162: The serializer knows how to handle this use case.
see environment.go `func (c *ConfigurationVariableType) ReadResourceData(fieldName string, d *schema.ResourceData) error {...`

env0/utils.go
```diff
@@ -71,13 +71,22 @@ func readResourceData(i interface{}, d *schema.ResourceData) error {
...
 
+		// custom field is a pointer.
 		if customField, ok := field.Interface().(CustomResourceDataField); ok {
 			if err := customField.ReadResourceData(fieldNameSC, d); err != nil {
 				return err
 			}
 			continue
 		}
 
+		// custom field is a value, check the pointer.
```
TomerHeber, line 82: This is a small improvement to support custom serialization for non-pointer types.

env0/utils.go
```diff
@@ -71,13 +71,22 @@ func readResourceData(i interface{}, d *schema.ResourceData) error {
...
+		if customField, ok := field.Addr().Interface().(CustomResourceDataField); ok {
```
sagilaufer1992, line 83: what happens when both those checks are not ok? will there be an error? or is it not actually possible?

sagilaufer1992 (commented):
looks really good. I just need one clarification about some checks that have no handling with the negative outcome

env0/utils.go
sagilaufer1992: does it require additional tests? for both cases of pointer and non-pointer?

TomerHeber: pointer one is tested in both acceptance test and unit tests.
The non-pointer is tested in just acceptance test. I'll add another unit teat.

env0/utils.go
TomerHeber: It's possible. In most cases, both will return false. (These cases are uncommon).
It will just continue the function run.

env0/utils.go
TomerHeber: Added a test see below.

env0/utils_test.go
```diff
@@ -199,6 +199,22 @@ func TestWriteCustomResourceData(t *testing.T) {
...
+		"name":        "name",
+		"description": "description",
+	})
+
+	params := client.ConfigurationVariableCreateParams{}
+
+	assert.Nil(t, readResourceData(&params, d))
+
+	assert.Equal(t, params.Name, "name")
+	assert.Equal(t, int(params.Type), 1)
```
TomerHeber, line 214: This is what it tests: the value "terraform" was successfully modified to 1.

sagilaufer1992: and it did it with a non-pointer, as opposed to previous test? mind mention where is the other type of test, so I'll see the difference, for future PRs?

TomerHeber: @sagilaufer1992 - you actually made a very good point. I actually missed the test for it, and it discovered an issue.
See my latest commit for the changes.

TomerHeber: https://github.com/env0/terraform-provider-env0/pull/413/commits/1e720ed825432818109ddf32879804e1f6a95fbc

env0/utils_test.go
```diff
@@ -199,6 +199,53 @@ func TestWriteCustomResourceData(t *testing.T) {
...
+	assert.Equal(t, params.Description, "description")
+}
+
+func TestReadByPointerCustomResourceData(t *testing.T) {
```
TomerHeber, line 218: additional test for the use-case.

env0/utils_test.go
```diff
@@ -199,6 +199,53 @@ func TestWriteCustomResourceData(t *testing.T) {
...
+	params := client.ConfigurationVariable{}
+
+	assert.Nil(t, readResourceData(&params, d))
+
+	assert.Equal(t, params.Name, "name")
+	assert.Equal(t, int(*params.Type), 1)
+	assert.Equal(t, params.Description, "description")
+}
+
+func TestReadByPointerNilCustomResourceData(t *testing.T) {
```
TomerHeber, line 234: I want to be sure if not set in schema, params.Type remains nil.

sagilaufer1992 (approved):
LGTM. After the tests will pass, merge it

Issue #394, opened 2022-05-26
Refactor credentials and cost.


TomerHeber:
There are large parts of code that are duplicated between GCP, Azure and AWS. Also between credentials and cost credentials, and finally in the acceptance tests.

TomerHeber:
This is now done. Reviewing the last part (the credential test files). I don't see much benefit in refactoring it (each varies. no real common code).

PR #414, opened 2022-06-09
Chore: refactor credentials and cost

TomerHeber:
### Solution

Moved some common API calls to use the same code.
Part of #394

### DIFF
--- a/client/cloud_credentials.go
+++ b/client/cloud_credentials.go
@@ -23,6 +23,10 @@ func (c *Credentials) HasPrefix(prefixList []string) bool {
 	return false
 }
 
+type CredentialCreatePayload interface {
+	SetOrganizationId(organizationId string)
+}
+
 type AzureCredentialsCreatePayload struct {
 	Name           string                       `json:"name"`
 	OrganizationId string                       `json:"organizationId"`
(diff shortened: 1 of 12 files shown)


client/cloud_credentials.go
```diff
@@ -75,6 +79,22 @@ type GcpCredentialsValuePayload struct {
...
+	c.OrganizationId = organizationId
+}
+
+func (c *GcpCredentialsCreatePayload) SetOrganizationId(organizationId string) {
+	c.OrganizationId = organizationId
+}
+
+func (c *AzureCredentialsCreatePayload) SetOrganizationId(organizationId string) {
+	c.OrganizationId = organizationId
+}
```
HeverFarber, line 96: I don't know if it is related. but maybe after you start to support [genric](https://github.com/env0/terraform-provider-env0/pull/417#pullrequestreview-1003569256) you can bundle it

HeverFarber (approved):
I approved it. but check if you can use the generic for it for more refactor

client/cloud_credentials.go
TomerHeber: Generics is still a little limited in 1.18.
I'll see if I can work around it. Otherwise, may need to revisit in go 1.19.

TomerHeber: @HeverFarber 

I tried, but it seems like a limitation in golang 1.18 generics (I cannot access struct fields).
More details in https://stackoverflow.com/questions/70358216/how-can-i-define-a-struct-field-in-my-interface-as-a-type-constraint-type-t-has

We could revisit it in the future (if the remove the restriction in go1.19).

Generics is very recent... so it has some limitations at the moment /:

PR #417, opened 2022-06-11
Chore: use generics in pagination

### LINKED ISSUE #415: Chore: use generics in pagination
A new version of Alpine was released that support go 1.18.
We can now proceed to use generics. (Revert to the //FIXME code)

TomerHeber:
closes #415

### Solution

Harness was running in go1.17 because the terraform docker image was using golang 1.17.
Switched to using a go1.18 docker image and installed terraform via `apk`.

Making this change allows me to use generics (A new feature in go1.18).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,12 +45,12 @@ jobs:
   integration-tests:
     name: Integration Tests
     runs-on: ubuntu-20.04
-    container: hashicorp/terraform:1.1.6
+    container: golang:1.18-alpine3.16
     timeout-minutes: 20
     steps:
-      - name: Install Go
-        run: apk add go
+      - name: Install Terraform
+        run: apk add terraform
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Run Harness tests
         run: go run tests/harness.go
--- a/client/pagination.go
+++ b/client/pagination.go
@@ -9,11 +9,8 @@ type Pagination struct {
 	params map[string]string
 }
 
-// Note: all the FIXME should be uncommented when Alpine adds support for go 1.18.
-// This will enable generics.
-
 type Paginated interface {
-	// FIXME: Environment
+	Environment
 	getEndpoint() string
 }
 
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -45,12 +45,12 @@ jobs:
   integration-tests:
     name: Integration Tests
     runs-on: ubuntu-20.04
-    container: hashicorp/terraform:1.1.6
+    container: golang:1.18-alpine3.16
     timeout-minutes: 20
     steps:
-      - name: Install Go
-        run: apk add go
```
TomerHeber, line 52: no need to install go - 1.18 is already installed as part of the docker image.
Before it would install go1.17.

.github/workflows/ci.yml
```diff
@@ -45,12 +45,12 @@ jobs:
...
+      - name: Install Terraform
+        run: apk add terraform
```
TomerHeber, line 52: installing terraform.

.github/workflows/ci.yml
```diff
@@ -45,12 +45,12 @@ jobs:
...
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
```
TomerHeber, line 54: noticed some depreciation warnings. Went away when switching to v3.

.pre-commit-config.yaml
```diff
@@ -12,7 +12,6 @@ repos:
     hooks:
       - id: go-fmt
       - id: go-vet
-      - id: go-imports
```
TomerHeber, line 15: go-imports doesn't seem to support generics.
This means the tool is not maintained. So better not use it anyways.

client/pagination.go
```diff
@@ -9,11 +9,8 @@ type Pagination struct {
 	params map[string]string
 }
 
-// Note: all the FIXME should be uncommented when Alpine adds support for go 1.18.
-// This will enable generics.
-
 type Paginated interface {
-	// FIXME: Environment
+	Environment
```
TomerHeber, line 13: Adding back the generics in this file...

.pre-commit-config.yaml
TomerHeber: the goimports on my machine was outdated...

HeverFarber (approved):

Issue #405, opened 2022-06-01
Feat: Support Non-Admin API key creation


tomer-landesman:
#### Motivation / Problem / The Why❔:

[API keys docs](https://docs.env0.com/docs/api-keys)

We recently implemented non-admin API key feature, which allows an Admin to create a partially privileged API key that can be added to a team or a specific project with specific permissions.

we need to also support that on TF provider.

#### Expected Result / DoD ✅:
- support creation of non-admin API key
- support deletion of non-admin API key
- support assigning non-admin key to a project
- support assigning non-admin key to a team

TomerHeber:
Hi @tomer-landesman - it would be helpful to point to the API documents.

https://developer.env0.com/docs/api

PR #419, opened 2022-06-13
Feat: support Non-Admin API key creation

TomerHeber:
### Solution

Updated integration test, to test api-key team assignment.
Added API calls to add support for api-key project assignment.

Part of #405.


tests/integration/020_api_key/main.tf
```diff
@@ -11,10 +11,20 @@ resource "env0_api_key" "test_api_key" {
...
+  name              = "my-little-user-api-key-${random_string.random.result}"
   organization_role = "User"
 }
 
+resource "env0_team" "team_resource" {
+  name        = "team-with-api-key"
+  description = "description"
+}
+
+resource "env0_user_team_assignment" "api_key_team_assignment" {
```
TomerHeber, line 23: This is for testing team api-key assignment (Which is just user team assignment).

client/api_key_project_assignment.go
```diff
@@ -0,0 +1,33 @@
+package client
+
+type ApiKeyRole string
+
+const (
+	ApiKeyRoleDeployer ApiKeyRole = "Deployer"
+	ApiKeyRoleViewer   ApiKeyRole = "Viewer"
+	ApiKeyRolePlanner  ApiKeyRole = "Planner"
+	ApiKeyRoleAdmin    ApiKeyRole = "Admin"
+)
```
avnerenv0, line 10: we've got these also defined in `client/team_project_assignment.go`
Those are the same roles a regular user/team have. Can we reuse them?

client/api_key_project_assignment.go
```diff
@@ -0,0 +1,33 @@
...
+type AssignAPIKeyToProjectPayload struct {
+	UserId string     `json:"userId"`
+	Role   ApiKeyRole `json:"role"`
+}
+
+type ApiKeyProjectAssignment struct {
+	Id string `json:"id"`
+}
+
+func (client *ApiClient) AssignApiKeyToProject(projectId string, payload *AssignAPIKeyToProjectPayload) (*ApiKeyProjectAssignment, error) {
```
avnerenv0, line 21: These endpoints aren't specific to api keys, they are for assigning users to projects, and a "non admin api key" is just a user at the end of it. 

WDYT about renaming this file/methods to reflect that?

tests/integration/020_api_key/main.tf
avnerenv0: not sure if it's part of the scope of this PR, but as part of #405 can we document somewhere that the `user_id` can also be an `api_key.id` ?

avnerenv0 (approved):
LGTM, comments for your consideration

client/api_key_project_assignment.go
TomerHeber: I think it's a good idea. I'll make some changes.

client/api_key_project_assignment.go
TomerHeber: Yes.

avnerenv0:
Go for it
<div><br/>via <a href="https://giphy.com/thefastsaga/">The Fast Saga</a> on <a href="https://giphy.com/gifs/thefastsaga-fast-and-furious-dnCNI8LEf7K4otcMqS">GIPHY</a></div>

PR #428, opened 2022-06-24
Feat: support newly added template types - pulumi, k8s & workflow

### LINKED ISSUE #425: Support newly added template types - pulumi, k8s & workflow
When creating a template in env0 you can create
- terraform
- terragrunt
- pulumi
- k8s
- complex

using the provider you can only create terraform or terragrunt. 
please make the solution generic since we are going to add more and more template types in the future

* that is also relevant for `data env_template`

for example https://github.com/env0/terraform-provider-env0/blob/29fd96b88387c7839f89bc14c253b126c6055723/env0/data_template.go#L43-L47

but we can query `pulumi` template using the provider

TomerHeber:
resolves #425 

### Solution

Added the new types to the schema (resource and data) + documentation.

### DIFF
--- a/env0/resource_template_test.go
+++ b/env0/resource_template_test.go
@@ -502,7 +502,7 @@ func TestUnitTemplateResource(t *testing.T) {
 						"repository": template.Repository,
 						"type":       "gruntyform",
 					}),
-					ExpectError: regexp.MustCompile(`must be one of: terragrunt, terraform`),
+					ExpectError: regexp.MustCompile(`must be one of`),
 				},
 			},
 		}
--- a/env0/templates.go
+++ b/env0/templates.go
@@ -20,7 +20,7 @@ var allowedTemplateTypes = []string{
 	"terragrunt",
 	"pulumi",
 	"k8s",
-	"complex",
+	"workflow",
 }
 
 func getTemplateSchema(templateType TemplateType) map[string]*schema.Schema {
--- a/env0/templates.go
+++ b/env0/templates.go
@@ -1,7 +1,9 @@
 package env0
 
 import (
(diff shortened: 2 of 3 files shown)


env0/templates.go
```diff
@@ -13,6 +15,14 @@ const (
 	TemplateTypeShared = 2
 )
 
+var allowedTemplateTypes = []string{
+	"terraform",
+	"terragrunt",
+	"pulumi",
+	"k8s",
+	"complex",
```
tomporat247, line 23: we renamed `complex` to `workflow`

env0/resource_template_test.go
```diff
@@ -502,7 +502,7 @@ func TestUnitTemplateResource(t *testing.T) {
 						"repository": template.Repository,
 						"type":       "gruntyform",
 					}),
-					ExpectError: regexp.MustCompile(`must be one of: terragrunt, terraform`),
+					ExpectError: regexp.MustCompile(`must be one of`),
```
tomporat247, line 505: Does this work? shouldn't it list all of the available types

tomporat247 (commented):
Looking good, Notice that we renamed the blueprint type so please adjust your PR accordingly

env0/templates.go
TomerHeber: fixed.

env0/resource_template_test.go
TomerHeber: It works. I didn't want us to modify the test every time a new type was added.
(This test isn't beneficial, to begin with).

Since it's a regex match, it passes because the prefix never changes.

tomporat247 (approved):

PR #430, opened 2022-06-26
Chore: remove wait_for

### LINKED ISSUE #424: Deprecate and remove `wait_for`
After some internal discussions we decided we would not like to support the `wait_for` property of the `environment` resource.

This is mainly due to 
- It's not really working well when there are failures in the environment's initial deployment - a common happening. That issue is also very hard to mitigate since TF will taint the resource. 
- We are introducing 'complex workflows' to the platform which are a much better of way of handling this use case.
- The provider should be used to configure env0, but not to "run" it, and 'wait_for' mixes the two needs.

We should deprecate this feature ASAP and remove it in the next major version.
The sooner the better - AFAIK no one is using this feature ATM.

TomerHeber:
…h deployment (#273)"

This reverts commit 2e045e5e31a36e4980c09cba74018336f7b234b4.

resolves #424 

### Solution

git revert was used against 2e045e5e31a36e4980c09cba74018336f7b234b4 + small manual changes (after fixing conflicts).


TomerHeber:
This PR is blocked until it's decided we can merge it (in about two weeks).

tests/integration/012_environment/main.tf
```diff
@@ -34,11 +33,6 @@ resource "env0_environment" "example" {
   vcs_commands_alias         = "alias"
 }
 
-data "env0_configuration_variable" "env_config_variable" {
```
TomerHeber, line 37: This was added in the original commit. I guess it no longer works properly without waiting.
Had to remove it for the test to pass.

TomerHeber:
Note: Will not merge until I get approval in the slack channel as well.

eranelbaz:
Is it breaking changes?
we need to look for something when we release it?

eranelbaz (approved):

yaronya:
> Is it breaking changes?
we need to look for something when we release it?

Yes, it'd require a new major version and to be noted as a breaking change in release notes @TomerHeber

TomerHeber:
> > Is it breaking changes?
> > we need to look for something when we release it?
> 
> Yes, it'd require a new major version and to be noted as a breaking change in release notes @TomerHeber

@yaronya - added a commit with a breaking change comment.
Let me know when I can push it and create a new version.

yaronya:
@TomerHeber go for it 🙏

Issue #374, opened 2022-05-11
Environment resource does not support new templateless creation


athak:
Hi,
As of a few weeks ago, [environments can be created without a template](https://docs.env0.com/changelog/environment-without-template), which is very useful. 

However, the `env0_environment` resource, still requires a template id to be supplied, effectively not allowing this kind of environment to be created via Terraform.

It would be great to add support of templateless environment to the resource.

Thanks,
Atha

TomerHeber:
@omry-hay @yaronya - I'm marking it as blocked for now. I recommend discussing this internally to decide what would be the best way to add support for this feature in the provider.

yaronya:
@TomerHeber I agree, we'll take this internally and report back

TomerHeber:
Suggested solution:

When creating an env, to extend the current existing variables for creation to include the VCS and other relevant data that is needed, all under a new fields template_data. if such a field exists, we use the api client to create a template by the given data, and pass isSingleUse=true. the returned template id is then used as in the regular creating of env, and we also pass isSingleUse=true

yaronya:
@TomerHeber what's the status of this one?

TomerHeber:
@yaronya - the schema is available. I still need to implement the "resource" code where there are two cases (one with a template in the schema the other with not).
I've been carefully and slowly working towards it for several reasons:
1. keep reviews small.
2. avoid breaking existing code.
3. avoid duplicating code (a lot of the work has been around refactoring and using the read/write serializers).

PR #432, opened 2022-06-28

TomerHeber:
### Solution

Some required refactoring for supporting templateless environment.
Part of #374.

### DIFF
--- a/client/template.go
+++ b/client/template.go
@@ -49,14 +49,14 @@ type Template struct {
 	Retry                TemplateRetry    `json:"retry"`
 	SshKeys              []TemplateSshKey `json:"sshKeys"`
 	Type                 string           `json:"type"`
-	GithubInstallationId int              `json:"githubInstallationId"`
+	GithubInstallationId int              `json:"githubInstallationId" tfschema:",omitempty"`
 	IsGitlabEnterprise   bool             `json:"isGitLabEnterprise"`
-	TokenId              string           `json:"tokenId,omitempty"`
+	TokenId              string           `json:"tokenId,omitempty" tfschema:",omitempty"`
 	UpdatedAt            string           `json:"updatedAt"`
 	TerraformVersion     string           `json:"terraformVersion"`
 	TerragruntVersion    string           `json:"terragruntVersion,omitempty"`
 	IsDeleted            bool             `json:"isDeleted,omitempty"`
-	BitbucketClientKey   string           `json:"bitbucketClientKey"`
+	BitbucketClientKey   string           `json:"bitbucketClientKey" tfschema:",omitempty"`
 	IsGithubEnterprise   bool             `json:"isGitHubEnterprise"`
 	IsBitbucketServer    bool             `json:"isBitbucketServer"`
 	IsSingleUse          bool             `json:"isSingleUse"`
--- a/client/template.go
+++ b/client/template.go
@@ -52,7 +52,6 @@ type Template struct {
 	GithubInstallationId int              `json:"githubInstallationId"`
 	IsGitlabEnterprise   bool             `json:"isGitLabEnterprise"`
 	TokenId              string           `json:"tokenId,omitempty"`
-	GitlabProjectId      int              `json:"gitlabProjectId,omitempty"`
 	UpdatedAt            string           `json:"updatedAt"`
 	TerraformVersion     string           `json:"terraformVersion"`
 	TerragruntVersion    string           `json:"terragruntVersion,omitempty"`
--- a/env0/resource_template.go
+++ b/env0/resource_template.go
@@ -71,6 +71,16 @@ func resourceTemplateCreate(ctx context.Context, d *schema.ResourceData, meta in
 	return nil
 }
 
+func templateReadRetryOnHelper(d *schema.ResourceData, retryType string, retryOn *client.TemplateRetryOn) {
+	if retryOn != nil {
+		d.Set("retries_on_"+retryType, retryOn.Times)
+		d.Set("retry_on_"+retryType+"_only_when_matches_regex", retryOn.ErrorRegex)
+	} else {
+		d.Set("retries_on_"+retryType, 0)
+		d.Set("retry_on_"+retryType+"_only_when_matches_regex", "")
+	}
+}
+
 func resourceTemplateRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	apiClient := meta.(client.ApiClientInterface)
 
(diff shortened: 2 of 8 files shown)


client/template.go
```diff
@@ -52,7 +52,6 @@ type Template struct {
 	GithubInstallationId int              `json:"githubInstallationId"`
 	IsGitlabEnterprise   bool             `json:"isGitLabEnterprise"`
 	TokenId              string           `json:"tokenId,omitempty"`
-	GitlabProjectId      int              `json:"gitlabProjectId,omitempty"`
```
TomerHeber, line 55: This is never returned as part of a template response. Removed it.

env0/resource_template.go
```diff
@@ -85,49 +95,12 @@ func resourceTemplateRead(ctx context.Context, d *schema.ResourceData, meta inte
...
-		d.Set("retries_on_destroy", template.Retry.OnDestroy.Times)
-		d.Set("retry_on_destroy_only_when_matches_regex", template.Retry.OnDestroy.ErrorRegex)
-	} else {
-		d.Set("retries_on_destroy", 0)
-		d.Set("retry_on_destroy_only_when_matches_regex", "")
-	}
-
-	if template.BitbucketClientKey != "" {
-		d.Set("bitbucket_client_key", template.BitbucketClientKey)
+	if err := writeResourceData(&template, d); err != nil {
```
TomerHeber, line 98: using writeResourceData instead....

env0/resource_template_test.go
```diff
@@ -308,8 +308,12 @@ func TestUnitTemplateResource(t *testing.T) {
 		if template.TokenId != "" {
 			templateAsDictionary["token_id"] = template.TokenId
 		}
-		if template.GitlabProjectId != 0 {
-			templateAsDictionary["gitlab_project_id"] = template.GitlabProjectId
+		if template.Id == gitlabTemplate.Id {
+			if template.Name == gitlabUpdatedTemplate.Name {
+				templateAsDictionary["gitlab_project_id"] = gitlabTemplateUpdatedProjectId
```
TomerHeber, line 313: This is where I set the gitlab_project_id... the "if" is used to detect the relevant use-case

env0/resource_template_test.go
```diff
@@ -697,44 +713,6 @@ func TestUnitTemplateResource(t *testing.T) {
 		})
 	}
 
-	t.Run("Should not trigger terraform changes when gitlab_project_id is provided", func(t *testing.T) {
```
TomerHeber, line 700: This test is incorrect.
It should trigger changes.

See: https://developer.env0.com/docs/api/c667ff5b5fecb-update-template

env0/templates.go
```diff
@@ -45,17 +45,21 @@ func getTemplateSchema(templateType TemplateType) map[string]*schema.Schema {
 		"is_github_enterprise",
 	}
 
+	appendByType := func(strs []string, str string) []string {
```
TomerHeber, line 48: just moved to a common function... (requested in previous PR).

env0/utils.go
```diff
@@ -230,10 +230,16 @@ func writeResourceData(i interface{}, d *schema.ResourceData) error {
 
 		switch fieldType.Kind() {
 		case reflect.String:
+			if field.String() == "" {
```
TomerHeber, line 233: This is a little naive. But assume if empty - skip (don't set).

env0/templates.go
```diff
@@ -45,17 +45,21 @@ func getTemplateSchema(templateType TemplateType) map[string]*schema.Schema {
...
 		ret := []string{}
 
 		for _, attr := range allVCSAttributes {
 			if sort.SearchStrings(strs, attr) >= len(strs) {
-				if templateType == TemplateTypeShared {
-					ret = append(ret, attr)
-				} else {
-					ret = append(ret, "template.0."+attr)
-				}
+				ret = appendByType(ret, attr)
```
razbensimon, line 62: I'm not sure what is `ret`. can you rename and use full words when naming variables? 🙏🏼

env0/utils.go
razbensimon: are you sure? 
don't we have cases where this is a legit value? (maybe it is wrong on the BW side but we still need to support this case if that happen)

env0/utils.go
```diff
@@ -230,10 +230,16 @@ func writeResourceData(i interface{}, d *schema.ResourceData) error {
...
+				continue
+			}
 			if err := d.Set(fieldName, field.String()); err != nil {
 				return err
 			}
 		case reflect.Int:
+			if field.Int() == 0 {
+				continue
+			}
```
razbensimon, line 242: same here, 0 is a valid int value.
Am I missing something?

env0/templates.go
TomerHeber: sure!

env0/utils.go
TomerHeber: @razbensimon - you may review the additional changes in https://github.com/env0/terraform-provider-env0/pull/432/commits/accbc9be68f7af3bc3d81abcba96d70e2382df2b

env0/utils.go
```diff
@@ -168,36 +169,54 @@ func readResourceDataSlice(field reflect.Value, resources []interface{}) error {
...
+	name      string
+	skip      bool
+	omitEmpty bool
+}
+
 // Returns the field name or skip.
-func getFieldName(field reflect.StructField) (string, bool) {
-	name := field.Name
+func getFieldName(field reflect.StructField) *parsedField {
+	var res parsedField
```
TomerHeber, line 180: since 3 values could be returned, switched to a struct.

env0/utils.go
```diff
@@ -168,36 +169,54 @@ func readResourceDataSlice(field reflect.Value, resources []interface{}) error {
...
 	// Assumes golang is CamalCase and Terraform is snake_case.
 	// This behavior can be overrided be used in the 'tfschema' tag.
-	name = toSnakeCase(name)
+	res.name = toSnakeCase(field.Name)
 	if tag, ok := field.Tag.Lookup("tfschema"); ok {
 		if tag == "-" {
-			return "", true
+			res.skip = true
+		} else {
+			tagParts := strings.Split(tag, ",")
```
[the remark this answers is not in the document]
yaronya: Perhaps `DEVELOPMENT.md` should also include these instructions?

yaronya (approved):

env0/utils.go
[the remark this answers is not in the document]
TomerHeber: @yaronya - I will merge this PR and add it next week. My internet doesn't allow me to use git.
I will be home next week with a normal internet and get it added.

PR #433, opened 2022-06-28
Feat: support TTL settings in project policies

### LINKED ISSUE #418: Feat: support TTL settings in project policies
We adding TTL settings to [project policies](https://registry.terraform.io/providers/env0/env0/latest/docs/resources/project_policy)

It is mostly the same as in [organization policies](https://registry.terraform.io/providers/env0/env0/latest/docs/resources/organization_policy), but to be backward compatible with existing usage of `env0_project_policy` when omitted it should mean "use organization settings" - in this case pass `inherit` value to the API. 
To set infinite TTL, send `null` to the API.

TomerHeber:
resolves #418 

### Solution

Added the new fields and updated the tests.
Fixed some observed bugs.

### DIFF
--- a/client/project_policy.go
+++ b/client/project_policy.go
@@ -1,26 +1,26 @@
 package client
 
-import "encoding/json"
-
 type Policy struct {
-	Id                          string `json:"id"`
-	ProjectId                   string `json:"projectId"`
-	NumberOfEnvironments        int    `json:"numberOfEnvironments"`
-	NumberOfEnvironmentsTotal   int    `json:"numberOfEnvironmentsTotal"`
-	RequiresApprovalDefault     bool   `json:"requiresApprovalDefault"`
-	IncludeCostEstimation       bool   `json:"includeCostEstimation"`
-	SkipApplyWhenPlanIsEmpty    bool   `json:"skipApplyWhenPlanIsEmpty"`
-	DisableDestroyEnvironments  bool   `json:"disableDestroyEnvironments"`
-	SkipRedundantDeployments    bool   `json:"skipRedundantDeployments"`
-	UpdatedBy                   string `json:"updatedBy"`
-	RunPullRequestPlanDefault   bool   `json:"runPullRequestPlanDefault"`
-	ContinuousDeploymentDefault bool   `json:"continuousDeploymentDefault"`
+	Id                          string  `json:"id"`
+	ProjectId                   string  `json:"projectId"`
+	NumberOfEnvironments        *int    `json:"numberOfEnvironments,omitempty"`
+	NumberOfEnvironmentsTotal   *int    `json:"numberOfEnvironmentsTotal,omitempty"`
+	RequiresApprovalDefault     bool    `json:"requiresApprovalDefault"`
+	IncludeCostEstimation       bool    `json:"includeCostEstimation"`
+	SkipApplyWhenPlanIsEmpty    bool    `json:"skipApplyWhenPlanIsEmpty"`
+	DisableDestroyEnvironments  bool    `json:"disableDestroyEnvironments"`
+	SkipRedundantDeployments    bool    `json:"skipRedundantDeployments"`
+	UpdatedBy                   string  `json:"updatedBy"`
+	RunPullRequestPlanDefault   bool    `json:"runPullRequestPlanDefault"`
+	ContinuousDeploymentDefault bool    `json:"continuousDeploymentDefault"`
+	MaxTtl                      *string `json:"maxTtl,omitempty"`
+	DefaultTtl                  *string `json:"defaultTtl,omitempty"`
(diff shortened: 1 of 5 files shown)


client/project_policy.go
```diff
@@ -1,68 +1,35 @@
...
 	RequiresApprovalDefault     bool   `json:"requiresApprovalDefault"`
 	IncludeCostEstimation       bool   `json:"includeCostEstimation"`
 	SkipApplyWhenPlanIsEmpty    bool   `json:"skipApplyWhenPlanIsEmpty"`
 	DisableDestroyEnvironments  bool   `json:"disableDestroyEnvironments"`
 	SkipRedundantDeployments    bool   `json:"skipRedundantDeployments"`
 	RunPullRequestPlanDefault   bool   `json:"runPullRequestPlanDefault"`
 	ContinuousDeploymentDefault bool   `json:"continuousDeploymentDefault"`
-}
-
-func (p PolicyUpdatePayload) MarshalJSON() ([]byte, error) {
```
TomerHeber, line 33: No longer needed. I switched numberOfEnvironments* to pointers.

client/project_policy.go
```diff
@@ -1,68 +1,35 @@
...
-
-	if p.NumberOfEnvironments != 0 {
-		s.NumberOfEnvironments = &p.NumberOfEnvironments
-	}
-	if p.NumberOfEnvironmentsTotal != 0 {
-		s.NumberOfEnvironmentsTotal = &p.NumberOfEnvironmentsTotal
-	}
-
-	return json.Marshal(s)
+	MaxTtl                      string `json:"maxTtl"`
```
TomerHeber, line 31: For the update case these are not pointers - because the schema defaults to "inherit" value will always be available.

env0/resource_project_policy.go
```diff
@@ -3,13 +3,18 @@ package env0
...
+	"fmt"
+	"strings"
 
 	"github.com/env0/terraform-provider-env0/client"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
 func resourcePolicy() *schema.Resource {
+	allowedProjectTtlValues := append(allowedTtlValues, "inherit")
```
TomerHeber, line 15: I decided to just add "inherit" and not "null".
The reason for not adding "null" - it's not available in the UI.

env0/resource_project_policy.go
```diff
@@ -92,9 +111,14 @@ func setPolicySchema(d *schema.ResourceData, policy client.Policy) error {
 }
 
 func resourcePolicyCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	if err := resourcePolicyUpdate(ctx, d, meta); err != nil {
```
TomerHeber, line 114: Should setId only if the update was successful (this was a bug I found during testing).

env0/resource_project_policy.go
```diff
@@ -28,6 +33,7 @@ func resourcePolicy() *schema.Resource {
 				Type:             schema.TypeString,
 				Description:      "id of the project",
 				Required:         true,
+				ForceNew:         true,
```
avnerenv0, line 36: why `ForceNew`?

avnerenv0: oh got it

env0/resource_project_policy.go
```diff
@@ -119,15 +143,24 @@ func resourcePolicyRead(ctx context.Context, d *schema.ResourceData, meta interf
...
-	d.SetId(projectId)
-
 	payload := client.PolicyUpdatePayload{}
 
 	if err := readResourceData(&payload, d); err != nil {
 		return diag.Errorf("schema resource data deserialization failed: %v", err)
 	}
 
+	// Validate if one is "inherit", the other must be too.
+	if (payload.MaxTtl == "inherit" || payload.DefaultTtl == "inherit") && payload.MaxTtl != payload.DefaultTtl {
```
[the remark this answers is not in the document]
avnerenv0: @chpl is that really the case? why?

avnerenv0 (approved):
@TomerHeber looks great.
Not sure about that "both must inherit" thing but it's not critical and if we decide to change it it can be done in a different PR.

env0/resource_project_policy.go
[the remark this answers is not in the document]
chpl: If user overrides only one, and then changes the org policy, we can have inconsistency.

TomerHeber:
> @TomerHeber looks great. Not sure about that "both must inherit" thing but it's not critical and if we decide to change it it can be done in a different PR.

Thanks @avnerenv0 !

PR #434, opened 2022-06-29
Chore: Use random names for all "team"s in harness tests

RLRabinowitz:
Right now some harness tests are using a static name for `env0_team` resource. That means that if any harness test fails, and the team still exists, running the same test on the same organization will fail.

### Solution
Let's make sure all harness' team resources are using a unique team name


GiliFaroEnv0 (approved):
Great catch!

RLRabinowitz:
[TF Provider IT fails again](https://trello.com/c/YqoBYMKX/6837-tf-provider-it-fails-again)

Issue #438, opened 2022-06-30
Add run-all support


Wassap124:
We've added the ability to run `terragrunt run-all`, which runs multiple modules simultaneously.

The structure of the output result has been changed (**only** when `isTerragruntRunAll` is provided) to be:
```json
[{
   "moduleName": "<some name>"
   "output": { }
}]
```

This is noted by the `isTerragruntRunAll` field on the blueprint / template.
The blueprint / template's `type` MUST be `terragrunt`, and the terragrunt `version` MUST be `0.28.1` or above.

[API Reference](https://env0.stoplight.io/docs/api/2b138939cbeec-create-template)

TomerHeber:
request:
```
{
  "name": "Terragrunt template for environment resource",
  "retry": {},
  "sshKeys": [],
  "isTerragruntRunAll": true,
  "terragruntVersion": "0.35.0",
  "terraformVersion": "0.15.1",
  "type": "terragrunt",
  "fileName": "",
  "repository": "https://github.com/env0/templates",
  "isGitHubEnterprise": false,
  "isGitLabEnterprise": false,
  "isBitbucketServer": false,
  "bitbucketClientKey": null,
  "gitlabProjectId": null,
  "isGitLab": false,
  "tokenName": "",
  "tokenId": "",
  "githubInstallationId": null,
  "revision": "",
  "path": "misc/null-resource",
  "projectIds": []
}
```

response:
```
{
  "repository": "https://github.com/env0/templates",
  "projectId": "ROOT",
  "terragruntVersion": "0.35.0",
  "path": "misc/null-resource",
  "authorId": "auth0|622629c1baf9800070484575",
  "createdAt": "2022-05-03T12:57:06.943Z",
  "isBitbucketServer": false,
  "isTerragruntRunAll": true,
  "name": "Terragrunt template for environment resource",
  "retry": {},
  "sshKeys": [],
  "isGitLab": false,
  "organizationId": "42f8d807-42c8-435c-be2d-97774758161f",
  "updatedAt": "2022-07-16T23:20:54.774Z",
  "terraformVersion": "0.15.1",
  "isGitLabEnterprise": false,
  "id": "b975ebdb-5ae3-4360-9aca-011c9e180e10",
  "isGitHubEnterprise": false,
  "type": "terragrunt",
  "projectIds": [],
  "author": {
    "created_at": "2022-03-07T15:50:25.582Z",
[...]

PR #443, opened 2022-07-14
Feat: refactoring for adding run-all support

TomerHeber:
### Solution

Some refactoring I wanted to do before working on #438.
(Instead of having one large PR).

Schema validation was moved to the correct place, instead of doing it before the API call.
Updated tests accordingly.

### DIFF
--- a/client/template.go
+++ b/client/template.go
@@ -6,6 +6,7 @@ package client
 
 import (
 	"errors"
+	"fmt"
 	"strconv"
 	"strings"
 )
@@ -54,7 +55,7 @@ type Template struct {
 	TokenId              string           `json:"tokenId,omitempty" tfschema:",omitempty"`
 	UpdatedAt            string           `json:"updatedAt"`
 	TerraformVersion     string           `json:"terraformVersion"`
-	TerragruntVersion    string           `json:"terragruntVersion,omitempty"`
+	TerragruntVersion    string           `json:"terragruntVersion,omitempty" tfschema:",omitempty"`
 	IsDeleted            bool             `json:"isDeleted,omitempty"`
 	BitbucketClientKey   string           `json:"bitbucketClientKey" tfschema:",omitempty"`
 	IsGithubEnterprise   bool             `json:"isGitHubEnterprise"`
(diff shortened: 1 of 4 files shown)


client/template.go
```diff
@@ -108,19 +109,33 @@ type VariablesFromRepositoryPayload struct {
...
 	if payload.Type != "terragrunt" && payload.TerragruntVersion != "" {
-		return Template{}, errors.New("can't define terragrunt version for non-terragrunt blueprint")
+		return errors.New("can't define terragrunt version for non-terragrunt template")
 	}
 	if payload.Type == "terragrunt" && payload.TerragruntVersion == "" {
-		return Template{}, errors.New("must supply Terragrunt version")
+		return errors.New("must supply terragrunt version")
+	}
+
+	if payload.Type == "cloudformation" && payload.FileName == "" {
```
TomerHeber, line 128: moved this validation from a different place (this is the correct place).

client/template.go
```diff
@@ -108,19 +109,33 @@ type VariablesFromRepositoryPayload struct {
 	Repository           string   `json:"repository"`
 }
 
-func (client *ApiClient) TemplateCreate(payload TemplateCreatePayload) (Template, error) {
```
TomerHeber, line 111: function not removed. diff mix-up.

client/template.go
```diff
@@ -149,18 +164,6 @@ func (client *ApiClient) TemplateDelete(id string) error {
 }
 
 func (client *ApiClient) TemplateUpdate(id string, payload TemplateCreatePayload) (Template, error) {
-	if payload.Name == "" {
```
TomerHeber, line 152: validation is done in the schema code and not in the API code.

client/template_test.go
```diff
@@ -220,59 +220,6 @@ var _ = Describe("Templates Client", func() {
 		})
 	})
 
-	Describe("Terragrunt version errors", func() {
```
TomerHeber, line 223: new tests added to the schema.

env0/resource_template.go
```diff
@@ -51,14 +51,8 @@ func templateCreatePayloadFromParameters(d *schema.ResourceData) (client.Templat
 	templateCreatePayloadRetryOnHelper(d, "deploy", &payload.Retry.OnDeploy)
 	templateCreatePayloadRetryOnHelper(d, "destroy", &payload.Retry.OnDestroy)
 
-	if payload.Type == "cloudformation" {
```
TomerHeber, line 54: added this to the above "validate" function.

env0/resource_template_test.go
```diff
@@ -57,6 +57,7 @@ func TestUnitTemplateResource(t *testing.T) {
 			},
 		},
 		Type:               "terragrunt",
+		TerragruntVersion:  "0.35.1",
```
TomerHeber, line 60: now the tests are working better. (They were supposed to fail without terragrunt_version, but didn't)

yaronya (approved):

PR #445, opened 2022-07-17
Feat: add run-all support

### LINKED ISSUE #438: Add run-all support

TomerHeber:
resolves #438 

### Solution
1. added the new boolean flag.
2. added unit tests.
3. added semver to go.mod for comparing versions.

### DIFF
--- a/client/template.go
+++ b/client/template.go
@@ -9,6 +9,8 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+
+	"github.com/Masterminds/semver/v3"
 )
 
 type TemplateRetryOn struct {
@@ -62,6 +64,7 @@ type Template struct {
 	IsBitbucketServer    bool             `json:"isBitbucketServer"`
 	IsSingleUse          bool             `json:"isSingleUse"`
 	FileName             string           `json:"fileName,omitempty" tfschema:",omitempty"`
+	IsTerragruntRunAll   bool             `json:"isTerragruntRunAll"`
 }
 
 type TemplateCreatePayload struct {
@@ -87,6 +90,7 @@ type TemplateCreatePayload struct {
 	IsBitbucketServer    bool             `json:"isBitbucketServer"`
 	IsSingleUse          bool             `json:"isSingleUse"`
 	FileName             string           `json:"fileName,omitempty"`
+	IsTerragruntRunAll   bool             `json:"isTerragruntRunAll"`
 }
 
 type TemplateAssignmentToProjectPayload struct {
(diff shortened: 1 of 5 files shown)


client/template.go
```diff
@@ -125,6 +129,21 @@ func (payload TemplateCreatePayload) Validate() error {
 		return errors.New("must supply terragrunt version")
 	}
 
+	if payload.IsTerragruntRunAll {
+		if payload.Type != "terragrunt" {
+			return errors.New(`can't set is_terragrunt_run_all to "true" for non-terragrunt template`)
+		}
+
+		c, _ := semver.NewConstraint(">= 0.28.1")
```
TomerHeber, line 137: I added semver library to compare versions and make sure the terragrunt version is above or equal 0.28.1

env0/resource_template_test.go
```diff
@@ -974,4 +982,24 @@ func TestUnitTemplateResource(t *testing.T) {
 
 		runUnitTest(t, testCase, func(mock *client.MockApiClientInterface) {})
 	})
+
+	t.Run("run all with outdated terragrunt version", func(t *testing.T) {
```
TomerHeber, line 986: This verifies the version comparision use-case (when lower than).

env0/resource_template_test.go
```diff
@@ -60,6 +60,7 @@ func TestUnitTemplateResource(t *testing.T) {
 		TerragruntVersion:  "0.35.1",
 		IsGitlabEnterprise: true,
 		TerraformVersion:   "0.15.1",
+		IsTerragruntRunAll: true,
```
TomerHeber, line 63: Updated the acceptance tests to check IsTerragruntRunAll.

env0/templates.go
```diff
@@ -208,6 +208,12 @@ func getTemplateSchema(templateType TemplateType) map[string]*schema.Schema {
 			Description: "the cloudformation file name. Required if the template type is cloudformation",
 			Optional:    true,
 		},
+		"is_terragrunt_run_all": {
```
TomerHeber, line 211: This is all that's required. The readResourceData and writeResourceData do all the work automatically.

env0/resource_template_test.go
sagilaufer1992: so now we don't have a non-run-all check? shouldn't it be added on top of a low version of terragrunt that should run as usual, a higher version (>=0.28.1) that isn't run-all that should work, the added low version that is run-all and should fail? total 4 cases (low-all, high-all, low-not-all, high-not-all). that seems to replace one case with another, without adding the counterpart.

sagilaufer1992 (commented):
Looks good to me. I think there is a missing test-case, please read my comment and make sure all the cases exist as I suppose they should.

env0/resource_template_test.go
TomerHeber: @sagilaufer1992 - by default IsTerragruntRunAll is false. Therefore, there are some testcases that run with that mode.
I can change one of them to a lower version terragrunt. That should cover all 4 use-cases.

sagilaufer1992: as we check versions as well, and testing shouldn't take long (unit testing, right?) I'd rather we'll check all 4 use-cases.

TomerHeber: 8c55b175bd3f06253fd8909710e9d8ee39dca700

TomerHeber: All 4 use-cases are tested.

gleeUpdatedTemplate - high-all
gitlabUpdatedTemplate - low-not-all
bitbucketUpdatedTemplate - high-not-all
"run all with outdated terragrunt version" - low-all (which is expected to fail).

TomerHeber: @sagilaufer1992

sagilaufer1992: was about to comment that I found all, and thank you. so extra thank you for making it easier for others as well 🙏

sagilaufer1992 (approved):
great job 👍

PR #446, opened 2022-07-18

TomerHeber:
### Solution
Part of the effort to support template-less environment #374.
Added readResouceDataEx to support a prefix use-case which is required for the effort.
Added a unit test.

### DIFF
--- a/env0/utils.go
+++ b/env0/utils.go
@@ -41,31 +41,28 @@ func intPtr(i int) *int {
 	return &i
 }
 
-// Extracts values from the resourcedata, and writes it to the interface.
-func readResourceData(i interface{}, d *schema.ResourceData) error {
+// Prepends prefix to the fieldName.
+func readResourceDataEx(prefix string, i interface{}, d *schema.ResourceData) error {
 	// TODO: add a mechanism that returns an error if fields were set in the resourceData but not in the struct.
 	// Blocked by: https://github.com/hashicorp/terraform-plugin-sdk/issues/910
 
 	val := reflect.ValueOf(i).Elem()
 	for i := 0; i < val.NumField(); i++ {
-		fieldName := val.Type().Field(i).Name
-		// Assumes golang is CamalCase and Terraform is snake_case.
-		// This behavior can be overrided be used in the 'tfschema' tag.
-		fieldNameSC := toSnakeCase(fieldName)
-		if resFieldName, ok := val.Type().Field(i).Tag.Lookup("tfschema"); ok {
-			if resFieldName == "-" {
-				continue
-			}
+		parsedField := getFieldName(val.Type().Field(i))
+		if parsedField.skip {
+			continue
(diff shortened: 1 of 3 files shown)


env0/utils_test.go
```diff
@@ -246,6 +246,30 @@ func TestReadByPointerNilCustomResourceData(t *testing.T) {
 	assert.Equal(t, params.Description, "description")
 }
 
+func TestReadResourceDataEx(t *testing.T) {
```
TomerHeber, line 249: This tests read resource data ex.

env0/utils.go
```diff
@@ -41,31 +41,28 @@ func intPtr(i int) *int {
...
 	for i := 0; i < val.NumField(); i++ {
-		fieldName := val.Type().Field(i).Name
-		// Assumes golang is CamalCase and Terraform is snake_case.
-		// This behavior can be overrided be used in the 'tfschema' tag.
-		fieldNameSC := toSnakeCase(fieldName)
-		if resFieldName, ok := val.Type().Field(i).Tag.Lookup("tfschema"); ok {
-			if resFieldName == "-" {
-				continue
-			}
+		parsedField := getFieldName(val.Type().Field(i))
```
TomerHeber, line 51: getFieldName covers more use-cases. Calling it instead.

sabasabo, line 51: Question:
Why here we get the field with
```val.Type().Field(i)```

While in line 61, we get the field with:
```field := val.Field(i)```

Is there a need for `.Type()`?

File: env0/utils.go [hunk 3]
```diff
@@ -41,31 +41,28 @@ func intPtr(i int) *int {
...
-			// 'resource' tag found. Override to tag value.
-			fieldNameSC = resFieldName
+		fieldName := parsedField.name
+		if prefix != "" {
+			fieldName = prefix + "." + fieldName
 		}
 
 		field := val.Field(i)
 
 		fieldType := field.Type()
```
sabasabo, line 63: Not related to this pr - so you can ignore this comment:
`fieldType` is defined in line 63 but is being used at line 90.
It might be better to move it closer to its usage

sabasabo (approved):
LGTM
Left a question

env0/utils.go
TomerHeber: I would think so (but yes agree it's confusing).
We don't want the field itself, but the metadata of the field.
I believe val.Type() returns the struct meta data.
With the metadata you can extract the field name.

In line 61 we're more interested in the field value.

https://pkg.go.dev/reflect#Type

I'll go over the code down the line and see if it can be done in a better way.

env0/utils.go
TomerHeber: It's a good point. I'll review the code down the line and see if I can improve a few things.
It's all covered with unit tests, so should be safe to try and make changes.

PR #449, opened 2022-07-22
Fix: organizationId() hides the errors

### LINKED ISSUE #448: organizationId() hides the errors
**Summary**:
Error not showing when calling `organizationId()`

Had an error for having multiple orgs When calling:
```
func (client *ApiClient) Organization() (Organization, error) {
```
The error was ignored here:
https://github.com/env0/terraform-provider-env0/blob/a5b9ad53bd151f07dc209fd039d4d9aedf46fb09/client/organization.go#L48-L50

**Repro steps**:
In a unitTest testing `client/organization.go`:
1. Call `client.organizationId()` 
2. Make `client.Organization()` return an error
3. See if `client.organizationId()`  returns the error (Currently not)

TomerHeber:
fixes #448 

### Solution
1. organizationId now returns error on failure.
2. Added a test that verified the use-case.

### DIFF
--- a/client/notification_test.go
+++ b/client/notification_test.go
@@ -1,6 +1,8 @@
 package client_test
 
 import (
+	"errors"
+
 	. "github.com/env0/terraform-provider-env0/client"
 	gomock "github.com/golang/mock/gomock"
 	"github.com/jinzhu/copier"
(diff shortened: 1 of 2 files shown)


client/notification_test.go
```diff
@@ -85,6 +87,25 @@ var _ = Describe("Notification Client", func() {
 				Expect(*createdNotification).To(Equal(mockNotification))
 			})
 		})
+
+		Describe("Failure - Multiple Organizations", func() {
```
TomerHeber, line 91: This is very annoying... but I can't test in organization_test.
organizationId() function is private and therefore can be tested (test is using a different package).
(In general tests shouldn't be using different package names, but at this stage it will be difficult to modify).

To work around that I picked a random API call that calls organizationId (E.g. NotificationCreate).
I mocked "/organizations" which is called by organizationId which is called by NotificationCreate.

This now tests the use-case... just a little ugly.

client/organization.go
```diff
@@ -46,7 +46,7 @@ func (client *ApiClient) organizationId() (string, error) {
 	}
 	organization, err := client.Organization()
 	if err != nil {
-		return "", nil
+		return "", err
```
TomerHeber, line 49: this is the fix.

client/notification_test.go
avnerenv0: can you please add that in a comment? So the next person who works on this will understand what this is doing here

avnerenv0 (approved):

client/notification_test.go
TomerHeber: @avnerenv0 - done:
https://github.com/env0/terraform-provider-env0/pull/449/commits/43991c6c76044c923a3098826408db478c69294f

PR #456, opened 2022-07-31
Fix: Auth0 throttling issue on large amount of resources

tomer-landesman:

Pleo experienced a rate limit error due to large amount of resources requesting authenticaiton from Auth0, and got throttled.

### Solution

add a request to our api @ `/organizations` in order to cache auth0 response before the resources requests fire.


env0/provider.go
```diff
@@ -127,7 +128,13 @@ func configureProvider(version string, p *schema.Provider) schema.ConfigureConte
 		if err != nil {
 			return nil, diag.Diagnostics{diag.Diagnostic{Severity: diag.Error, Summary: err.Error()}}
 		}
+		apiClient := client.NewApiClient(httpClient)
+		if apiClient == nil {
```
TomerHeber, line 132: Not required. apiClient will never be nil.

```
func NewApiClient(client http.HttpClientInterface) ApiClientInterface {
	return &ApiClient{
		http:                 client,
		cachedOrganizationId: "",
	}
}

```

env0/provider.go
```diff
@@ -127,7 +128,13 @@ func configureProvider(version string, p *schema.Provider) schema.ConfigureConte
...
+			return nil, diag.Diagnostics{diag.Diagnostic{Severity: diag.Error, Summary: "Failed to fetch organiztions."}}
+		}
 
-		return client.NewApiClient(httpClient), nil
+		apiClient.Organization()
```
TomerHeber, line 136: Since you're already calling this method.
You could set the `cachedOrganizationId` in `ApiClient` struct.

With that said, I do wonder how it will affect some acceptance tests. (I guess we'll wait and see).

env0/provider.go
```diff
@@ -127,7 +128,13 @@ func configureProvider(version string, p *schema.Provider) schema.ConfigureConte
...
+		log.Printf("organizations fetched to cache Auth0 API response.")
```
TomerHeber, line 134: prepend [INFO].

`log.Println("[INFO] Resolving api key by name: ", id)`

I do wonder if this should be a log message or instead just a comment in the code....

env0/provider.go
tomer-landesman: good thought :) I'll run test as soon as I know it fixes the issue, currently when testing I can still see multiple numbers of invocation of authorizer lambda

env0/provider.go
tomer-landesman: I logged only for my debugging, I can remove it

tomer-landesman: [8257409](https://github.com/env0/terraform-provider-env0/pull/456/commits/8257409ecc7aa4e95734b4fcd4bda992911f18d8)

env0/provider.go
tomer-landesman: [e137d6d](https://github.com/env0/terraform-provider-env0/pull/456/commits/e137d6d4ba10183216f6e9d2e64f5246732fcd2f)

env0/provider.go
tomer-landesman: @TomerHeber this one solve the issue I had, and also addresses your suggestion :)
[5049f75](https://github.com/env0/terraform-provider-env0/pull/456/commits/5049f757d2745b15506f444ad1998c7764e163ff)

tomer-landesman:
["warm the cache" in our Terraform provider](https://trello.com/c/lU6tgjbO/7107-warm-the-cache-in-our-terraform-provider)

client/organization.go
```diff
@@ -40,6 +40,10 @@ func (client *ApiClient) Organization() (Organization, error) {
 	return result[0], nil
 }
 
+func (client *ApiClient) OrganizationIdCaching() (string, error) {
```
TomerHeber, line 43: I'm starting to think we should just change organizationId() to OrganizationId()
(it should be easy to "replace all" in the IDE).

You will also need to re-generate go-mock:
`go generate ./...` (see README.md).

env0/provider.go
```diff
@@ -127,7 +127,13 @@ func configureProvider(version string, p *schema.Provider) schema.ConfigureConte
 		if err != nil {
 			return nil, diag.Diagnostics{diag.Diagnostic{Severity: diag.Error, Summary: err.Error()}}
 		}
+		apiClient := client.NewApiClient(httpClient)
 
-		return client.NewApiClient(httpClient), nil
+		// organizations fetched to cache Auth0 API response.
+		_, organizationError := apiClient.OrganizationIdCaching()
+		if organizationError != nil {
```
TomerHeber, line 134: in general error variables are named `err`.
This is the syntax you want to use:
```
if _, err := apiClient.OrganizationIdCaching(); err != nil {
   return nil, diag.Diagnostics{diag.Diagnostic{Severity: diag.Error, Summary: err.Error()}}
}
```

client/organization.go
tomer-landesman: I agree, wasn't confidant of doing that change but it felt required
[46d3f7a](https://github.com/env0/terraform-provider-env0/pull/456/commits/46d3f7acd4e5f8f146e04018b2ffbc25f81c672b)

env0/provider.go
tomer-landesman: thanks 
[156b5a5](https://github.com/env0/terraform-provider-env0/pull/456/commits/156b5a5da71d5a61a2537beeadda4af499586fa4)

TomerHeber (approved):
LGTM!

PR #458, opened 2022-08-01
Feat: add Data source for self-hosted-agent helm values

### LINKED ISSUE #450: Add Data source for self-hosted-agent helm values
The data source can be named `self_hosted_agent_values`
Is based on this API = https://developer.env0.com/docs/api/6101bf9d06325-get-agent-values-file-content
Should GET the values (I think it's just JSON) and put them in a format they can be passed to a [Helm Release resource](https://registry.terraform.io/providers/hashicorp/helm/latest/docs/resources/release) as the `values` prop (or as one of the values so the user can merge with other values)

TomerHeber:
resolves #450 

### Solution
Added the data resource.
Added acceptance tests.


env0/data_self_host_agent_values.go
```diff
@@ -0,0 +1,44 @@
...
+
+	agentKey := d.Get("agent_key").(string)
+
+	values, err := apiClient.AgentValues(agentKey)
+	if err != nil {
+		return diag.Errorf("could not get agent values: %v", err)
+	}
+
+	d.Set("values", values)
+	d.SetId(agentKey)
```
yaronya, line 41: Can't we use the `writeResourceSchema` util?

env0/provider.go
```diff
@@ -48,33 +48,34 @@ func Provider(version string) plugin.ProviderFunc {
...
+				"env0_notifications":            dataNotifications(),
+				"env0_module":                   dataModule(),
+				"env0_git_token":                dataGitToken(),
+				"env0_api_key":                  dataApiKey(),
+				"env0_agents":                   dataAgents(),
+				"env0_user":                     dataUser(),
+				"env0_cloud_credentials":        dataCloudCredentials(),
+				"env0_source_code_variables":    dataSourceCodeVariables(),
+				"env0_ip_ranges":                dataIpRanges(),
+				"env0_self_hosted_agent_values": dataSelfHostedAgentValues(),
```
yaronya, line 78: I wonder if we should name it `env0_self_hosted_agent_values` and not `env0_agent_values` (as we already named it `env0_agents` on another data resource). WDYT?

TomerHeber: you prefer it to be named env0_agent_values?
I'm okay either way.

yaronya: Yes. I'd keep it consistent with the existing `env0_agents` data resource 🙏

env0/data_self_host_agent_values.go
TomerHeber: In this case probably not. It's generally for structs. In this case, it's a string.
Therefore, there's no benefit in using writeResourceSchema (and also probably not work).

env0/provider.go
TomerHeber: sure!

TomerHeber: @yaronya - done!
b2c3075fa2dde61bc71d36f07d1a86312f32a402

yaronya (approved):

PR #462, opened 2022-08-03

TomerHeber:
### Solution

Updated resource_configuration_variable.go to use writeResourceData (still more work in this file).
Fixed some issues and edge-cases found during the work.

### DIFF
--- a/env0/resource_configuration_variable.go
+++ b/env0/resource_configuration_variable.go
@@ -203,26 +203,10 @@ func resourceConfigurationVariableRead(ctx context.Context, d *schema.ResourceDa
 		return ResourceGetFailure("configuration variable", d, err)
 	}
 
-	d.Set("name", variable.Name)
-	d.Set("description", variable.Description)
-	d.Set("value", variable.Value)
-	d.Set("is_sensitive", variable.IsSensitive)
-	d.Set("is_read_only", variable.IsReadOnly)
-	d.Set("is_required", variable.IsRequired)
-	d.Set("regex", variable.Regex)
-	if variable.Type != nil && *variable.Type == client.ConfigurationVariableTypeTerraform {
-		d.Set("type", "terraform")
-	} else {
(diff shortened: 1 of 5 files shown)


env0/resource_configuration_variable.go
```diff
@@ -203,26 +203,10 @@ func resourceConfigurationVariableRead(ctx context.Context, d *schema.ResourceDa
...
-	if variable.Type != nil && *variable.Type == client.ConfigurationVariableTypeTerraform {
-		d.Set("type", "terraform")
-	} else {
-		d.Set("type", "environment")
-	}
-	if variable.Schema != nil {
-		if len(variable.Schema.Enum) > 0 {
-			d.Set("enum", variable.Schema.Enum)
-		}
+	d.Set("type", "environment")
```
TomerHeber, line 206: This stayed. See the if-else in 213-217,

env0/utils.go
```diff
@@ -233,7 +233,7 @@ func writeResourceData(i interface{}, d *schema.ResourceData) error {
 			continue
 		}
 
-		if customField, ok := field.Interface().(CustomResourceDataField); ok {
+		if customField, ok := field.Interface().(CustomResourceDataField); ok && !field.IsNil() {
```
TomerHeber, line 236: edge case caused a crash in.

env0/utils.go
```diff
@@ -242,6 +242,12 @@ func writeResourceData(i interface{}, d *schema.ResourceData) error {
 
 		if fieldType.Kind() == reflect.Ptr {
 			if field.IsNil() {
+				if !parsedField.omitEmpty {
```
TomerHeber, line 245: Unless omitempty is set. (In the PR notice I added some additional omitempty to some of the struct).
If the value is nil, it makes sense to set the field value to nil. This will actually set it to the default value as defined in the schema.

yaronya (approved):

PR #463, opened 2022-08-04

TomerHeber:
…ation logic

### Solution

part of #376 

Refactored project resource and project data resource.

### DIFF
--- a/env0/resource_project.go
+++ b/env0/resource_project.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/env0/terraform-provider-env0/client"
 	"github.com/google/uuid"
-	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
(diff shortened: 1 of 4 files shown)


env0/resource_project.go
```diff
@@ -82,10 +70,12 @@ func resourceProjectRead(ctx context.Context, d *schema.ResourceData, meta inter
 
 	project, err := apiClient.Project(d.Id())
 	if err != nil {
-		return diag.Errorf("could not get project: %v", err)
+		return ResourceGetFailure("project", d, err)
```
TomerHeber, line 73: this should handle drift issues as well.

env0/resource_project.go
```diff
@@ -94,18 +84,16 @@ func resourceProjectUpdate(ctx context.Context, d *schema.ResourceData, meta int
 	apiClient := meta.(client.ApiClientInterface)
 
 	id := d.Id()
-	payload := client.ProjectCreatePayload{
-		Name:        d.Get("name").(string),
-		Description: d.Get("description").(string),
+	var payload client.ProjectCreatePayload
+
+	if err := readResourceData(&payload, d); err != nil {
```
GiliFaroEnv0, line 89: so from now on we first read the resource and deserialized it
and before write we serialize it?

GiliFaroEnv0 (commented):
@TomerHeber approve just a Q to confirm I understand correctly

env0/resource_project.go
TomerHeber: @GiliFaroEnv0 

readResourceData extracts from resourceData and write to the interface (E.g. `var payload client.ProjectCreatePayload` in this case).

writeResourceData extracts from the interface (a struct usually) and write to the resourceData).

Note:
For most cases this works well. But in some cases manual d.Get("...") or d.Set("...") is required.
These two functions use golang reflection.

GiliFaroEnv0 (approved):
Great job 👍

PR #468, opened 2022-08-08
Feat: add outputs attribute to the environment resources

### LINKED ISSUE #465: Add `outputs` attribute to the environment resources
In certain cases, you would like to pass outputs from one environment to another environment as an input.
The  `env0_environment` resource should have the `outputs` data of the environment as an output, both in the resource and the data resource.
Pay attention that this might be empty if the environment is still in progress.

TomerHeber:
resolves #465 

### Solution
Added the "output" field.
Updated the acceptance tests.


env0/resource_environment.go
```diff
@@ -91,6 +91,12 @@ func resourceEnvironment() *schema.Resource {
 				Description: "id of the last deployment",
 				Computed:    true,
 			},
+			"output": {
```
TomerHeber, line 94: named it "output" and not "outputs". This aligns with the API field name.

env0/resource_environment.go
```diff
@@ -200,11 +206,17 @@ func setEnvironmentSchema(d *schema.ResourceData, environment client.Environment
 		return fmt.Errorf("schema resource data serialization failed: %v", err)
 	}
 
-	if environment.LatestDeploymentLog != (client.DeploymentLog{}) {
+	if environment.LatestDeploymentLog.BlueprintId != "" {
```
TomerHeber, line 209: The old `!=` can no longer work because json.RawMessage was added as field. It cannot be compared.

env0/resource_environment_test.go
```diff
@@ -197,20 +202,32 @@ func TestUnitEnvironmentResource(t *testing.T) {
 
 				return format
 			}
+
 			formatResourceWithConfiguration := func(env client.Environment, variables []client.ConfigurationVariable) string {
+				output := "null"
+				if len(env.LatestDeploymentLog.Output) > 0 {
+					output = strings.ReplaceAll(string(env.LatestDeploymentLog.Output), `"`, `\"`)
```
TomerHeber, line 209: need to escape `"`

env0/data_environment.go
```diff
@@ -70,6 +70,11 @@ func dataEnvironment() *schema.Resource {
 				Description: "the last deployed revision",
 				Computed:    true,
 			},
+			"output": {
```
TomerHeber, line 73: need to be added to data resource as well.

yaronya (approved):

PR #469, opened 2022-08-10

TomerHeber:
### Solution

Part of #374.
Implements the create.
(Will do the update in the next PR).

### DIFF
--- a/client/environment.go
+++ b/client/environment.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"encoding/json"
 	"fmt"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
@@ -109,6 +110,42 @@ type EnvironmentCreate struct {
 	VcsCommandsAlias            string                `json:"vcsCommandsAlias"`
 }
 
+// When converted to JSON needs to be flattened. See custom MarshalJSON below.
+type EnvironmentCreateWithoutTemplate struct {
+	EnvironmentCreate EnvironmentCreate
+	TemplateCreate    TemplateCreatePayload
+}
+
+// The custom marshalJSON is used to return a flat JSON.
+func (create *EnvironmentCreateWithoutTemplate) MarshalJSON() ([]byte, error) {
+	// 1. Marshal to JSON both structs.
+	ecb, err := json.Marshal(&create.EnvironmentCreate)
+	if err != nil {
+		return nil, err
+	}
+	tcb, err := json.Marshal(&create.TemplateCreate)
+	if err != nil {
+		return nil, err
+	}
+
+	// 2. Unmarshal both JSON byte arrays to two maps.
+	var ecm, tcm map[string]interface{}
+	if err := json.Unmarshal(ecb, &ecm); err != nil {
+		return nil, err
+	}
+	if err := json.Unmarshal(tcb, &tcm); err != nil {
+		return nil, err
+	}
+
+	// 3. Merged the maps.
+	for k, v := range ecm {
+		tcm[k] = v
+	}
+
+	// 4. Marshal the merged map back to JSON.
+	return json.Marshal(tcm)
+}
+
 type EnvironmentUpdate struct {
 	Name                        string `json:"name,omitempty"`
 	RequiresApproval            *bool  `json:"requiresApproval,omitempty"`
@@ -162,6 +199,22 @@ func (client *ApiClient) EnvironmentCreate(payload EnvironmentCreate) (Environme
 	if err != nil {
 		return Environment{}, err
 	}
+	return result, nil
+}
+
+func (client *ApiClient) EnvironmentCreateWithoutTemplate(payload EnvironmentCreateWithoutTemplate) (Environment, error) {
+	var result Environment
+
+	organizationId, err := client.OrganizationId()
+	if err != nil {
+		return result, nil
+	}
+	payload.TemplateCreate.OrganizationId = organizationId
+
+	if err := client.http.Post("/environments/without-template", payload, &result); err != nil {
+		return result, err
+	}
+
 	return result, nil
 }
 
--- a/client/environment.go
+++ b/client/environment.go
@@ -117,7 +117,7 @@ type EnvironmentCreateWithoutTemplate struct {
 }
 
 // The custom marshalJSON is used to return a flat JSON.
-func (create *EnvironmentCreateWithoutTemplate) MarshalJSON() ([]byte, error) {
+func (create EnvironmentCreateWithoutTemplate) MarshalJSON() ([]byte, error) {
 	// 1. Marshal to JSON both structs.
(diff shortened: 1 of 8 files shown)


client/environment.go
```diff
@@ -109,6 +110,42 @@ type EnvironmentCreate struct {
 	VcsCommandsAlias            string                `json:"vcsCommandsAlias"`
 }
 
+// When converted to JSON needs to be flattened. See custom MarshalJSON below.
+type EnvironmentCreateWithoutTemplate struct {
```
TomerHeber, line 114: The struct was defined in a manner to support future changes to the template and environment create payloads.

client/environment.go
```diff
@@ -109,6 +110,42 @@ type EnvironmentCreate struct {
...
+	EnvironmentCreate EnvironmentCreate
+	TemplateCreate    TemplateCreatePayload
+}
+
+// The custom marshalJSON is used to return a flat JSON.
+func (create *EnvironmentCreateWithoutTemplate) MarshalJSON() ([]byte, error) {
```
TomerHeber, line 120: Since I wanted the support as explained above, I require a custom JSON marshal that merged the two structs to a single JSON.

client/environment_test.go
```diff
@@ -331,3 +386,29 @@ var _ = Describe("Environment Client", func() {
 		})
 	})
 })
+
+func TestMarshalEnvironmentCreateWithoutTemplate(t *testing.T) {
```
TomerHeber, line 390: This is a basic test for testing the marhsal.

client/template.go
```diff
@@ -62,7 +62,6 @@ type Template struct {
 	BitbucketClientKey   string           `json:"bitbucketClientKey" tfschema:",omitempty"`
 	IsGithubEnterprise   bool             `json:"isGitHubEnterprise"`
 	IsBitbucketServer    bool             `json:"isBitbucketServer"`
-	IsSingleUse          bool             `json:"isSingleUse"`
```
TomerHeber, line 65: not required.

env0/resource_environment.go
```diff
@@ -182,6 +183,17 @@ func resourceEnvironment() *schema.Resource {
 					},
 				},
 			},
+			"without_template_settings": {
```
TomerHeber, line 186: couldn't think of a better name...

env0/resource_environment_test.go
```diff
@@ -949,8 +950,7 @@ func TestUnitEnvironmentResource(t *testing.T) {
 	testApiFailures()
 }
 
-/* TODO
```
TomerHeber, line 952: uncommented the test and made some changes.
Will update the test in the next PR to test update as well.

env0/resource_template.go
```diff
@@ -23,7 +23,7 @@ var allowedTemplateTypes = []string{
 	"cloudformation",
 }
 
-func resourceTemplate() *schema.Resource {
+func getTemplateSchema(prefix string) map[string]*schema.Schema {
```
TomerHeber, line 26: Not a big change here (just looks scary).
The schema varies a little bit between the template use-case and an environment-without-template use-case.
In this way, if more fields are added to the template resource, no work is needed.

client/environment_test.go
```diff
@@ -212,6 +222,51 @@ var _ = Describe("Environment Client", func() {
...
+			httpCall.Times(1)
+		})
+
+		It("Should not return error", func() {
+			Expect(err).To(BeNil())
+		})
+
+		It("Should return the created environment", func() {
+			Expect(createdEnvironment).To(Equal(mockEnvironment))
+		})
```
ItamarMalka, line 267: I think we are missing a was template creation called assertion here

ItamarMalka (commented):
Looks good 👍 left a few minor comments

client/environment.go
ItamarMalka: embedding those like [this](https://stackoverflow.com/a/24642709) won't flatten this out? (instead of the marshel -> unmarshel flatten) or maybe this is a restriction by the env without template struct you receive?

client/environment_test.go
```diff
@@ -331,3 +386,29 @@ var _ = Describe("Environment Client", func() {
...
+	environmentCreateWithoutTemplate := EnvironmentCreateWithoutTemplate{
+		EnvironmentCreate: environmentCreate,
+		TemplateCreate:    templateCreate,
+	}
+
+	b, err := json.Marshal(environmentCreateWithoutTemplate)
+	require.NoError(t, err)
+	s := string(b)
+	require.True(t, strings.Contains(s, `"sshKeys":[{"id":"id1","name":"name1"}]`))
+	require.True(t, strings.Contains(s, `"name":"name"`))
```
ItamarMalka, line 412: this test looks odd, as it doesn't really assert that the flattening works (as those strings should also be present for the nested form of it)

env0/resource_environment_test.go
```diff
@@ -1034,7 +1034,7 @@ func TestUnitTemplatelessEnvironmentResource(t *testing.T) {
 			terragrunt_working_directory = "%s"
 			force_destroy = true
 			vcs_commands_alias = "%s"
-			template {
+			without_template_settings {
 				repository = "%s"
 				terraform_version = "%s"
```
ItamarMalka, line 1039: probably an out of scope question but are we only allowing tf templates here no tg/pulumi etc...

client/environment_test.go
TomerHeber: @ItamarMalka - we don't call the create template.
The API call creates both environment and template in the same API call.

client/environment_test.go
TomerHeber: @ItamarMalka 

ok. Let me improve it.
I'll send an update shortly.

env0/resource_environment_test.go
TomerHeber: @ItamarMalka - this schema is equivalent (with some minor changes) to the template creation schema.
In other words, everything should be supported.

client/environment_test.go
TomerHeber: @ItamarMalka - good catch it actually found a bug (after changing the test).
This is now fixed. 

d2f17f686e17d0164db78efb1bf7668e7bf75ee8

ItamarMalka: what about [this](https://github.com/env0/terraform-provider-env0/pull/469#discussion_r943600924) for unmarshaling this won't work?

client/environment.go
TomerHeber: It will not work because of some conflicts between the fields.
E.g. "Name".
In case of a conflict golang just ignores the field. (Therefore I would need to create a custom MarshalJSON any how).
I found this a little nicer and intuitive (The struct I mean).

client/environment_test.go
TomerHeber: @ItamarMalka - sorry I missed your question. Responded in the other thread.

ItamarMalka (approved):

PR #473, opened 2022-08-12
Feat: the environment configuration section is missing some attributes

### LINKED ISSUE #472: The environment configuration section is missing some attributes
In the `configuration` section in the `env0_environment` resource, we are missing some attributes compared to the `env0_configuration_variable` resource. We are missing the following:

- `format`
- `enum`
- `is_read_only`
- `is_required`
- `regex`

TomerHeber:
resolves #472 

### Solution

Added the missing fields and updated acceptance tests.

### DIFF
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -180,6 +180,23 @@ func resourceEnvironment() *schema.Resource {
 							Optional:     true,
 							ValidateFunc: ValidateConfigurationPropertySchema,
 						},
+						"is_read_only": {
+							Type:        schema.TypeBool,
+							Description: "is the variable read only",
+							Optional:    true,
+							Default:     false,
+						},
+						"is_required": {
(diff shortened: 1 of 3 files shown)


env0/resource_environment.go
```diff
@@ -232,9 +249,18 @@ func setEnvironmentConfigurationSchema(d *schema.ResourceData, configurationVari
 		if configurationVariable.Description != "" {
 			variable["description"] = configurationVariable.Description
 		}
+		if configurationVariable.Regex != "" {
```
TomerHeber, line 252: envrionment is a bit of a mess...
The only that does not use readResource and writeResource.

razbensimon: > The only that does not use readResource and writeResource.

because?

razbensimon (approved):
few questions but looks good

PR #474, opened 2022-08-14
Fix: CD and PR plan settings on environment are not applied correctly

### LINKED ISSUE #471: CD and PR plan settings on environment are not applied correctly
Step to reproduce:
1. Create a new project.
2. Go to the project settings  -> policies.
3. Make sure that in the Continuous Deployment section both `Run Terraform plan on Pull Requests` and `Redeploy on every push` are enabled.
4. Create a new environment using the terraform provider and set the `deploy_on_push` to false and `run_plan_on_pull_requests ` to false.
5. Once the environment is created, run another `terraform plan` without changing anything. 
6. The plan will say it's going to change the `deploy_on_push` and the `run_plan_on_pull_requests ` from `true` to `false
7. You can also check the UI and see that under the environment in the settings tab that both PR plan and CD are enabled.

TomerHeber:
fixes #471 

### Solution

Using GetOkExist.
Updarted unit tests.

The bug was that HasChange behaves the same for field not set and false.
The only way around it is to use GetOkExist. The supports: true, false, and unset. (All 3 modes are required).

### DIFF
--- a/env0/resource_environment_test.go
+++ b/env0/resource_environment_test.go
@@ -80,6 +80,8 @@ func TestUnitEnvironmentResource(t *testing.T) {
 							resource.TestCheckResourceAttr(accessor, "vcs_commands_alias", environment.VcsCommandsAlias),
 							resource.TestCheckResourceAttr(accessor, "revision", environment.LatestDeploymentLog.BlueprintRevision),
 							resource.TestCheckResourceAttr(accessor, "output", string(updatedEnvironment.LatestDeploymentLog.Output)),
+							resource.TestCheckNoResourceAttr(accessor, "deploy_on_push"),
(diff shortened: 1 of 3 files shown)


env0/resource_environment_test.go
```diff
@@ -601,11 +605,10 @@ func TestUnitEnvironmentResource(t *testing.T) {
 					BlueprintId: "template-id",
 				},
 
-				AutoDeployOnPathChangesOnly: &truthyFruity,
```
TomerHeber, line 604: Switched order... start with false, and update to true.

TomerHeber: initially without the fixes, false would be ignored as not set.

PR #477, opened 2022-08-16
Chore: Fix flaky tests that use template project assignment resource

ghost:
On some occasions, when we run the tests they'll fail because the environment is created before `env0_template_project_assignment` is created. It happens because the environment is not dependent on the assignment.

### Solution
Added `depends_on` on relevant tests to prevent race condition of TF creating environment before associating the template to the correct project.


GiliFaroEnv0 (approved):
Nice catch 😄 
LGTM

PR #479, opened 2022-08-16

TomerHeber:
…ation logic

### Solution

Part of #376 refactoring to use read/write serializer.


RLRabinowitz (approved):
:+1

PR #481, opened 2022-08-16
Feat: fix is_self_hosted attribute of organization resource

### LINKED ISSUE #480: Fix `is_self_hosted` attribute of organization resource
We've deprecated the unused `isSelfHosted` attribute of the organization API.

The new value, `isSelfHostedK8s` should be the one that indicated whether resource/data resource is `is_self_hosted`.
I believe this is a simple change using the `tfschema` tag we've introduced.

TomerHeber:
resolves #480 

### Solution
Modified to use IsSelfHostedK8s instead of IsSelfHosted.
Updated acceptance test.

### DIFF
--- a/client/organization.go
+++ b/client/organization.go
@@ -17,8 +17,8 @@ type Organization struct {
 	CreatedAt                           string  `json:"createdAt"`
 	UpdatedAt                           string  `json:"updatedAt"`
 	Role                                string  `json:"role"`
-	IsSelfHosted                        bool    `json:"isSelfHosted"`
-	IsSelfHostedK8s                     bool    `json:"isSelfHostedK8s"`
+	IsSelfHosted                        bool    `json:"isSelfHosted" tfschema:"-"`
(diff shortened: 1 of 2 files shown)


client/organization.go
```diff
@@ -17,8 +17,8 @@ type Organization struct {
 	CreatedAt                           string  `json:"createdAt"`
 	UpdatedAt                           string  `json:"updatedAt"`
 	Role                                string  `json:"role"`
-	IsSelfHosted                        bool    `json:"isSelfHosted"`
-	IsSelfHostedK8s                     bool    `json:"isSelfHostedK8s"`
+	IsSelfHosted                        bool    `json:"isSelfHosted" tfschema:"-"`
```
yaronya, line 20: IsSelfHosted no longer exists in API, it should be removed

TomerHeber: removed.

yaronya (approved):

PR #482, opened 2022-08-18
Fix: TF module name uniqueness in integration tests

yaronya:
Integration tests might fail due to having a non-random env0 module resource name

### Solution
Randomize it


roni-frantchi (approved):
# 🚀

PR #484, opened 2022-08-18
Fix: make sure all resource names are unique in integration tests

### LINKED ISSUE #483: Make sure all resource names are unique in integration tests
We recently[ found a non-unique resource name](https://github.com/env0/terraform-provider-env0/pull/482) once again which might fail our integration tests.

We need to go over all resource types again and make sure they're unique using a random string

TomerHeber:
fixes #483 

### Solution

Updated the integration tests.

### DIFF
--- a/tests/integration/002_project/main.tf
+++ b/tests/integration/002_project/main.tf
@@ -1,9 +1,9 @@
 provider "random" {}
 
 resource "random_string" "random" {
-  length    = 5
+  length    = 8
   special   = false
-  min_lower = 5
+  min_lower = 8
 }
 
 resource "env0_project" "test_project" {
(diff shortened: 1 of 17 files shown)


sagilaufer1992 (approved):
Looks great.
Can you by any chance make sure all the random strings across all the tests will be generated the same way? For example, in test 2 we use length 5. it's still a VERY low probability for conflicts, but for the conformity, make them all 8 (or higher if some other test is higher...)

TomerHeber:
> Looks great. Can you by any chance make sure all the random strings across all the tests will be generated the same way? For example, in test 2 we use length 5. it's still a VERY low probability for conflicts, but for the conformity, make them all 8 (or higher if some other test is higher...)

@sagilaufer1992 - sure! I'll make sure all are 8.

PR #486, opened 2022-08-21
Fix: env0_template_project_assignment resource is racing with env0_en…

### LINKED ISSUE #478: `env0_template_project_assignment` resource is racing with `env0_environment` resource if they exist in the same file
Because the `env0_environment` resource does not depend on `env0_template_project_assignment` if they are created in the same file ([like this test)](https://github.com/env0/terraform-provider-env0/blob/d89bd7d08f3709aa746cc5abda708e0861846927/tests/integration/012_environment/main.tf), their creation order will be random. This is an issue because in a future fix we'll not allow new environments to be created with templates whose project Id is not matching to that environment's project Id.
So the `env0_template_project_assignment` must be created before `env0_environment`, or our users will receive an error.

Two possible solutions I had in mind:
1. We can only mention it in our documentation, and make sure that the `env0_environment` has a `depends_on` to the `env0_template_project_assignment` resource ([example here](https://github.com/env0/terraform-provider-env0/blob/main/tests/integration/012_environment/main.tf#L28)). We should mention it [here in this example](https://github.com/env0/terraform-provider-env0/blob/main/examples/resources/env0_environment/resource.tf) (right now it missing the `env0_template_project_assignment` resource).
2. Add the `env0_template_project_assignment` as a subschema to the `env0_template` resource, just like `configuration` is used in the `env0_environment` resource. I don't know if it's possible.

TomerHeber:
…vironment resource if they exist in the same file

fixes #478 

### Solution

Updated documents and error messages with additional details.

### DIFF
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -42,7 +42,7 @@ func resourceEnvironment() *schema.Resource {
 			},
 			"template_id": {
 				Type:         schema.TypeString,
-				Description:  "the template id the environment is to be created from",
+				Description:  "the template id the environment is to be created from. Important note: the template must first be assigned to the same project as the environment (project_id). Use 'env0_template_project_assignment' to assign the template to the project. In addition, be sure to leverage 'depends_on' if applicable.",
 				Optional:     true,
 				ForceNew:     true,
 				ExactlyOneOf: []string{"without_template_settings", "template_id"},
@@ -311,7 +311,7 @@ func resourceEnvironmentCreate(ctx context.Context, d *schema.ResourceData, meta
 		environment, err = apiClient.EnvironmentCreateWithoutTemplate(payload)
 	}
 	if err != nil {
-		return diag.Errorf("could not create environment: %v", err)
+		return diag.Errorf(`could not create environment: %v\nNote: if "template_id" is used make sure the template is first assigned to the project.`, err)
 	}
 	environmentConfigurationVariables := client.ConfigurationChanges{}
 	if environmentPayload.DeployRequest.ConfigurationChanges != nil {
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -286,6 +286,23 @@ func setEnvironmentConfigurationSchema(d *schema.ResourceData, configurationVari
 	}
 }
 
+// Validate that the template is assigned to the "project_id".
+func validateTemplateProjectAssignment(d *schema.ResourceData, apiClient client.ApiClientInterface) error {
+	projectId := d.Get("project_id").(string)
+	templateId := d.Get("template_id").(string)
+
+	template, err := apiClient.Template(templateId)
+	if err != nil {
+		return fmt.Errorf("could not get template: %v", err)
+	}
+
+	if projectId != template.ProjectId && !stringInSlice(projectId, template.ProjectIds) {
+		return errors.New("could not create environment: template is not assigned to project")
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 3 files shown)


env0/resource_environment.go
```diff
@@ -42,7 +42,7 @@ func resourceEnvironment() *schema.Resource {
 			},
 			"template_id": {
 				Type:         schema.TypeString,
-				Description:  "the template id the environment is to be created from",
+				Description:  "the template id the environment is to be created from. Important note: the template must first be assigned to the same project as the environment (project_id). Use 'env0_template_project_assignment' to assign the template to the project. In addition, be sure to leverage 'depends_on' if applicable.",
```
ItamarMalka, line 45: Maybe worth '/n' stuff here to make it a a bit more comfortable

ItamarMalka (approved):
looks great 👍

ItamarMalka (commented):
@TomerHeber Just had a look at the issue this should fix, 
can't this be asserted to see that if we have a `env0_template_project_assignment ` resource it's `env0_environment` depends on it?

TomerHeber:
> @TomerHeber Just had a look at the issue this should fix, can't this be asserted to see that if we have a `env0_template_project_assignment ` resource it's `env0_environment` depends on it?

@ItamarMalka 
Sure! Do you want me to query in case of an error and only show the explanation if the query shows that the assignment is missing?

env0/resource_environment.go
```diff
@@ -297,7 +314,10 @@ func resourceEnvironmentCreate(ctx context.Context, d *schema.ResourceData, meta
 	var environment client.Environment
 	var err error
 
-	if d.Get("template_id") != "" {
+	if d.Get("template_id").(string) != "" {
+		if err := validateTemplateProjectAssignment(d, apiClient); err != nil {
```
TomerHeber, line 318: @ItamarMalka - added validation before trying to create the environment.

env0/resource_environment.go
```diff
@@ -42,7 +42,7 @@ func resourceEnvironment() *schema.Resource {
 			},
 			"template_id": {
 				Type:         schema.TypeString,
-				Description:  "the template id the environment is to be created from",
+				Description:  "the template id the environment is to be created from.\nImportant note: the template must first be assigned to the same project as the environment (project_id). Use 'env0_template_project_assignment' to assign the template to the project. In addition, be sure to leverage 'depends_on' if applicable.",
```
TomerHeber, line 45: @ItamarMalka 
added '\n'

TomerHeber:
> @TomerHeber Just had a look at the issue this should fix, can't this be asserted to see that if we have a `env0_template_project_assignment ` resource it's `env0_environment` depends on it?

@ItamarMalka - updated the PR.

env0/resource_environment.go
```diff
@@ -286,6 +286,23 @@ func setEnvironmentConfigurationSchema(d *schema.ResourceData, configurationVari
 	}
 }
 
+// Validate that the template is assigned to the "project_id".
+func validateTemplateProjectAssignment(d *schema.ResourceData, apiClient client.ApiClientInterface) error {
+	projectId := d.Get("project_id").(string)
+	templateId := d.Get("template_id").(string)
+
+	template, err := apiClient.Template(templateId)
```
ItamarMalka, line 294: so in case of the race condition described [here](https://github.com/env0/terraform-provider-env0/issues/478) if we get here and we didn't assign the template yet it would throw an exception here?

TomerHeber: @ItamarMalka 
I don't think so. The template must exist irrespective of the project assignment.

PR #488, opened 2022-08-23
Fix: diff issues with environment configuration variables

### LINKED ISSUE #459: lexicographical sorting of configuration items
I am trying to define my environment's configuration options using a map and a dynamic block.  I am doing this to make the terraform file much more readable and easier to maintain.  

Here is an example
```
locals {
  db-config = {
    allocated_storage = "40"
    db_instance_class = "db.t3.small"
    db_user           = "dbuser"
    service           = "myservice"
  }
}
resource "env0_environment" "automatron-db" {
  name       = "testdb"
  project_id = data.env0_project.this.id

  template_id                = data.env0_template.rds.id
  deploy_on_push             = true
  run_plan_on_pull_requests  = true
  vcs_commands_alias         = "test-db"

  dynamic "configuration" {
    for_each = local.db-config
    content {
      name        = configuration.key
      value       = configuration.value
      type        = "terraform"
      schema_type = "string"
    }
  }
}

```

This code will apply and works great.  I can then create separate dynamic configuration blocks if I need to have a different type.  I find it much easier to read and adjust the settings compared to having each configuration setting block defined separately.

The problem is that future plans regularly want to update all of the configuration blocks that the dynamic block had created.  What I have determined is that the provider gets the options in a different order then what the map is in.  Terraform maps are automatically sorted in lexicographical order.  Its worth noting that if I look at the configuration options using terraform state show they are also listed in lexicographical order.  But when I run terraform plan, they seem to be returned in a different order.

TomerHeber:
### Solution

Found a few bugs that may cause the issue observed.
May resolve the issue described in #459

### DIFF
--- a/client/configuration_variable.go
+++ b/client/configuration_variable.go
@@ -24,7 +24,7 @@ const (
 )
 
 type ConfigurationVariableSchema struct {
-	Type   string   `json:"type"`
+	Type   string   `json:"type,omitempty"`
 	Enum   []string `json:"enum"`
 	Format Format   `json:"format,omitempty"`
 }
--- a/env0/resource_environment_test.go
+++ b/env0/resource_environment_test.go
@@ -252,7 +252,6 @@ func TestUnitEnvironmentResource(t *testing.T) {
 				Name:  configurationVariables.Name,
 				Type:  &newVarType,
 				Schema: &client.ConfigurationVariableSchema{
-					Type:   "string",
 					Format: client.Text,
 				},
 				Regex: "regex2",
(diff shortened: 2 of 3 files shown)


env0/resource_environment_test.go
```diff
@@ -348,6 +347,149 @@ func TestUnitEnvironmentResource(t *testing.T) {
 			})
 		})
 
+		t.Run("Create configuration variables - default values", func(t *testing.T) {
```
TomerHeber, line 350: added a couple of tests that fail without the fix.
(plan diff).

client/configuration_variable.go
```diff
@@ -24,7 +24,7 @@ const (
 )
 
 type ConfigurationVariableSchema struct {
-	Type   string   `json:"type"`
+	Type   string   `json:"type,omitempty"`
```
TomerHeber, line 27: just to avoid sending "" which may confuse the API.

GiliFaroEnv0 (approved):
Wow nice catch!
don't know how it didn't pop sooner

PR #491, opened 2022-08-31
Fix: on environment redeploy, BLUEPRINT and GLOBAL scope variables af…

### LINKED ISSUE #490: Bug: On environment redeploy, BLUEPRINT and GLOBAL scope variables affect `configurationChanges`
When an environment has a BLUEPRINT scope variable with name "MY_VAR", trying to set `configuration` block on an `env0_environment` resource with the same name "MY_VAR" could corrupt that BLUEPRINT scope variable.

Source of the issue is [getUpdateConfigurationVariables](https://github.com/env0/terraform-provider-env0/blob/d720f0951b2a63c22c3fbc6a0218ccf6aced4207/env0/resource_environment.go#L631-L639) which is called to fetch current ENVIRONMENT variables and remove/link with the existing `configuration` block. `ConfigurationVariablesByScope` also returns BLUEPRINT and GLOBAL scope variables. If there's a BLUEPRINT scope with the same name, then in the `deploy` API we send `configurationChanges` with BLUEPRINT scope, instead of DEPLOYMENT scope. This issue only happens on redeploy

The result of `ConfigurationVariablesByScope` should be filtered, we should validate that we only take ENVIRONMENT scope variables into account

Steps to reproduce:
1. Create a template, and give it a variable MY_VARIABLE (Either in TF provider or outside of it)
2. In TF Provider, create `env0_environment` resource with same template, without anything in `configuration` block
3. `tf apply`
4. Add a `configuration` block to the `env0_environment` resource, with `name = "MY_VARIABLE"`
5. TF apply -> The `POST environments/{id}/deployments` API call is sent with corrupt data - `configurationChanges` has a `BLUEPRINT` scope

TomerHeber:
…fect configurationChanges

fixes #490 

### Solution
Filter out template/blueprint in all relevant API calls.

### DIFF
--- a/client/configuration_variable.go
+++ b/client/configuration_variable.go
@@ -85,6 +85,16 @@ func (client *ApiClient) ConfigurationVariablesById(id string) (ConfigurationVar
 	return result, nil
 }
 
+func filterOutConfigurationVariables(variables []ConfigurationVariable, scope Scope) []ConfigurationVariable {
+	filteredVariables := []ConfigurationVariable{}
+	for _, variable := range variables {
+		if variable.Scope != scope {
+			filteredVariables = append(filteredVariables, variable)
+		}
+	}
+	return filteredVariables
+}
+
 func (client *ApiClient) ConfigurationVariablesByScope(scope Scope, scopeId string) ([]ConfigurationVariable, error) {
 	organizationId, err := client.OrganizationId()
 	if err != nil {
@@ -110,15 +120,12 @@ func (client *ApiClient) ConfigurationVariablesByScope(scope Scope, scopeId stri
 		return []ConfigurationVariable{}, err
 	}
 
+	// The API returns global and template scopes for environment (and other) scopes. Filter them out.
 	if scope != ScopeGlobal {
-		// Filter out global scopes. If a non global scope is requested.
-		filteredResult := []ConfigurationVariable{}
-		for _, variable := range result {
-			if variable.Scope != ScopeGlobal {
-				filteredResult = append(filteredResult, variable)
-			}
+		result = filterOutConfigurationVariables(result, ScopeGlobal)
+		if scope != ScopeTemplate {
+			result = filterOutConfigurationVariables(result, ScopeTemplate)
 		}
-		return filteredResult, nil
 	}
 
 	return result, nil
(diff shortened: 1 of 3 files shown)


env0/resource_environment.go
```diff
@@ -628,8 +628,25 @@ func getTTl(date string) client.TTL {
 	}
 }
 
+func getConfigurationVariables(apiClient client.ApiClientInterface, environmentId string) ([]client.ConfigurationVariable, error) {
```
TomerHeber, line 631: created a common function. called from 3 places.

env0/resource_environment.go
```diff
@@ -628,8 +628,25 @@ func getTTl(date string) client.TTL {
...
+	variables, err := apiClient.ConfigurationVariablesByScope(client.ScopeEnvironment, environmentId)
+	if err != nil {
+		return nil, err
+	}
+
+	// Filter out template/blueprint variables.
+	filteredVariables := []client.ConfigurationVariable{}
+	for _, variable := range variables {
+		if variable.Scope != client.ScopeTemplate {
```
TomerHeber, line 640: global is already filtered out in the API call.

RLRabinowitz: Blueprint as the same as global (kinda), if we fetch by Environment or Deployment scope, then we might get back Blueprint scope variables.

I think this filter logic should be in the API call itself

RLRabinowitz (commented):
I think this filtering logic should be added to the API as well. 
I think this also affects the `data env0_configuration_variable`. If the scope is Environment there you might get a blueprint scope variable

client/configuration_variable.go
```diff
@@ -110,15 +120,12 @@ func (client *ApiClient) ConfigurationVariablesByScope(scope Scope, scopeId stri
 		return []ConfigurationVariable{}, err
 	}
 
+	// The API returns global and template scopes for environment (and other) scopes. Filter them out.
```
TomerHeber, line 123: global and blueprint filtering is done in the API.

env0/resource_environment.go
TomerHeber: @RLRabinowitz - modified to the API call itself.

client/configuration_variable.go
TomerHeber: if not global - remove global...
if not global and not template,  remove template as well.

yaronya (commented):
@TomerHeber can we assert it with a unit test please?

TomerHeber:
> @TomerHeber can we assert it with a unit test please?

@yaronya  108658c899ffc51460f1cbbc66c779f9a24c2797

yaronya (approved):

PR #495, opened 2022-09-06
Fix - 401 errors from auth0 sue to cache misses when using user/pass

eladmosh:
We're seeing an elevated rate of throttling errors  in the `tf_provder` tests as we do not benefit from cache when using user/password.

### Solution
Use user/password once to obtain JWT and use that on following API calls.

### DIFF
--- a/client/http/client.go
+++ b/client/http/client.go
@@ -3,8 +3,6 @@ package http
 //go:generate mockgen -destination=client_mock.go -package=http . HttpClientInterface
 
 import (
-	"fmt"
-
 	"github.com/go-resty/resty/v2"
 )
 
@@ -38,11 +36,14 @@ func NewHttpClient(config HttpClientConfig) (*HttpClient, error) {
 		ApiSecret: config.ApiSecret,
 		client:    config.RestClient.SetHostURL(config.ApiEndpoint).SetHeader("User-Agent", config.UserAgent),
 	}
+	var res string
 	req := httpClient.client.R().SetBasicAuth(httpClient.ApiKey, httpClient.ApiSecret)
+	response, err := req.SetQueryParams(map[string]string{"encoded": "true"}).SetResult(&res).Get("auth/token")
+	if err != nil {
+		return nil, err
+	}
 
-	req.SetQueryParams(nil).SetResult(httpClient.jwtToken).Get("auth/token?encoded=true")
-	fmt.Println("***", httpClient.jwtToken)
-
+	httpClient.jwtToken = string(response.Body())
 	return httpClient, nil
 }
 
--- a/client/http/client.go
+++ b/client/http/client.go
@@ -51,7 +51,6 @@ func NewHttpClient(config HttpClientConfig) (*HttpClient, error) {
 	return httpClient, err
 }
 
-
 func (client *HttpClient) request() *resty.Request {
 	return client.client.R().SetAuthToken(client.jwtToken)
 }
--- a/client/http/client.go
+++ b/client/http/client.go
@@ -17,7 +17,7 @@ type HttpClientInterface interface {
 }
 
 type HttpClient struct {
-	jwtToken    string
+	jwtToken  string
 	ApiKey    string
 	ApiSecret string
 	Endpoint  string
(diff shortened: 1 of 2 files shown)


client/http/client_test.go
```diff
@@ -29,7 +29,7 @@ type RequestBody struct {
 const BaseUrl = "https://fake.env0.com"
 const ApiKey = "MY_USER"
 const ApiSecret = "MY_PASS"
-const ExpectedBasicAuth = "Basic TVlfVVNFUjpNWV9QQVNT"
+const ExpectedBasicAuth = "Bearer mockedJwtToken"
```
avnerenv0, line 32:
```suggestion
const ExpectedJWTAuth = "Bearer mockedJwtToken"
```

client/http/client.go
```diff
@@ -30,15 +31,24 @@ type HttpClientConfig struct {
...
-	return &HttpClient{
+	httpClient := &HttpClient{
 		ApiKey:    config.ApiKey,
 		ApiSecret: config.ApiSecret,
 		client:    config.RestClient.SetHostURL(config.ApiEndpoint).SetHeader("User-Agent", config.UserAgent),
-	}, nil
+	}
+	var res string
+	req := httpClient.client.R().SetBasicAuth(httpClient.ApiKey, httpClient.ApiSecret)
+	response, err := req.SetQueryParams(map[string]string{"encoded": "true"}).SetResult(&res).Get("auth/token")
```
avnerenv0, line 41: why do we `.SetResult(&res)` if we use `response` for the result?

avnerenv0: I see that fixed the tests also

client/http/client_test.go
eladmosh: [changing var name](https://github.com/env0/terraform-provider-env0/pull/495/commits/6e57eea77dcaa2080897d841e812d33526caefd5)

avnerenv0 (approved):
👍

client/http/client.go
```diff
@@ -30,15 +31,23 @@ type HttpClientConfig struct {
...
 		ApiKey:    config.ApiKey,
 		ApiSecret: config.ApiSecret,
 		client:    config.RestClient.SetHostURL(config.ApiEndpoint).SetHeader("User-Agent", config.UserAgent),
-	}, nil
+	}
+	req := httpClient.client.R().SetBasicAuth(httpClient.ApiKey, httpClient.ApiSecret)
+	response, err := req.SetQueryParams(map[string]string{"encoded": "true"}).Get("auth/token")
+	if err != nil {
+		return nil, err
+	}
```
yaronya, line 43: can we extract it into a new method please? would make code more descriptive...

eladmosh: [refactoring](https://github.com/env0/terraform-provider-env0/pull/495/commits/23c4177a5c02f1bae871715ecc1a8ddf0ce24809)

client/http/client.go
```diff
@@ -29,16 +30,29 @@ type HttpClientConfig struct {
...
+
 func NewHttpClient(config HttpClientConfig) (*HttpClient, error) {
-	return &HttpClient{
+	httpClient := &HttpClient{
 		ApiKey:    config.ApiKey,
 		ApiSecret: config.ApiSecret,
 		client:    config.RestClient.SetHostURL(config.ApiEndpoint).SetHeader("User-Agent", config.UserAgent),
-	}, nil
+	}
+	response, err := getJWTToken(httpClient)
```
yaronya, line 48: Don't we need to validate there's no error here too?

client/http/client.go
```diff
@@ -29,16 +30,29 @@ type HttpClientConfig struct {
...
+
+	httpClient.jwtToken = string(response.Body())
+	return httpClient, err
```
yaronya, line 51: You can't return both an `err` and a valid response, it goes against the format...

```suggestion
	return httpClient, nil

client/http/client.go
eladmosh: [fix error return](https://github.com/env0/terraform-provider-env0/pull/495/commits/9fe7e97c5fbc0a0dc360d6df31627a72d0b9d6fd)

client/http/client.go
eladmosh: [fix error return](https://github.com/env0/terraform-provider-env0/pull/495/commits/9fe7e97c5fbc0a0dc360d6df31627a72d0b9d6fd)

Issue #496, opened 2022-09-09
Chore: upgrade libraries and fix security issues.


TomerHeber:
Upgrade the go libraries in go.mod.
This should fix reported security issues.

PR #497, opened 2022-09-09
Chore: upgrade libraries and fix security issues

TomerHeber:
### Solution

Upgrade go libraries. Part of the effort to resolve security issues.

#496


main.go
```diff
@@ -25,11 +23,7 @@ func main() {
 	opts := &plugin.ServeOpts{ProviderFunc: env0.Provider(version)}
 
 	if debugMode {
-		err := plugin.Debug(context.Background(), "registry.terraform.io/env0/env0", opts)
-		if err != nil {
-			log.Fatal(err.Error())
-		}
-		return
+		opts.Debug = true
```
razbensimon, line 26: can you explain more about this change compared to what we had before?

utils/test_reporter.go
```diff
@@ -86,3 +86,7 @@ func (r *TestReporter) Errorf(format string, args ...interface{}) {
 	r.T.Helper()
 	r.T.Errorf(format, args...)
 }
+
+func (r *TestReporter) Cleanup(cb func()) {
+	r.T.Helper()
+}
```
razbensimon, line 92: no call?

razbensimon (approved):

main.go
TomerHeber: The old code has been deprecated.
I reviewed the documentation, and this is what they suggest doing instead.

utils/test_reporter.go
razbensimon: > I'll add it.
I don't say we need it, just asking how we add a function that we don't use. what is the reason for this one

razbensimon (approved):

utils/test_reporter.go
TomerHeber: It's required because it must implement the interface. This is a new function added to the interface.

PR #501, opened 2022-09-19
Feat: allow update and create remote state backend

tomer-landesman:

we want to support changing the is_remote_backend configuration of an environment through tf provider, in order to make it easier for out clients to perform the migration from TFC to our remote backend

### Solution

added the property `is_remote_backend` to the environment struct


env0/resource_environment.go
```diff
@@ -418,7 +424,7 @@ func shouldDeploy(d *schema.ResourceData) bool {
 }
 
 func shouldUpdate(d *schema.ResourceData) bool {
-	return d.HasChanges("name", "approve_plan_automatically", "deploy_on_push", "run_plan_on_pull_requests", "auto_deploy_by_custom_glob", "auto_deploy_on_path_changes_only", "terragrunt_working_directory", "vcs_commands_alias")
```
RLRabinowitz, line 421: IIRC you can't update `isRemoteBackend` for an existing environment, no?

This seems like a bit of a unique case. We'd want to prevent terraform from allowing a user to update this field

RLRabinowitz: OK nvm, we added that to the `update` lambda, apparently

client/environment.go
```diff
@@ -94,6 +94,7 @@ type Environment struct {
 	TerragruntWorkingDirectory  string        `json:"terragruntWorkingDirectory,omitempty"`
 	VcsCommandsAlias            string        `json:"vcsCommandsAlias"`
 	BlueprintId                 string        `json:"blueprintId" tfschema:"-"`
+	IsRemoteBackend             *bool         `json:"isRemoteBackend" tfschema:"-"`
```
RLRabinowitz, line 97: `tfschema:"-"` means that we want to remove it from the schema. 
It's used for schema auto-assignment that's done in `writeResourceData`

i'm not 100% sure what adding the `tfschema:"-"` would affect, but I think that it would mean that drifts won't be detected

env0/resource_environment.go
```diff
@@ -528,6 +533,9 @@ func getCreatePayload(d *schema.ResourceData, apiClient client.ApiClientInterfac
 		return client.EnvironmentCreate{}, err
 	}
 
+	isRemoteBackend := d.Get("is_remote_backend").(bool)
+	payload.IsRemoteBackend = &isRemoteBackend
```
RLRabinowitz, line 537: `readResourceData` should do that automatically, with `tfschema`

RLRabinowitz (changes requested):

env0/resource_environment.go
```diff
@@ -217,6 +217,11 @@ func resourceEnvironment() *schema.Resource {
 					Schema: getTemplateSchema("without_template_settings.0."),
 				},
 			},
+			"is_remote_backend": {
+				Type:        schema.TypeBool,
+				Description: "should use remote backend",
+				Optional:    true,
```
TomerHeber, line 223: for type boolean it's always recommended to set a default value...

TomerHeber: you then don't need to set IsRemoteBackend as a pointer.
In addition, you won't need the "tfschema..."

RLRabinowitz (approved):
Nice. One small comment on a test

PR #502, opened 2022-09-19
Fix: Revert PR495

roni-frantchi:
### Issue
#495 introduced a fix which used an endpoint to obtain a JWT by api key to avoid 429/401.  
We've since added proper caching on api key based authorizations and will no longer need to force clients to obtain the token first.  

### Solution
Revert #495

### DIFF
--- a/client/http/client.go
+++ b/client/http/client.go
@@ -15,7 +15,6 @@ type HttpClientInterface interface {
 }
 
 type HttpClient struct {
-	jwtToken  string
 	ApiKey    string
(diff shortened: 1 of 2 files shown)


roni-frantchi:
[Revert CLI/TF Provider clients using JWT instead of using api key](https://trello.com/c/TZuWTpdw/7607-revert-cli-tf-provider-clients-using-jwt-instead-of-using-api-key)

TomerHeber (approved):
LGTM

PR #505, opened 2022-09-20
Fix: readResourceData does not work for bool or *bool fields

### LINKED ISSUE #504: `readResourceData` does not work for `bool` or `*bool` fields
The util `readResourceData` which is used to automatically read data from the schema into a struct, does not work for boolean fields in the schema

If the existing field in schema is currently `false`, then `readResourceData` would skip over that field.
For example, on `getCreatePayload`, we [explicitly set all boolean fields](https://github.com/env0/terraform-provider-env0/blob/f25bf25e924a87a3a4c10bb06dca40347847c78d/env0/resource_environment.go#L503-L507) using `d.GetOkExists`, and don't trust the result of `readResourceData` (for good reason). If we hadn't, then changing `deploy_on_push` from `false` to `true` would do nothing (we'd detect a change, and call the `update` lambda, but we'll still send `deploy_on_push`: `false`)

The reason is that `readResourceData` uses `GetOk`, which returns `ok: false` if the variable is either `0` or `false` - [code](https://github.com/env0/terraform-provider-env0/blob/f25bf25e924a87a3a4c10bb06dca40347847c78d/env0/utils.go#L75)
Ideally, we should change it to `GetOkExists`, but that might have some unforeseen side-effects

TomerHeber:
resolves #504 

### Solution

Added some special handling for "zero" values.
More details in the comments.

### DIFF
--- a/client/module.go
+++ b/client/module.go
@@ -13,8 +13,8 @@ type Module struct {
 	LogoUrl              string         `json:"logoUrl"`
 	TokenId              string         `json:"tokenId"`
 	TokenName            string         `json:"tokenName"`
-	GithubInstallationId *int           `json:"githubInstallationId"`
-	BitbucketClientKey   *string        `json:"bitbucketClientKey"`
+	GithubInstallationId *int           `json:"githubInstallationId" tfschema:",omitempty"`
+	BitbucketClientKey   *string        `json:"bitbucketClientKey" tfschema:",omitempty"`
 	IsGitlab             bool           `json:"isGitLab"`
 	SshKeys              []ModuleSshKey `json:"sshkeys"`
 	Type                 string         `json:"type"`
--- a/env0/utils.go
+++ b/env0/utils.go
@@ -50,6 +50,29 @@ func stringInSlice(str string, strs []string) bool {
 	return false
 }
 
+func reasourceDataGetValue(fieldName string, d *schema.ResourceData) interface{} {
+	dval := d.Get(fieldName)
+	if dval == nil {
+		return nil
+	}
+
+	_, okInt := dval.(int)
+	_, okBool := dval.(bool)
+	_, okString := dval.(string)
+	if okString || okBool || okInt {
+		//lint:ignore SA1019 reason: https://github.com/hashicorp/terraform-plugin-sdk/issues/817
+		if _, exists := d.GetOkExists(fieldName); !exists {
+			return nil
+		}
+	}
+
+	if sval, ok := dval.([]interface{}); ok && len(sval) == 0 {
+		return nil
+	}
+
+	return dval
+}
+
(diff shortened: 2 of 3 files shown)


env0/utils.go
```diff
@@ -50,6 +50,29 @@ func stringInSlice(str string, strs []string) bool {
 	return false
 }
 
+func reasourceDataGetValue(fieldName string, d *schema.ResourceData) interface{} {
```
TomerHeber, line 53: The way terraform sdk handles `Get` is clunky.
I'll explain in the next few lines....

env0/utils.go
```diff
@@ -50,6 +50,29 @@ func stringInSlice(str string, strs []string) bool {
...
+	dval := d.Get(fieldName)
```
TomerHeber, line 54: This will always return nil if a field does not exist in the schema. (This lines checks if the field exists in the schema).

`exists := r.Exists && !r.Computed`

env0/utils.go
```diff
@@ -50,6 +50,29 @@ func stringInSlice(str string, strs []string) bool {
...
+	if dval == nil {
+		return nil
+	}
+
+	_, okInt := dval.(int)
+	_, okBool := dval.(bool)
+	_, okString := dval.(string)
+	if okString || okBool || okInt {
+		//lint:ignore SA1019 reason: https://github.com/hashicorp/terraform-plugin-sdk/issues/817
+		if _, exists := d.GetOkExists(fieldName); !exists {
```
TomerHeber, line 64: If it's an int, boolean or string: use GetOkExists.
(GetOk would return false, also if the values are zeroes).

This will allow "zero" values of int, boolean or string to be written to the interface if they're explicitly configured.

E.g.
```
some_resource some_resource_name {
    some_value = false
}
```

env0/utils.go
```diff
@@ -50,6 +50,29 @@ func stringInSlice(str string, strs []string) bool {
...
+			return nil
+		}
+	}
+
+	if sval, ok := dval.([]interface{}); ok && len(sval) == 0 {
```
TomerHeber, line 69: If a "list" is empty, no need to write an empty list to the interface. `nil` is more correct.

client/module.go
```diff
@@ -13,8 +13,8 @@ type Module struct {
 	LogoUrl              string         `json:"logoUrl"`
 	TokenId              string         `json:"tokenId"`
 	TokenName            string         `json:"tokenName"`
-	GithubInstallationId *int           `json:"githubInstallationId"`
-	BitbucketClientKey   *string        `json:"bitbucketClientKey"`
+	GithubInstallationId *int           `json:"githubInstallationId" tfschema:",omitempty"`
```
TomerHeber, line 16: Some fixes in the tests.

RLRabinowitz (approved):

PR #506, opened 2022-09-23
Fix: auto_deploy_on_path_changes_only has a default of true, and that…

### LINKED ISSUE #503: `auto_deploy_on_path_changes_only` has a default of true, and that shouldn't always be the case
`auto_deploy_on_path_changes_only` should technically only true when either `deploy_on_push` or `run_plan_on_pull_requests` is true, and that can only be true when the template is VCS-integrated. We should rely on the backend's default, instead, for `auto_deploy_on_path_changes_only`'s value
Currently env0 does allow environments to be created with `auto_deploy_on_path_changes_only = true` even when the template is not VCS integrated (but that shouldn't happen). When an environment is updated, we **do** validate that `auto_deploy_on_path_changes_only` is not set when the template is not VCS integrated.

So, the following scenario might occur:

1. Create an environment and template via the provider. The template should not be VCS integrated (no `github_installation_id`, `gitlab_project_id`, etc.) - The environment is created with `auto_deploy_on_path_changes_only`: `true` - Which should not be the case (backend returns `auto_deploy_on_path_changes_only: false` in such a case
2. Change any information on the environment that will trigger an update, for example, change the `name`
3. Run another apply - You get an error from Backend saying that you cannot update the environment to have `auto_deploy_on_path_changes_only: true` when template is not VCS integrated (This happens because `auto_deploy_on_path_changes_only: true` was sent in the request, even when there were no changes).

So, there are two issues here:
1. `auto_deploy_on_path_changes_only`'s default should be determined only by the backend's response. If backend returned an empty `auto_deploy_on_path_changes_only`, then it should be `false`
2. When there's a change that requires a call to the `update` endpoint, `autoDeployOnPathChangesOnly` is always sent, for no reason. See [this code](https://github.com/env0/terraform-provider-env0/blob/f25bf25e924a87a3a4c10bb06dca40347847c78d/env0/resource_environment.go#L580-L581)

The first issue is the more important one. Once it is fixed the second one wouldn't matter too much

TomerHeber:
… shouldn't always be the case

fixes #503 

### Solution
Added additional validations to assertDeploymentTriggers.
Updated unit tests.

### DIFF
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -578,10 +578,6 @@ func assertDeploymentTriggers(autoDeployByCustomGlob string, continuousDeploymen
 		}
 	}
 
-	if autoDeployOnPathChangesOnly && !continuousDeployment && !pullRequestPlanDeployments {
-		return diag.Errorf("cannot set auto_deploy_on_path_changes_only when both deploy_on_push and run_plan_on_pull_requests are disabled")
-	}
-
 	return nil
 }
 
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -79,7 +79,6 @@ func resourceEnvironment() *schema.Resource {
 				Type:        schema.TypeBool,
 				Description: "redeploy only on path changes only",
 				Optional:    true,
-				Default:     true,
 			},
 			"auto_deploy_by_custom_glob": {
 				Type:        schema.TypeString,
@@ -344,6 +343,7 @@ func resourceEnvironmentCreate(ctx context.Context, d *schema.ResourceData, meta
 	}
 	d.SetId(environment.Id)
 	d.Set("deployment_id", environment.LatestDeploymentLogId)
+	d.Set("auto_deploy_on_path_changes_only", environment.AutoDeployOnPathChangesOnly)
 	setEnvironmentSchema(d, environment, environmentConfigurationVariables)
 
 	return nil
@@ -563,6 +563,11 @@ func assertDeploymentTriggers(autoDeployByCustomGlob string, continuousDeploymen
 			return diag.Errorf("cannot set auto_deploy_by_custom_glob when auto_deploy_on_path_changes_only is disabled")
 		}
 	}
+
+	if autoDeployOnPathChangesOnly && !continuousDeployment && !pullRequestPlanDeployments {
+		return diag.Errorf("cannot set auto_deploy_on_path_changes_only when both deploy_on_push and run_plan_on_pull_requests are disabled")
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


env0/resource_environment.go
```diff
@@ -79,7 +79,6 @@ func resourceEnvironment() *schema.Resource {
 				Type:        schema.TypeBool,
 				Description: "redeploy only on path changes only",
 				Optional:    true,
-				Default:     true,
```
TomerHeber, line 82: Removed the default - true.
It might cause issues to existing deployments. Users may have to set it to false explicitly.

env0/resource_environment.go
```diff
@@ -344,6 +343,7 @@ func resourceEnvironmentCreate(ctx context.Context, d *schema.ResourceData, meta
 	}
 	d.SetId(environment.Id)
 	d.Set("deployment_id", environment.LatestDeploymentLogId)
+	d.Set("auto_deploy_on_path_changes_only", environment.AutoDeployOnPathChangesOnly)
```
TomerHeber, line 346: Requested that upon creation, set the schema to the returned value.

env0/resource_environment.go
```diff
@@ -563,6 +563,11 @@ func assertDeploymentTriggers(autoDeployByCustomGlob string, continuousDeploymen
 			return diag.Errorf("cannot set auto_deploy_by_custom_glob when auto_deploy_on_path_changes_only is disabled")
 		}
 	}
+
+	if autoDeployOnPathChangesOnly && !continuousDeployment && !pullRequestPlanDeployments {
```
TomerHeber, line 567: I hope this use-case is enough. Please let me know if there are additional use-cases required.

env0/resource_environment_test.go
```diff
@@ -999,6 +991,59 @@ func TestUnitEnvironmentResource(t *testing.T) {
 			}
 			runUnitTest(t, autoDeployWithCustomGlobEnabled, func(mock *client.MockApiClientInterface) {})
 		})
+
+		t.Run("Failure in validation for auto_deploy_on_path_changes_only", func(t *testing.T) {
```
TomerHeber, line 995: Added two tests. One for creation failure and one for update failure.

env0/resource_environment.go
RLRabinowitz: > It might cause issues to existing deployments. Users may have to set it to false explicitly

Why is that? Wouldn't they just see it as a change on the first apply?

env0/resource_environment.go
RLRabinowitz: Thinking about this again, this client doesn't really have enough information to perform this validation. For example, in BE we might have a policy for an organisation that states what the default `continuousDeployment` and `pullRequestPlanDeployments` for new environments is.

I think the issue here is just that the backend validation doesn't exist on `create`, and does exist on `update`.

So we don't need this validation at all IMO.
We still need to remove the `true` default, and only send `autoDeployOnPathChangesOnly` on `update` if it has changed, as you've done in this PR

RLRabinowitz (commented):
Left a comment. After a second thought, we need that validation in BE, and not in the client, that doesn't have enough info

env0/resource_environment.go
TomerHeber: @RLRabinowitz - yes, that's all. I just wanted to mention it.

TomerHeber:
@RLRabinowitz 
Made the changes -
1a92ce8cc42b6b2977a17b4df9eb21407ee991e4

RLRabinowitz (approved):

PR #508, opened 2022-09-28
Fix: when updating without_template_settings terraform force environm…

### LINKED ISSUE #507: Second apply of environment 'without_template_settings' causes recreation
This one seems like a bug, 
- create `resource env0_environment` with `without_template_settings`
- ~~change one of `without_template_settings` attributes such as `path`~~
- run `terraform apply` once again

it forces re-creation due to `template_id` changing back to `null` - instead of updating the environment

TomerHeber:
…ent re-creation

fixes #507 

### Solution
One of the issues I observed is that the BluePrint id field of the environment is returned only during creation.
Afterward, it will be omitted.

We need to keep the template id in the schema and use it instead.

I noticed other consistency issues as well and resolved them.

Updated the acceptance test.

### DIFF
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -13,6 +13,11 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+func isTemplateless(d *schema.ResourceData) bool {
+	_, ok := d.GetOk("without_template_settings.0")
+	return ok
+}
+
 func resourceEnvironment() *schema.Resource {
 	return &schema.Resource{
 		CreateContext: resourceEnvironmentCreate,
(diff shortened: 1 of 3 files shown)


env0/resource_environment.go
```diff
@@ -13,6 +13,11 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+func isTemplateless(d *schema.ResourceData) bool {
```
TomerHeber, line 16: common code, moved to a function.

env0/resource_environment.go
```diff
@@ -410,12 +423,7 @@ func resourceEnvironmentUpdate(ctx context.Context, d *schema.ResourceData, meta
 }
 
 func shouldUpdateTemplate(d *schema.ResourceData) bool {
-	if d.Get("template_id") != "" {
-		// Using an environment with a template.
-		return false
-	}
-
-	return d.HasChange("without_template_settings.0")
+	return isTemplateless(d) && d.HasChange("without_template_settings.0")
```
TomerHeber, line 426: one line instead of five.

env0/resource_environment_test.go
```diff
@@ -1414,25 +1417,44 @@ func TestUnitEnvironmentWithoutTemplateResource(t *testing.T) {
...
-			mock.EXPECT().TemplateUpdate(environment.BlueprintId, templateUpdatePayload).Times(1).Return(updatedTemplate, nil)
-			mock.EXPECT().Template(environment.BlueprintId).Times(1).Return(updatedTemplate, nil)
-
-			// Step3
-			mock.EXPECT().Environment(environment.Id).Times(2).Return(environment, nil)
-			mock.EXPECT().ConfigurationVariablesByScope(client.ScopeEnvironment, environment.Id).Times(2).Return(client.ConfigurationChanges{}, nil)
-			mock.EXPECT().Template(environment.BlueprintId).Times(2).Return(updatedTemplate, nil)
-			mock.EXPECT().EnvironmentDestroy(environment.Id).Times(1)
+
+			gomock.InOrder(
```
TomerHeber, line 1421: inOrder.

env0/resource_environment_test.go
```diff
@@ -1414,25 +1417,44 @@ func TestUnitEnvironmentWithoutTemplateResource(t *testing.T) {
...
+				// Step1
+				// Create
+				mock.EXPECT().EnvironmentCreateWithoutTemplate(createPayload).Times(1).Return(environmentWithBluePrint, nil),
```
TomerHeber, line 1424: only this call will return an environment with a blueprint id.

avnerenv0 (approved):
@TomerHeber LGTM, but could you just add a comment in `env0/resource_environment.go` that

> the BluePrint id field of the environment is returned only during creation.
> Afterward, it will be omitted.

I know it's contained in the tests but I think it should also be put in the code, otherwise it's a little hard to understand WHY you are doing those conditionals.

PR #510, opened 2022-09-29
Fix: second apply of environment 'without_template_settings' causes r…

### LINKED ISSUE #507: Second apply of environment 'without_template_settings' causes recreation

TomerHeber:
…ecreation

Resolves #507 

### Solution

Uses CustomizeDiff to avoid forceNew in templateless mode.

### DIFF
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -10,6 +10,7 @@ import (
 	"github.com/env0/terraform-provider-env0/client"
 	"github.com/google/uuid"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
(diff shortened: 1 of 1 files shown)


env0/resource_environment.go
```diff
@@ -228,6 +228,15 @@ func resourceEnvironment() *schema.Resource {
...
 			},
 		},
+
+		CustomizeDiff: customdiff.ForceNewIf("template_id", func(ctx context.Context, d *schema.ResourceDiff, meta interface{}) bool {
+			// For templateless: any changes in template_id, no need to do anything (template id can't change).
+			// This is done due to historical bugs/issues.
+			if _, ok := d.GetOk("without_template_settings.0"); ok {
+				return false
+			}
+			return true
```
TomerHeber, line 238: If the values don't change it ignores the "true" - this is exactly what we want.

avnerenv0 (approved):
QA'd and works perfectly 🤞 🍀

PR #516, opened 2022-10-19
Fix: aws credentials validation

### LINKED ISSUE #515: AWS credentials validation
When we creating AWS credentials that include `External Id` we need to validate this property  as [AWS DOCS](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html) describe.

the regex we apply according to the DOCS above is   `/^[A-Za-z0-9+=\-.,@:/]{2,1224}$/`

So the expected behavior  - for example : 
```
resource "env0_aws_credentials" "my_role_by_arn" {
  name        = "Test Role arn ${random_string.random.result}"
  arn         = "Role ARN"
  external_id = "External id"
}
```
will fail because the white space `external_id = "External id"`.

This example will work
 ```
resource "env0_aws_credentials" "my_role_by_arn" {
  name        = "Test Role arn ${random_string.random.result}"
  arn         = "Role ARN"
  external_id = "External-id"
}
```
to summarise - make sure `external id` is correct as the requires of AWS

### PR DESCRIPTION
**Pull request by TomerHeber**
fixes #515 

### Solution

Added regex validators and add unit tests.

### CODE REVIEW

File: env0/resource_aws_credentials.go [hunk 1]
```diff
@@ -31,13 +31,14 @@ func resourceAwsCredentials() *schema.Resource {
...
-				ConflictsWith: []string{"access_key_id"},
-				RequiredWith:  []string{"arn"},
+				Type:             schema.TypeString,
+				Description:      "the aws role external id",
+				Optional:         true,
+				Sensitive:        true,
+				ForceNew:         true,
+				ConflictsWith:    []string{"access_key_id"},
+				RequiredWith:     []string{"arn"},
+				ValidateDiagFunc: NewRegexValidator(`^[A-Za-z0-9+=\-.,@:/]{2,1000}$`),
```
TomerHeber, line 41: It took a while to figure out what's wrong. I limited the regex to 1000 (more than 1000 causes a panic in Golang).

env0/resource_cost_credentials.go
```diff
@@ -24,11 +24,12 @@ func resourceCostCredentials(providerName string) *schema.Resource {
...
-			Description: "the aws role external id",
-			Sensitive:   true,
-			ForceNew:    true,
-			Required:    true,
+			Type:             schema.TypeString,
+			Description:      "the aws role external id",
+			Sensitive:        true,
+			ForceNew:         true,
+			Required:         true,
+			ValidateDiagFunc: NewRegexValidator(`^[A-Za-z0-9+=\-.,@:/]{2,1000}$`),
```
TomerHeber, line 32: It took a while to figure out what's wrong. I limited the regex to 1000 (more than 1000 causes a panic in Golang).

eladmosh:
This feature was reverted on our main repo, going to close this PR for now.

yaronya:
This needs to be done but currently blocked until we implement this on our backend

samuel-br:
@yaronya I think this PR is not relevant any more cause in the TF provider we cant have indication about SHAG, if I am right I update @TomerHeber  and remove [issue](https://github.com/env0/terraform-provider-env0/issues/515). WDYT

yaronya:
@samuel-br that is a great point actually
It means we need to alter this PR to run the validation programmatically on the plan phase. This way we can check the current organization (which is already present in cache) and see if it's self hosted or not

@TomerHeber makes sense to you?

samuel-br:
> @samuel-br that is a great point actually It means we need to alter this PR to run the validation programmatically on the plan phase. This way we can check the current organization (which is already present in cache) and see if it's self hosted or not
> 
> @TomerHeber makes sense to you?

can the TF provider can make the query as our backend about SHAG?

samuel-br:
anyway remove pending-final-review for now....

TomerHeber:
@samuel-br - could you please clarify what's needed?

samuel-br:
> @samuel-br - could you please clarify what's needed?

TL;DR - the external-id validation relevant only when user work on regular environment, if the environment is **self-hosted-agent** 

long story - when using ENV0 with self-hosted-agent we dont save the credentials on our side just a reference to the costumer external secret store as described [here](https://docs.env0.com/docs/self-hosted-kubernetes-agent#sensitive-secrets). 
As you see in the link above the "external-id" there is not a real raw external-id, what mean the validation isn't relevant there.
If we will apply the validation in TF provider, costumer uses self-hosted-agent will fail for nothing.
what is need to be done is to check if the `organization Id ` is self hosted agent and if so not validate his external-id input.
I asked @yaronya  if this ability of checking the organization-id against  backend to determent if its self-hosted-agent is optional in our TF provider, if its possible so we can validate the input but just where we can check the organization-id,
 @yaronya suggested to check it in plan phase where we have the organization-id.
Anyway  validate external-id in aws-credentials schema wont work for us.

yaronya:
@TomerHeber might be worth saving the self hosted indication on cache, the same as we do with the organization id 💡

client/api_client.go
```diff
@@ -7,8 +7,8 @@ import (
 )
 
 type ApiClient struct {
-	http                 http.HttpClientInterface
-	cachedOrganizationId string
+	http               http.HttpClientInterface
+	cachedOrganization *Organization
```
TomerHeber, line 11: Caching two fields. Therefore, I'll just cache the whole struct.

client/organization.go
```diff
@@ -41,16 +44,39 @@ func (client *ApiClient) Organization() (Organization, error) {
 	return result[0], nil
 }
 
-func (client *ApiClient) OrganizationId() (string, error) {
-	if client.cachedOrganizationId != "" {
-		return client.cachedOrganizationId, nil
+func (client *ApiClient) getCachedOrganization() (*Organization, error) {
+	cachedOrganizationLock.Lock()
```
TomerHeber, line 48: Added a mutex. I want to prevent race conditions... (shouldn't occur. But just to be on the safe side).

avnerenv0:
Sorry @yaronya can we close this please?
We're validating it in the BE, it's a little more complicated than a regex and all this logic shouldn't be in the provider. 
Having errors only in the apply stage is ok.

Can we close?

env0/resource_aws_credentials.go
```diff
@@ -2,12 +2,36 @@ package env0
...
+		return false, err
+	}
+
+	if isSelfHosted {
+		// Validating external id isn't required in self hosted organizations.
+		return true, nil
+	}
+
+	externalId, ok := d.GetOk("external_id")
+	if !ok || externalId.(string) == "" {
```
samuel-br, line 27: why is external-id empty is valid in case of the schema do contain this property?

env0/resource_aws_credentials.go
```diff
@@ -2,12 +2,36 @@ package env0
...
+var externalIdRegex *regexp.Regexp = regexp.MustCompile(`^[A-Za-z0-9+=\-.,@:/]{2,1000}$`)
+
+func IsExternalIdValid(d *schema.ResourceData, apiClient client.ApiClientInterface) (bool, error) {
+	isSelfHosted, err := apiClient.IsOrganizationSelfHostedAgent()
+
+	if err != nil {
+		return false, err
+	}
+
+	if isSelfHosted {
```
samuel-br, line 21: suggestion - to `self-hosted` we have validation to the `secret-reference ` looks like - ${ssm/gcp/azure:<secret-id>}, maybe we can add this validation too

env0/resource_aws_credentials.go
```diff
@@ -72,6 +96,14 @@ func resourceAwsCredentialsCreate(ctx context.Context, d *schema.ResourceData, m
 
 	apiClient := meta.(client.ApiClientInterface)
 
+	valid, err := IsExternalIdValid(d, apiClient)
+	if err != nil {
+		return diag.Errorf("failed to validate external_id: %v", err)
+	}
+	if !valid {
+		return diag.Errorf("external_id is invalid")
```
samuel-br, line 104: maybe add some info like - 'need to match pattern - ^[A-Za-z0-9+=\-.,@:/]{2,1000}$ '

samuel-br (changes requested):
when the external-id is for COST even if the environment is `self-hosted` the regex need to be applied, is it something we can validate?  if we cant its kind of blocking the task
To summarize:
1. self hosted
- Cost - always validate external id
- Deployment - never validate (different validation - https://docs.env0.com/docs/self-hosted-kubernetes-agent#sensitive-secrets )
2. Non-self hosted
- Cost - always
- Deployment - always

env0/resource_aws_credentials.go
TomerHeber: I'm unsure we need validation for a field we received from the env0 backend.
Could you explain a little more? (I might be missing some context).

yaronya:
> Having errors only in the apply stage is ok.

@avnerenv0 I wanted to avoid this for the user yet it might be too complicated to count as ROI positive so I'm fine with closing it as well.

TomerHeber:
on hold. There are still internal discussions about this issue.

PR #518, opened 2022-10-21
Fix: lexicographical sorting of configuration items

### LINKED ISSUE #459: lexicographical sorting of configuration items

TomerHeber:
Fixes #459 

### Solution
Maintains ordering of state as much as possible (vs the order returned by the API).
Added and updated acceptance tests accordingly.

### DIFF
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -320,7 +320,11 @@ func createVariable(configurationVariable *client.ConfigurationVariable) interfa
 }
 
 func setEnvironmentConfigurationSchema(d *schema.ResourceData, configurationVariables []client.ConfigurationVariable) {
-	ivariables := d.Get("configuration")
+	ivariables, ok := d.GetOk("configuration")
+	if !ok {
+		return
+	}
+
 	if ivariables == nil {
 		ivariables = make([]interface{}, 0)
 	}
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -178,6 +178,7 @@ func resourceEnvironment() *schema.Resource {
 							Type:        schema.TypeString,
 							Description: "the type the variable must be of",
 							Optional:    true,
+							Default:     "string",
 						},
 						"schema_enum": {
 							Type:        schema.TypeList,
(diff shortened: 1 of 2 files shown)


env0/resource_environment.go
```diff
@@ -178,6 +178,7 @@ func resourceEnvironment() *schema.Resource {
 							Type:        schema.TypeString,
 							Description: "the type the variable must be of",
 							Optional:    true,
+							Default:     "string",
```
TomerHeber, line 181: I checked... this value is always returned from the backend.

env0/resource_environment.go
```diff
@@ -277,43 +278,97 @@ func setEnvironmentSchema(d *schema.ResourceData, environment client.Environment
 	return nil
 }
 
+func createVariable(configurationVariable *client.ConfigurationVariable) interface{} {
```
TomerHeber, line 281: This is not new code. I just moved it to a function.

env0/resource_environment.go
```diff
@@ -277,43 +278,97 @@ func setEnvironmentSchema(d *schema.ResourceData, environment client.Environment
...
+		variable["schema_enum"] = configurationVariable.Schema.Enum
+		variable["schema_format"] = configurationVariable.Schema.Format
+	}
+
+	return variable
+}
+
 func setEnvironmentConfigurationSchema(d *schema.ResourceData, configurationVariables []client.ConfigurationVariable) {
-	var variables []interface{}
+	ivariables, ok := d.GetOk("configuration")
```
TomerHeber, line 323: date_environment does not have the configuration field. Without GetOk, this will crash.

env0/resource_environment.go
```diff
@@ -277,43 +278,97 @@ func setEnvironmentSchema(d *schema.ResourceData, environment client.Environment
...
-			variable["is_read_only"] = configurationVariable.IsReadOnly
+	if ivariables == nil {
+		ivariables = make([]interface{}, 0)
+	}
+
+	variables := ivariables.([]interface{})
+
+	newVariables := make([]interface{}, 0)
+
+	// The goal is to maintain existing state order as much as possible. (The backend response order may vary from state).
```
TomerHeber, line 336: comments explain it. Also you can see it in the unit test.
The order of the state and the order returned from the API may vary.

To avoid drifts we try to keep the order as described in the state.

env0/resource_environment_test.go
```diff
@@ -486,7 +486,88 @@ func TestUnitEnvironmentResource(t *testing.T) {
...
-				mock.EXPECT().ConfigurationVariablesByScope(client.ScopeEnvironment, environment.Id).Times(1).Return(client.ConfigurationChanges{}, nil) // read after create -> on update -> read after update
+				mock.EXPECT().ConfigurationVariablesByScope(client.ScopeEnvironment, environment.Id).Times(1).Return(client.ConfigurationChanges{configurationVariables}, nil) // read after create -> on update -> read after update
+				mock.EXPECT().Environment(environment.Id).Times(1).Return(environment, nil)
+
+				mock.EXPECT().EnvironmentDestroy(environment.Id).Times(1)
+			})
+		})
+
+		// Tests use-cases where the response returned by the backend varies from the order of the state.
+		t.Run("Create unordered configuration variables", func(t *testing.T) {
```
TomerHeber, line 497: The state is order is conf1 conf2.
The API response is conf2 conf1.

This test failed without my fix. After my fix this test passes.

env0/resource_environment_test.go
```diff
@@ -415,7 +415,7 @@ func TestUnitEnvironmentResource(t *testing.T) {
 				mock.EXPECT().Template(environment.LatestDeploymentLog.BlueprintId).Times(1).Return(template, nil)
 				mock.EXPECT().EnvironmentCreate(gomock.Any()).Times(1).Return(environment, nil)
 
-				mock.EXPECT().ConfigurationVariablesByScope(client.ScopeEnvironment, environment.Id).Times(1).Return(client.ConfigurationChanges{}, nil) // read after create -> on update -> read after update
+				mock.EXPECT().ConfigurationVariablesByScope(client.ScopeEnvironment, environment.Id).Times(1).Return(client.ConfigurationChanges{configurationVariables}, nil) // read after create -> on update -> read after update
```
TomerHeber, line 418: The test was incorrect. In truth, this is the real use-case (configurations are returned).

eladmosh (approved):

PR #521, opened 2022-10-27
Chore: Get teams using name filter

### LINKED ISSUE #520: Add name QS to getting teams in the team data source
When we get the `team` data source, we get ALL the teams in the org and filter it in the client.
For organizations with lots of teams, this now fails because the request is too large.

https://github.com/env0/env0/pull/9814 adds a `name` querystring param to the endpoint, so the provider should use this querying string when calling that endpoint. 

The response will be the same - but can still contain more than one team (the server checks `begins_with`) - so leave the 'client side filtering' in place.

avnerenv0:
resolves #520 

### Solution
Add `name` QS to get-teams-by-name

### DIFF
--- a/client/team.go
+++ b/client/team.go
@@ -69,28 +69,23 @@ func (client *ApiClient) TeamUpdate(id string, payload TeamUpdatePayload) (Team,
 	return result, nil
 }
 
-func (client *ApiClient) Teams() ([]Team, error) {
+func (client *ApiClient) GetTeams(params map[string]string) ([]Team, error) {
 	organizationId, err := client.OrganizationId()
 	if err != nil {
 		return nil, err
 	}
 	var result []Team
-	err = client.http.Get("/teams/organizations/"+organizationId, nil, &result)
+	err = client.http.Get("/teams/organizations/"+organizationId, params, &result)
 	if err != nil {
 		return nil, err
 	}
 	return result, err
 }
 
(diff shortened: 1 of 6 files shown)


client/team.go
```diff
@@ -81,3 +81,16 @@ func (client *ApiClient) Teams() ([]Team, error) {
 	}
 	return result, err
 }
+
+func (client *ApiClient) TeamsByName(name string) ([]Team, error) {
```
TomerHeber, line 85: To avoid code duplication, I recommend creating a helper function that receives a map.
For Teams() pass an empty map.
For TeamsByName() pass the name: name map.

TomerHeber (changes requested):
LGTM
I left a minor comment to fix.
QQ: why aren't you using pagination instead?

client/team.go
avnerenv0: done in https://github.com/env0/terraform-provider-env0/pull/521/commits/01343c92fb08adfc82d356fd38ad98eec05011cc

avnerenv0:
> why aren't you using pagination instead?

The problem is in find by name.
Pagination means we'll need to get 2800 teams  (for the current problem) to the client - only to filter for one, that seems a little unwarranted, filtering by name at the "source" sounds better to me.

BTW we will do pagination anyway, but the provider won't need to use that - the frontend will.

TomerHeber (approved):
LGTM

PR #527, opened 2022-11-03
Fix: cannot set project TTL values to infinite

### LINKED ISSUE #523: Cannot set project TTL values to infinite
when using the latest terraform provider:
+ provider [registry.terraform.io/env0/env0](http://registry.terraform.io/env0/env0) v1.2.10

I am trying to create a project and specify the TTL values.  If I set the value to days, weeks, or months, it works fine.  However, I cannot set the max value to infinite as it's not an accepted variable/string, even though it's available in the UI.

Works:

```
  project_id                    = data.env0_project.customer-test.id
  default_ttl                   = "1-M"
  max_ttl                       = "1-M"
}
```

Fails:

```
  project_id                    = data.env0_project.customer-test.id
  default_ttl                   = "1-M"
  max_ttl                       = "Infinite"
}
```

Error:

```
│ Error: 'Infinite' must be one of: 6-h, 12-h, 1-d, 3-d, 1-w, 2-w, 1-M, inherit
│ 
│   with env0_project_policy.example,
│   on projects.tf line 14, in resource "env0_project_policy" "example":
│   14:   default_ttl                   = "Infinite"
```

TomerHeber:
fixes #523 

### Solution

Added "Infinite".
Fixed some bugs.
Updated acceptance tests.
Added an integration test.

### DIFF
--- a/client/project_policy.go
+++ b/client/project_policy.go
@@ -1,7 +1,7 @@
 package client
 
 type Policy struct {
-	Id                          string  `json:"id"`
+	Id                          string  `json:"id" tfschema:"-"`
 	ProjectId                   string  `json:"projectId"`
 	NumberOfEnvironments        *int    `json:"numberOfEnvironments,omitempty"`
 	NumberOfEnvironmentsTotal   *int    `json:"numberOfEnvironmentsTotal,omitempty"`
@@ -28,8 +28,8 @@ type PolicyUpdatePayload struct {
 	SkipRedundantDeployments    bool   `json:"skipRedundantDeployments"`
 	RunPullRequestPlanDefault   bool   `json:"runPullRequestPlanDefault"`
 	ContinuousDeploymentDefault bool   `json:"continuousDeploymentDefault"`
-	MaxTtl                      string `json:"maxTtl"`
-	DefaultTtl                  string `json:"defaultTtl"`
+	MaxTtl                      string `json:"maxTtl,omitempty"`
+	DefaultTtl                  string `json:"defaultTtl,omitempty"`
 }
 
 // Policy retrieves a policy from the API
--- a/env0/resource_project_policy.go
+++ b/env0/resource_project_policy.go
@@ -12,14 +12,14 @@ import (
 )
 
 func resourcePolicy() *schema.Resource {
-	allowedProjectTtlValues := append(allowedTtlValues, "inherit")
+	allowedProjectTtlValues := append(allowedTtlValues, "Infinite", "inherit")
 	allowedProjectTtlValuesStr := fmt.Sprintf("(allowed values: %s)", strings.Join(allowedProjectTtlValues, ", "))
 
 	return &schema.Resource{
 		CreateContext: resourcePolicyCreate,
 		ReadContext:   resourcePolicyRead,
 		UpdateContext: resourcePolicyUpdate,
-		DeleteContext: resourcePolicyReset,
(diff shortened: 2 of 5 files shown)


client/project_policy.go
```diff
@@ -1,7 +1,7 @@
 package client
 
 type Policy struct {
-	Id                          string  `json:"id"`
+	Id                          string  `json:"id" tfschema:"-"`
```
TomerHeber, line 4: The "Id" isn't used for resource_project_policy - instead the project id is used.
Ignoring this field.

client/project_policy.go
```diff
@@ -28,8 +28,8 @@ type PolicyUpdatePayload struct {
 	SkipRedundantDeployments    bool   `json:"skipRedundantDeployments"`
 	RunPullRequestPlanDefault   bool   `json:"runPullRequestPlanDefault"`
 	ContinuousDeploymentDefault bool   `json:"continuousDeploymentDefault"`
-	MaxTtl                      string `json:"maxTtl"`
-	DefaultTtl                  string `json:"defaultTtl"`
+	MaxTtl                      string `json:"maxTtl,omitempty"`
```
TomerHeber, line 31: If empty string. Don't send a json with an empty string. Instead don't send the field.
This should be equivalent to "null".

env0/resource_project_policy.go
```diff
@@ -185,7 +189,9 @@ func resourcePolicyReset(ctx context.Context, d *schema.ResourceData, meta inter
 	}
 
 	d.SetId(projectId)
-	setPolicySchema(d, policy)
+	if err := writeResourceData(&policy, d); err != nil {
```
TomerHeber, line 192: bug - writeResourceData was failing and was not caught. Adding "tfschema:"-"" fixes it.

TomerHeber:
GitHub is having an issue... it's missing a commit.
Hopefully they fix it soon.

yaronya (approved):
LGTM
Just final question to get that straight - old behavior is still kept, is that correct? (i.e - when any TTL policy attribute is removed then it'd "reset" it)

chrisgnoon:
Is this resolved now?

Best Regards

*Chris Noon*
Sales Engineer, env0

***@***.***

Schedule a meeting with me
<https://t.sidekickopen86.com/s3t/c/5/f18dQhb0S7kF8bp_dgVgZkbN59hl3kW7_k2841CX6NGW36PRXq1D8nbXW2PqZkh8wpHmCf197v5Y04?te=W3R5hFj4cm2zwW3F4Fph41QWmBW1JxwY51LwWh-W1N36fv49M9lff41WtSxV3&si=8000000024089145&pi=cc7087c4-e13c-4d58-b57f-534b24210641>

TomerHeber:
> LGTM Just final question to get that straight - old behavior is still kept, is that correct? (i.e - when any TTL policy attribute is removed then it'd "reset" it)

@yaronya - I'm assuming that the "PUT" API call is not a "PATCH", and therefore should act as expected. I will test it just to be sure.

TomerHeber:
> > LGTM Just final question to get that straight - old behavior is still kept, is that correct? (i.e - when any TTL policy attribute is removed then it'd "reset" it)
> 
> @yaronya - I'm assuming that the "PUT" API call is not a "PATCH", and therefore should act as expected. I will test it just to be sure.

@yaronya - made a quick test. (It's also documented). The default value is "inherit". Therefore, if removed it will switch to inherit as expected.

PR #529, opened 2022-11-07
Fix: unable to set max environments in projects

### LINKED ISSUE #524: Unable to set max environments in projects
when using the latest terraform provider:
+ provider [registry.terraform.io/env0/env0](http://registry.terraform.io/env0/env0) v1.2.10

I am trying to create a project with the environment maximums set to null.  If I set them to a number >= 1 then the code works, but if I specify 0 or null I receive errors.

Works:

```
  project_id                    = data.env0_project.customer-test.id
  number_of_environments        = 1
  number_of_environments_total  = 1
}
```

Fails:

```
  project_id                    = data.env0_project.customer-test.id
  number_of_environments        = null
  number_of_environments_total  = null
}
```

Error:

```
│ Error: could not update policy: 400 Bad Request: /body/numberOfEnvironments must be >= 1, /body/numberOfEnvironments must be null, /body/numberOfEnvironments must match a schema in anyOf, /body/numberOfEnvironments must be null, /body/numberOfEnvironments must match a schema in anyOf, /body/numberOfEnvironmentsTotal must be >= 1, /body/numberOfEnvironmentsTotal must be null, /body/numberOfEnvironmentsTotal must match a schema in anyOf, /body/numberOfEnvironmentsTotal must be null, /body/numberOfEnvironmentsTotal must match a schema in anyOf
│ 
│   with env0_project_policy.example,
│   on projects.tf line 10, in resource "env0_project_policy" "example":
│   10: resource "env0_project_policy" "example" {
```

TomerHeber:
Fixes #524 

### Solution

Fixed a few issues (see in comments), and tested manually.

### DIFF
--- a/client/project_policy.go
+++ b/client/project_policy.go
@@ -3,8 +3,8 @@ package client
 type Policy struct {
 	Id                          string  `json:"id" tfschema:"-"`
 	ProjectId                   string  `json:"projectId"`
-	NumberOfEnvironments        *int    `json:"numberOfEnvironments,omitempty"`
-	NumberOfEnvironmentsTotal   *int    `json:"numberOfEnvironmentsTotal,omitempty"`
+	NumberOfEnvironments        *int    `json:"numberOfEnvironments,omitempty" tfschema:",omitempty"`
+	NumberOfEnvironmentsTotal   *int    `json:"numberOfEnvironmentsTotal,omitempty" tfschema:",omitempty"`
 	RequiresApprovalDefault     bool    `json:"requiresApprovalDefault"`
 	IncludeCostEstimation       bool    `json:"includeCostEstimation"`
 	SkipApplyWhenPlanIsEmpty    bool    `json:"skipApplyWhenPlanIsEmpty"`
@@ -19,8 +19,8 @@ type Policy struct {
 
 type PolicyUpdatePayload struct {
 	ProjectId                   string `json:"projectId"`
-	NumberOfEnvironments        *int   `json:"numberOfEnvironments,omitempty"`
-	NumberOfEnvironmentsTotal   *int   `json:"numberOfEnvironmentsTotal,omitempty"`
+	NumberOfEnvironments        *int   `json:"numberOfEnvironments" tfschema:",omitempty"`
+	NumberOfEnvironmentsTotal   *int   `json:"numberOfEnvironmentsTotal" tfschema:",omitempty"`
 	RequiresApprovalDefault     bool   `json:"requiresApprovalDefault"`
 	IncludeCostEstimation       bool   `json:"includeCostEstimation"`
 	SkipApplyWhenPlanIsEmpty    bool   `json:"skipApplyWhenPlanIsEmpty"`
@@ -44,6 +44,7 @@ func (client *ApiClient) Policy(projectId string) (Policy, error) {
 	if err != nil {
 		return Policy{}, err
 	}
+
 	return result, nil
 }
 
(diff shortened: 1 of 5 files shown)


client/project_policy.go
```diff
@@ -13,23 +13,23 @@ type Policy struct {
 	UpdatedBy                   string  `json:"updatedBy"`
 	RunPullRequestPlanDefault   bool    `json:"runPullRequestPlanDefault"`
 	ContinuousDeploymentDefault bool    `json:"continuousDeploymentDefault"`
-	MaxTtl                      *string `json:"maxTtl,omitempty" tfschema:",omitempty"`
-	DefaultTtl                  *string `json:"defaultTtl,omitempty" tfschema:",omitempty"`
+	MaxTtl                      *string `json:"maxTtl" tfschema:",omitempty"`
```
TomerHeber, line 16: unfortunately, the API needs to pass "null" explicitly therefore, I had to remove the omitempty.
This is incorrect behavior in the backend (Ok for PATCH requests but wrong for PUT requests).

client/project_policy.go
```diff
@@ -13,23 +13,23 @@ type Policy struct {
...
+	DefaultTtl                  *string `json:"defaultTtl" tfschema:",omitempty"`
 }
 
 type PolicyUpdatePayload struct {
 	ProjectId                   string `json:"projectId"`
-	NumberOfEnvironments        *int   `json:"numberOfEnvironments,omitempty"`
-	NumberOfEnvironmentsTotal   *int   `json:"numberOfEnvironmentsTotal,omitempty"`
+	NumberOfEnvironments        *int   `json:"numberOfEnvironments"`
```
TomerHeber, line 22: unfortunately, the API needs to pass "null" explicitly therefore, I had to remove the omitempty.
This is incorrect behavior in the backend (Ok for PATCH requests but wrong for PUT requests).

client/project_policy.go
```diff
@@ -19,8 +19,8 @@ type Policy struct {
 
 type PolicyUpdatePayload struct {
 	ProjectId                   string `json:"projectId"`
-	NumberOfEnvironments        *int   `json:"numberOfEnvironments,omitempty"`
-	NumberOfEnvironmentsTotal   *int   `json:"numberOfEnvironmentsTotal,omitempty"`
+	NumberOfEnvironments        *int   `json:"numberOfEnvironments" tfschema:",omitempty"`
```
TomerHeber, line 22: Removed the omitempty for the json.
In general it shouldn't be required, but it seems to be an issue with how the backend works (expects "null").

env0/utils.go
```diff
@@ -50,15 +50,25 @@ func stringInSlice(str string, strs []string) bool {
...
+func reasourceDataGetValue(fieldName string, omitEmpty bool, d *schema.ResourceData) interface{} {
 	dval := d.Get(fieldName)
 	if dval == nil {
 		return nil
 	}
 
-	_, okInt := dval.(int)
+	ival, okInt := dval.(int)
+
+	if omitEmpty && okInt && ival == 0 {
```
TomerHeber, line 61: handles omit empty for empty string and or zero int.
Up until now, it wasn't required. The first use-case it's needed, so added it.

yaronya (commented):
Please also add a sentence to the docs, describing which TF value is required to reset the number of total environments. We can also specify that removing the value completely does the same (if it's not documented till now...)

TomerHeber:
> Please also add a sentence to the docs, describing which TF value is required to reset the number of total environments. We can also specify that removing the value completely does the same (if it's not documented till now...)



> Please also add a sentence to the docs, describing which TF value is required to reset the number of total environments. We can also specify that removing the value completely does the same (if it's not documented till now...)

@yaronya - updated the text. Hopefully, now it's clearer.

yaronya (approved):

PR #540, opened 2022-11-17
Feat: add data source for getting all custom roles

### LINKED ISSUE #535: Add data source for getting all custom roles
Add a new data source to list all custom roles in the org.
This should provide a string list of all the custom roles in a given Org.

This list should be able to be used to create a custom role data source based on the names returned.

Usage example:
```
data "env0_custom_roles" "roles" {}

data "env0_custom_role" "my_roles" {
for_each = toset(data.env0_custom_roles.roles.names)
name = each.value
}
```
Relevant API:
https://developer.env0.com/docs/api/ffe8f19130338-get-all-roles

TomerHeber:
resolves #535

### Solution
1. Added a new data source.
2. Added acceptance tests.
3. Added integration tests.
4. Added examples.


env0/data_custom_roles.go
```diff
@@ -0,0 +1,51 @@
...
+	apiClient := meta.(client.ApiClientInterface)
+	roles, err := apiClient.Roles()
+	if err != nil {
+		return diag.Errorf("Failed to get custom roles: %v", err)
+	}
+
+	data := []string{}
+
+	for _, role := range roles {
+		// Built-in roles do not have an identifier.
```
TomerHeber, line 39: If this ever changes, it will break.
There is no other way to differentiate between custom and pre-built.
At the moment both are returned via the API.

ItamarMalka: actually we have a isDefaultRole on the response for that api which we omitted from the api, the omission might be a mistake on our end, I asked in our rbac channel and will update here 👍

ItamarMalka: so there is actually no reason for the `isDefaultRole` omission on the api so ill just open a pr to add it to it, but I am pretty sure can use it from the response already 🙂

TomerHeber: @ItamarMalka - thanks for the feedback.
I have fixed it.

PR #544, opened 2022-11-29
Fix: switching a VCS provider of a VCS Environment fails with 400

### LINKED ISSUE #543: Switching a VCS provider of a VCS Environment fails with 400
### Steps to reproduce
1. Create a [VCS Environment](https://docs.env0.com/docs/setting-up-a-new-environment#based-on-a-direct-vcs-integration)
```terraform
  without_template_settings {
    repository           = "https://bitbucket.org/foo/bar"
    path                 = "boo"
    type                 = "terraform"
    terraform_version    = "1.2.9"
    bitbucket_client_key = "xxxx"
  }
```
3. Deploy the Environment
4. Change the it so the provider would be GitHub
```terraform
  without_template_settings {
    repository           = "https://github.com/foo/bar"
    path                 = "boo"
    type                 = "terraform"
    terraform_version    = "1.2.9"
    github_installation_id = 30413641
  }
```
6. Deploy fails with 400:
```
"You may only specify at most one of the parameters {\"githubInstallationId\":xxxx,\"bitbucketClientKey\":\"ari:cloud:bitbucket::app/{xxxxx}/env0\"}"
```

Our API expects `bitbucket_client_key` to be nulled out.  
Trying to explicitly do so, fails with the same error

TomerHeber:
fixes #543 

### Solution

explicitly passing null.

Note:
It may have been better to fix it in the backend.
E.g. backend should treat variables as "null" if they're undefined.


client/template.go
```diff
@@ -76,15 +76,15 @@ type TemplateCreatePayload struct {
 	Path                 string           `json:"path"`
 	IsGitLab             bool             `json:"isGitLab"`
 	TokenName            string           `json:"tokenName"`
-	TokenId              string           `json:"tokenId,omitempty"`
-	GithubInstallationId int              `json:"githubInstallationId,omitempty"`
+	TokenId              NullString       `json:"tokenId"`
```
TomerHeber, line 79: NullString and NullInt pass null explicitly when marshaled.

roni-frantchi:
[Unable to switch VCS Provider using TF Provider](https://trello.com/c/nLoRFSRJ/8247-unable-to-switch-vcs-provider-using-tf-provider)

env0/resource_environment_test.go
```diff
@@ -1329,16 +1329,16 @@ func TestUnitEnvironmentWithoutTemplateResource(t *testing.T) {
 	templateCreatePayload := client.TemplateCreatePayload{
 		Repository:           template.Repository,
 		Description:          template.Description,
-		GithubInstallationId: template.GithubInstallationId,
+		GithubInstallationId: client.NullInt(template.GithubInstallationId),
```
sagilaufer1992, line 1332: can the types be imported once? IMO it's less elegant to use `client.TypeX` for casting every time

env0/resource_template_test.go
```diff
@@ -469,17 +469,17 @@ func TestUnitTemplateResource(t *testing.T) {
 				Name:                 templateUseCase.template.Name,
 				Repository:           templateUseCase.template.Repository,
 				Description:          templateUseCase.template.Description,
-				GithubInstallationId: templateUseCase.template.GithubInstallationId,
+				GithubInstallationId: client.NullInt(templateUseCase.template.GithubInstallationId),
```
sagilaufer1992, line 472: same here, import once, use just the type for casting.

sagilaufer1992 (approved):
looks great!
minor comments regarding style, non blocking.
I agree regarding the fact that our BE should solve this problem beforehand, but double verification probably won't hurt

roni-frantchi:
We're a stab at fixing it on the backend side - let's wait on this one

roni-frantchi:
Fixed our backend to cope with these fields being omitted - so https://github.com/env0/terraform-provider-env0/issues/543 should now be resolved.  
Closing this PR and the issue for now.

PR #546, opened 2022-11-29
Feat: new custom role resource

### LINKED ISSUE #498: New Custom role resource
New resource
Accepts name and list of permissions
We can validate that list in the backend (no need to have all possible strings in the provider code)
Returns the id of the new role

Example:

```
resource “env0_custom_role” “my_custom_role” {
  name = “my custom role 1”
  permissions = [
    “view_project”,
    “edit_project_settings”
  ]
}
```
Role assignment resources - project level
Should be implemented in the ​​env0_team_project_assignment and ​​env0_user_project_assignment resources
Accept “custom_role_id” 
either that or role are mandatory - but only one can be given.

Example:

```
resource "env0_user_project_assignment" "project_assignment_example" { 
  user_id = data.env0_user.user_example.id 
  project_id = env0_project.project_example.id 
  custom_role_id = env0_custom_role.my_custom_role.id
}
```
Role assignment resources - org level
Should be implemented in the ​​env0_team_organization_assignment and ​​env0_user_organization_assignment resources (new resources, do not currently exist)
Accept “custom_role_id” 
either that or role is mandatory - but only one can be given.

may also update the team created to have a new default role at the org level

Example:

```
resource "env0_user_organization_assignment" "org_assignment_example" { 
  user_id = data.env0_user.user_example.id 
  custom_role_id = env0_custom_role.my_custom_role.id
}
```

TomerHeber:
fixes #498 

### Solution

Added integration tests and examples.

### DIFF
--- a/client/http/client.go
+++ b/client/http/client.go
@@ -62,10 +62,13 @@ func (client *HttpClient) Post(path string, request interface{}, response interf
 }
 
 func (client *HttpClient) Put(path string, request interface{}, response interface{}) error {
-	result, err := client.request().
-		SetBody(request).
-		SetResult(response).
-		Put(path)
+	req := client.request().SetBody(request)
+	if response != nil {
+		req = req.SetResult(response)
+	}
+
+	result, err := req.Put(path)
+
 	return client.httpResult(result, err)
 }
 
--- a/examples/resources/env0_user_organization_assignment/resource.tf
+++ b/examples/resources/env0_user_organization_assignment/resource.tf
@@ -1,0 +1,15 @@
+data "env0_user" "user_example" {
+  email = "example@email.com"
+}
+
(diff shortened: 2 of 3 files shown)


client/http/client.go
```diff
@@ -62,10 +62,13 @@ func (client *HttpClient) Post(path string, request interface{}, response interf
 }
 
 func (client *HttpClient) Put(path string, request interface{}, response interface{}) error {
-	result, err := client.request().
-		SetBody(request).
-		SetResult(response).
-		Put(path)
+	req := client.request().SetBody(request)
```
TomerHeber, line 65: While running the integration test the code crashed.
SetResult with nil is invalid. Added an "if" to resolve it.

yaronya (approved):

PR #548, opened 2022-12-01
Feat: add support for Azure DevOps templates

### LINKED ISSUE #547: Add support for Azure DevOps templates
Right now we don't have a way to add an AzureDevOps template.
We should add an is_azure_devops to match the same field on our [api](https://developer.env0.com/docs/api/b3A6MTg2NzAxMTQ-create-template)
note that a Azure DevOps template needs a tokenId like Gitlab templates

TomerHeber:
closes #547 

### Solution

1. Added the new field.
2. Updated acceptance tests.

Note: integration test not added. I could add the integration test if there is an existing template in the test account.

### DIFF
--- a/client/template.go
+++ b/client/template.go
@@ -64,7 +64,7 @@ type Template struct {
 	IsBitbucketServer    bool             `json:"isBitbucketServer"`
 	FileName             string           `json:"fileName,omitempty" tfschema:",omitempty"`
 	IsTerragruntRunAll   bool             `json:"isTerragruntRunAll"`
-	IsAzureDevOps        bool             `json:"isAzureDevOps"`
+	IsAzureDevOps        bool             `json:"isAzureDevOps" tfschema:"is_azure_devops"`
 }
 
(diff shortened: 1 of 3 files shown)


ItamarMalka (changes requested):
@TomerHeber Can you change those `azure_dev_ops` into `azure_devops`?
as it fits our naming scheme better

TomerHeber:
> @TomerHeber Can you change those `azure_dev_ops` into `azure_devops`? as it fits our naming scheme better Other then that the PR looks great 👍
> 
> > integration test not added. I could add the integration test if there is an existing template in the test account
> 
> We are doing these atm it should be there in the next couple of days

@ItamarMalka sure!
The confusion was because in the API it's called "azureDevOps" (and not "azureDevops").

TomerHeber:
@ItamarMalka c8ce38b8e6c022628115337d005d470cb50d5ca5

ItamarMalka (approved):
LGTM! 👍🏻

PR #551, opened 2022-12-09
Fix: "Infinite" isn't a valid value for the max/default TTL Policies API

### LINKED ISSUE #549: "Infinite" isn't a valid value for the max/default TTL Policies API
**_Unfortunately_**, `"Infinite"` is **not** a supported value for max/default TTL on Policies API.  
An explicit `null` is used to denote there is **no** TTL restriction should be applied.  

To reproduce:  

```terraform
terraform {
  required_providers {
    env0 = {
      source = "env0/env0"
      version = "1.2.11"
    }
  }
}

provider "env0" {
  api_key      = "xxxx"
  api_secret   = "yyy"
}

  project_id                    = data.env0_project.default_project.id
  number_of_environments        = null
  number_of_environments_total  = null
  max_ttl                       = "Infinite"  # <--- this will soon be blocked by our API as it is not a supported value, explicit `null` is
  default_ttl                   = "1-w"
}

data "env0_project" "default_project" {
  name = "My First Project"
}

```  

### DoD  
Either the provider map `"Infinite"` to `null` (and vice-versa on `data` and verification) on both this resource and `organization_policy`, or drop `"Infinite"` as valid value altogether as it doesn't work.  
Note there's an integration test to be changed as well

TomerHeber:
fixes #549 

### Solution
The acceptance test was crashing, and therefore it never ran successfully (and was hiding the bug).
Once I fixed the test I saw a couple of problems. I fixed them.
This should now work as expected.

### DIFF
--- a/client/project_policy.go
+++ b/client/project_policy.go
@@ -19,8 +19,8 @@ type Policy struct {
 
 type PolicyUpdatePayload struct {
 	ProjectId                   string `json:"projectId"`
-	NumberOfEnvironments        *int   `json:"numberOfEnvironments,omitempty" tfschema:",omitempty"`
-	NumberOfEnvironmentsTotal   *int   `json:"numberOfEnvironmentsTotal,omitempty" tfschema:",omitempty"`
+	NumberOfEnvironments        int    `json:"numberOfEnvironments,omitempty"`
+	NumberOfEnvironmentsTotal   int    `json:"numberOfEnvironmentsTotal,omitempty"`
 	RequiresApprovalDefault     bool   `json:"requiresApprovalDefault"`
 	IncludeCostEstimation       bool   `json:"includeCostEstimation"`
 	SkipApplyWhenPlanIsEmpty    bool   `json:"skipApplyWhenPlanIsEmpty"`
--- a/client/project_policy.go
+++ b/client/project_policy.go
@@ -19,8 +19,8 @@ type Policy struct {
 
 type PolicyUpdatePayload struct {
 	ProjectId                   string `json:"projectId"`
-	NumberOfEnvironments        *int   `json:"numberOfEnvironments" tfschema:",omitempty"`
-	NumberOfEnvironmentsTotal   *int   `json:"numberOfEnvironmentsTotal" tfschema:",omitempty"`
+	NumberOfEnvironments        *int   `json:"numberOfEnvironments,omitempty" tfschema:",omitempty"`
+	NumberOfEnvironmentsTotal   *int   `json:"numberOfEnvironmentsTotal,omitempty" tfschema:",omitempty"`
 	RequiresApprovalDefault     bool   `json:"requiresApprovalDefault"`
 	IncludeCostEstimation       bool   `json:"includeCostEstimation"`
 	SkipApplyWhenPlanIsEmpty    bool   `json:"skipApplyWhenPlanIsEmpty"`
--- a/env0/resource_project_policy.go
+++ b/env0/resource_project_policy.go
@@ -38,13 +38,13 @@ func resourcePolicy() *schema.Resource {
 			},
 			"number_of_environments": {
 				Type:             schema.TypeInt,
-				Description:      "Max number of environments a single user can have in this project.\nSetting to `null` or omitting, removes the restriction.",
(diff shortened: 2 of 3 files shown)


env0/resource_project_policy.go
```diff
@@ -162,7 +162,7 @@ func resourcePolicyUpdate(ctx context.Context, d *schema.ResourceData, meta inte
 		payload.DefaultTtl = ""
 	}
 
-	if payload.MaxTtl == "Infinine" {
+	if payload.MaxTtl == "Infinite" {
```
TomerHeber, line 165: This is what probably caused the issue. Because the acceptance was crashing it didn't catch the issue in the test.

env0/resource_project_policy_test.go
```diff
@@ -48,110 +49,107 @@ func TestUnitPolicyResource(t *testing.T) {
...
-					resource.TestCheckNoResourceAttr(accessor, "number_of_environments_total"),
-					resource.TestCheckResourceAttr(accessor, "requires_approval_default", strconv.FormatBool(updatedPolicy.RequiresApprovalDefault)),
-					resource.TestCheckResourceAttr(accessor, "include_cost_estimation", strconv.FormatBool(updatedPolicy.IncludeCostEstimation)),
-					resource.TestCheckResourceAttr(accessor, "skip_apply_when_plan_is_empty", strconv.FormatBool(updatedPolicy.SkipApplyWhenPlanIsEmpty)),
-					resource.TestCheckResourceAttr(accessor, "disable_destroy_environments", strconv.FormatBool(updatedPolicy.DisableDestroyEnvironments)),
-					resource.TestCheckResourceAttr(accessor, "skip_redundant_deployments", strconv.FormatBool(updatedPolicy.SkipRedundantDeployments)),
-					resource.TestCheckResourceAttr(accessor, "max_ttl", "Infinite"),
-					resource.TestCheckResourceAttr(accessor, "default_ttl", *updatedPolicy.DefaultTtl),
-				),
+	t.Run("create and update", func(t *testing.T) {
```
TomerHeber, line 52: Not too many changes. 
1. It's now under a subtest.
2. Fixed the place where is caused a crash (get a value for nil pointer --- ooops).

env0/resource_project_policy_test.go
```diff
@@ -48,110 +49,107 @@ func TestUnitPolicyResource(t *testing.T) {
...
-			NumberOfEnvironments:       resetPolicy.NumberOfEnvironments,
-			NumberOfEnvironmentsTotal:  resetPolicy.NumberOfEnvironmentsTotal,
-			RequiresApprovalDefault:    resetPolicy.RequiresApprovalDefault,
-			IncludeCostEstimation:      resetPolicy.IncludeCostEstimation,
-			SkipApplyWhenPlanIsEmpty:   resetPolicy.SkipApplyWhenPlanIsEmpty,
-			DisableDestroyEnvironments: resetPolicy.DisableDestroyEnvironments,
-			SkipRedundantDeployments:   resetPolicy.SkipRedundantDeployments,
-		}).Times(1).Return(resetPolicy, nil)
+		runUnitTest(t, testCase, func(mock *client.MockApiClientInterface) {
+			gomock.InOrder(
```
TomerHeber, line 111: switched to inOrder.

client/project_policy.go
```diff
@@ -19,8 +19,8 @@ type Policy struct {
 
 type PolicyUpdatePayload struct {
 	ProjectId                   string `json:"projectId"`
-	NumberOfEnvironments        *int   `json:"numberOfEnvironments" tfschema:",omitempty"`
-	NumberOfEnvironmentsTotal   *int   `json:"numberOfEnvironmentsTotal" tfschema:",omitempty"`
+	NumberOfEnvironments        int    `json:"numberOfEnvironments,omitempty"`
```
TomerHeber, line 22: not necessary, but I like it better without the pointer version.

env0/resource_project_policy.go
roni-frantchi: > Because the acceptance was crashing it didn't catch the issue in the test

@TomerHeber @yaronya Can we open an issue and follow up on fixing _that_, please?

roni-frantchi (approved):

PR #557, opened 2022-12-23
Feat: provider should retry API requests

### LINKED ISSUE #556: Provider should retry API requests
When we get a retry-able request - a >500 or a 'dns lookup' error - we should retry the request.
This helps a lot since the provider is used a lot in CI environments, and it's a shame that a small glitch in the flow will kill the entire run.

reference - https://github.com/go-resty/resty/blob/master/retry_test.go

TomerHeber:
resolves #556 

### Solution
Enabled the retry mechanism and added conditions for retrying: network error and 5xx.

Note: Integration tests will be used to validate this change. The benefit of adding unit/acceptance tests for this is low (no point in testing "rest")

### DIFF
--- a/env0/provider.go
+++ b/env0/provider.go
@@ -2,6 +2,7 @@ package env0
 
 import (
 	"context"
+	"time"
 
 	"github.com/env0/terraform-provider-env0/client"
 	"github.com/env0/terraform-provider-env0/client/http"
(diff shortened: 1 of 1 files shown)


env0/provider.go
```diff
@@ -124,16 +125,33 @@ func configureProvider(version string, p *schema.Provider) schema.ConfigureConte
 	userAgent := p.UserAgent("terraform-provider-env0", version)
 
 	return func(ctx context.Context, d *schema.ResourceData) (interface{}, diag.Diagnostics) {
+		restyClient := resty.New()
+
+		restyClient.
+			SetRetryCount(5).
+			SetRetryWaitTime(time.Second).
```
TomerHeber, line 132: Note: exponential backoff is used.

yaronya (commented):
Can we maybe add an opt-out option per API call? In case retrying would do more harm than good

TomerHeber:
> Can we maybe add an opt-out option per API call? In case retrying would do more harm than good

@yaronya responded here: https://env0.slack.com/archives/C01QBBEGBEH/p1672006658414369?thread_ts=1671526429.220979&cid=C01QBBEGBEH

I was thinking of an environment variable. Let me know your thoughts.

yaronya (approved):
In a second thought, what I asked you to add is a little premature at this point - so let's wait for that rainy day 😊

Issue #550, opened 2022-12-05
Custom Flow Support


yarivg:
add support for custom flow configuration

```
resource "env0_custom_flow" "example" {
  name                   = "example"
  repository             = "example"
  path                   = "example"
  revision               = "example"
  token_id               = 0
  ssh_keys               = []
  gitlab_project_id      = 0
  github_installation_id = 0
  bitbucket_client_key   = 0
  is_bitbucket_server    = "true/false"
  is_gitlab_enterprise   = "true/false"
  is_github_enterprise   = "true/false"
  is_git_lab             = "true/false"
  is_azure_devOps        = "true/false"
}

resource "env0_custom_flow_assignment" "assignment_example" {
  depends_on = [env0_custom_flow.example]
  scope      = "PROJECT"
  scopeId    = "example"
  templateId = env0_custom_flow.example.id
}
```

**Custom Flow API**
- [Create](https://developer.env0.com/docs/api/b0edc780360d8-create-a-custom-flow)
- [Update](https://developer.env0.com/docs/api/5c9f86bda309b-update-a-custom-flow)
- [Delete](https://developer.env0.com/docs/api/9bf511118e20d-delete-a-custom-flow)
- [Assign](https://developer.env0.com/docs/api/8314044c581b9-assign-custom-flow)
- [Unassign](https://developer.env0.com/docs/api/7e931f6b0acdf-unassign-custom-flow)

yaronya:
@yarivg is is something you'll be implementing or should @TomerHeber address it?

yarivg:
Tomer

yarivg:
> @yarivg is it something you'll be implementing or should @TomerHeber address it?

Tomer

TomerHeber:
@yaronya  @yarivg I will need more information before I proceed.

The API for creation is as follows:
https://developer.env0.com/docs/api/b0edc780360d8-create-a-custom-flow

I need a list of all the constraints and invalid/valid use-case, so I can "block"/"allow" them in the provider.

Thanks!

yarivg:
> @yaronya @yarivg I will need more information before I proceed.
> 
> The API for creation is as follows: https://developer.env0.com/docs/api/b0edc780360d8-create-a-custom-flow
> 
> I need a list of all the constraints and invalid/valid use-case, so I can "block"/"allow" them in the provider.
> 
> Thanks!

Answered in slack.

PR #558, opened 2022-12-23
Feat: custom flow support

TomerHeber:
### Solution

Part of the effort is to add custom-flow support #550.

In this PR I added API client support for CRUD operations on custom flows.
Unit tests were added as well.

### DIFF
--- a/client/custom_flow.go
+++ b/client/custom_flow.go
@@ -1,0 +1,103 @@
+package client
+
+type CustomFlowCreatePayload struct {
+	Name                 string           `json:"name"`
+	Repository           string           `json:"repository"`
+	Path                 string           `json:"path,omitempty"`
+	Revision             string           `json:"revision,omitempty"`
+	TokenId              string           `json:"tokenId,omitempty"`
+	SshKeys              []TemplateSshKey `json:"sshKeys,omitempty"`
+	GithubInstallationId int              `json:"githubInstallationId,omitempty"`
+	BitbucketClientKey   string           `json:"bitbucketClientKey,omitempty"`
+	IsBitbucketServer    bool             `json:"isBitbucketServer"`
+	IsGitlabEnterprise   bool             `json:"isGitLabEnterprise"`
+	IsGithubEnterprise   bool             `json:"isGitHubEnterprise"`
+	IsGitLab             bool             `json:"isGitLab"`
+	IsAzureDevOps        bool             `json:"isAzureDevOps" tfschema:"is_azure_devops"`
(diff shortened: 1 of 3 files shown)


client/custom_flow.go
```diff
@@ -0,0 +1,103 @@
...
+	SshKeys              []TemplateSshKey `json:"sshKeys,omitempty"`
+	GithubInstallationId int              `json:"githubInstallationId,omitempty"`
+	BitbucketClientKey   string           `json:"bitbucketClientKey,omitempty"`
+	IsBitbucketServer    bool             `json:"isBitbucketServer"`
+	IsGitlabEnterprise   bool             `json:"isGitLabEnterprise"`
+	IsGithubEnterprise   bool             `json:"isGitHubEnterprise"`
+	IsGitLab             bool             `json:"isGitLab"`
+	IsAzureDevOps        bool             `json:"isAzureDevOps" tfschema:"is_azure_devops"`
+	IsTerragruntRunAll   bool             `json:"isTerragruntRunAll"`
+}
```
yaronya, line 18: Is it possible to add some composition with the already existing attribute of the template resource?

client/custom_flow.go
```diff
@@ -0,0 +1,103 @@
...
+
+type CustomFlow struct {
```
yaronya, line 20: Would be great to DRY this a bit and add a shared `struct` between this one and `CustomFlowCreatePayload`

client/custom_flow.go
TomerHeber: Yes! It's possible. It makes more sense to me.

Embedded composition in golang is explicit. See this for an explanation:
https://blog.dlow.me/programming/golang/struct-embedding/#golang-struct-embedding

This means that I will have to make changes in all the code locations where Template is initialized
E.g.:
```
	template := client.Template{
		Id:               "id0",
		Name:             "template",
...

```

It's not an issue - just additional (many) cosmetic changes in the PR.
Let me know if I should proceed. Another option is to apply it as a follow-up PR.

client/custom_flow.go
TomerHeber: They vary from a "tagging" perspective. (json and tfschema).
Trying to merge may become cumbersome, and the benefit is small. It is currently consistent with the rest of the code.

client/custom_flow.go
yaronya: Yeah it's probably not worth it...

yaronya (approved):

PR #559, opened 2022-12-26
Fix: using git token_id on Templates always assumes GitLab

### LINKED ISSUE #545: Using git `token_id` on Templates always assumes GitLab
Git tokens in env0 can be used to integrate with "Other VCS", GitLab, and soon Azure DevOps.  
Setting `token_id` on this provider [always assumes](https://github.com/env0/terraform-provider-env0/blob/75af2bebbc1b57400acfa1d6d68e54c577b7fd86/env0/resource_template.go#L342) we're using Gitlab, causing the request to fail.  

# Steps to reproduce:
Apply:  
```terraform
resource "env0_environment" "foobar" { #could also be just a template_resource
  name        = "foobar"
  project_id  = data.env0_project.default_project.id
  
  without_template_settings {
    repository = "https://github.com/env0/templates"
    token_id = "68ba48c9-ec6e-4177-a130-e6f3ffb02f9b"
  }
}
```  

This will fail with:  
```
│ Error: could not update template: 400 Bad Request: You have no access to that repository
```  

Examining the request made by the client shows the template is marked as **gitlab** despite not setting `isGitlab: true`

TomerHeber:
fixes #545 

### Solution

For templates implicilty sets IsGitlab if tokenId is set and azure_devops isn't configured (backward compatible).
For modules no longer sets IsGitlab implicitly.


env0/resource_module.go
```diff
@@ -99,10 +99,6 @@ func resourceModuleCreate(ctx context.Context, d *schema.ResourceData, meta inte
 		return diag.Errorf("schema resource data deserialization failed: %v", err)
 	}
 
-	if len(payload.TokenId) > 0 {
```
TomerHeber, line 102: for modules I decided to remove the implicit IsGitlab (This is less backward compatible).

env0/resource_template.go
```diff
@@ -344,11 +344,17 @@ func templateCreatePayloadFromParameters(prefix string, d *schema.ResourceData)
...
+	isAzureDevOpsKey := "is_azure_devops"
+
 	if prefix != "" {
 		tokenIdKey = prefix + "." + tokenIdKey
+		isAzureDevOpsKey = prefix + "." + isAzureDevOpsKey
 	}
+
 	if tokenId, ok := d.GetOk(tokenIdKey); ok {
-		payload.IsGitLab = tokenId != ""
+		if isAzureDevOps := d.Get(isAzureDevOpsKey); !isAzureDevOps.(bool) {
```
TomerHeber, line 355: only set implicit isGitlab if is_azure_devops is false.
This is more backward compatible.

env0/resource_module.go
```diff
@@ -51,7 +51,7 @@ func resourceModule() *schema.Resource {
 			},
 			"token_id": {
 				Type:         schema.TypeString,
-				Description:  "the token id used for integration with GitLab",
+				Description:  "the token id used for integration with GitLab, Azure Devops, and more...",
```
yaronya, line 54: This isn't 100% correct because it's also used as a simple git token. I would simple say:
```suggestion
				Description:  "the git token id to be used",

File: env0/resource_template.go [hunk 4]
```diff
@@ -138,9 +138,9 @@ func getTemplateSchema(prefix string) map[string]*schema.Schema {
 		},
 		"token_id": {
 			Type:          schema.TypeString,
-			Description:   "the token id used for private git repos or for integration with GitLab, you can get this value by using a data resource of an existing Gitlab template or contact our support team",
+			Description:   "the token id is used for integration with repositories such as: GitLab, Azure DevOps, and more...",
```
yaronya, line 141: Same as previous comment

env0/resource_template.go
yaronya: Not sure I understand the logic here.

What's the prefix is used for? how come we assert it from the existing schema although it's not static? I'm really not in the details so an explanation would be superb please 🙏

TomerHeber: Ahh no problems!
A standalone template and an environment with a template inside use the same code and schema.
(It took a bit of time to get it done correctly, but now any changes in the template apply to both use cases).
In some cases, an "if" is required.

yaronya: Would appreciate adding some comment on that as code is (rightfully) quite generic on this case

TomerHeber: Hopefully, I understood what you meant. I added a comment.

TomerHeber: 0cdf4d42c5c171549e579fad7f7c3f690736faeb

yaronya (approved):

PR #561, opened 2022-12-28
Feat: ignore archived environments when searching by name

### LINKED ISSUE #560: Ignore archive environments when searching by name
The user uses the `env0_environemt `data resource and searches the environment by name and the API also returns an archived environment.
We specifically say the environment’s name should be unique, but the user could not find the old archived environment since it was archived more than 6 months ago and can’t be seen in the FE.
And we can't ignore all archive environments since users will lose access to this environment.

Suggestion:
Add option to ignore archive environments with default false to preserve current behavior 

```
data "env0_environment" "same_name_environment" {
  name = "already used name"
  exclude_archived = true
}
```

TomerHeber:
resolves #560 

### Solution

1. added a new optional schema field called: `exclude_archived`.
2. added acceptance tests to cover the new use-case.


TomerHeber:
Updated `client/api_client_mock.go`. Missed it in one of the previous PRs.
(Unrelated to this PR).

yaronya (approved):

PR #563, opened 2022-12-31
Feat: custom flow support

TomerHeber:
### Solution

Part of the effort to add custom flow support #550 
In this PR I implemented the assignment API calls and added unit tests.

### DIFF
--- a/client/custom_flow.go
+++ b/client/custom_flow.go
@@ -38,8 +38,8 @@ type CustomFlow struct {
 type CustomFlowAssignmentScope string
 
 const (
-	CustomFlowOrganizationScope CustomFlowAssignmentScope = "ORGANIZATION"
-	CustomFlowProjectScope      CustomFlowAssignmentScope = "PROJECT"
+	// CustomFlowOrganizationScope CustomFlowAssignmentScope = "ORGANIZATION" - to be added if required in the future.
+	CustomFlowProjectScope CustomFlowAssignmentScope = "PROJECT"
 )
 
 type CustomFlowAssignment struct {
--- a/client/custom_flow.go
+++ b/client/custom_flow.go
@@ -35,6 +35,19 @@ type CustomFlow struct {
 	IsTerragruntRunAll   bool             `json:"isTerragruntRunAll"`
 }
 
+type CustomFlowAssignmentScope string
+
+const (
+	CustomFlowOrganizationScope CustomFlowAssignmentScope = "ORGANIZATION"
+	CustomFlowProjectScope      CustomFlowAssignmentScope = "PROJECT"
+)
+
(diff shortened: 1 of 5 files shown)


client/custom_flow.go
```diff
@@ -101,3 +114,20 @@ func (client *ApiClient) CustomFlows(name string) ([]CustomFlow, error) {
...
+
+func (client *ApiClient) CustomFlowAssign(assignments []CustomFlowAssignment) error {
+	return client.http.Post("/custom-flow/assign", assignments, nil)
+}
+
+func (client *ApiClient) CustomFlowUnassign(assignments []CustomFlowAssignment) error {
+	return client.http.Post("/custom-flow/unassign", assignments, nil)
+}
+
+func (client *ApiClient) CustomFlowGetAssignments(assignments []CustomFlowAssignment) ([]CustomFlowAssignment, error) {
```
TomerHeber, line 126: BluePrint ID will be empty.
The response will contain the blueprint ID.

client/custom_flow_test.go
```diff
@@ -38,11 +38,8 @@ var _ = Describe("Custom Flow Client", func() {
 				Do(func(path string, request interface{}, response *CustomFlow) {
 					*response = mockCustomFlow
 				})
-			returnedCustomFlow, _ = apiClient.CustomFlow(mockCustomFlow.Id)
-		})
-
-		It("Should send GET request", func() {
 			httpCall.Times(1)
```
TomerHeber, line 41: A little bit confusing from the git diff. The change is that httpCall is now part of the test.
Running in "It" has no affect on the test. (Doesn't test anything).

client/http/client.go
```diff
@@ -54,10 +54,13 @@ func (client *HttpClient) httpResult(response *resty.Response, err error) error
 }
 
 func (client *HttpClient) Post(path string, request interface{}, response interface{}) error {
-	result, err := client.request().
-		SetBody(request).
-		SetResult(response).
-		Post(path)
+	req := client.request().SetBody(request)
```
TomerHeber, line 57: This is to support POST requests with nil responses. (Same implementation as the PUT requests).

client/custom_flow.go
```diff
@@ -35,6 +35,19 @@ type CustomFlow struct {
 	IsTerragruntRunAll   bool             `json:"isTerragruntRunAll"`
 }
 
+type CustomFlowAssignmentScope string
+
+const (
+	CustomFlowOrganizationScope CustomFlowAssignmentScope = "ORGANIZATION"
```
HeverFarber, line 41: For now, you can remove the ability to configure org scope. keep the ability to add it in the future (When we will decide we support it)

HeverFarber (approved):

PR #564, opened 2023-01-07
Feat: custom flow support

TomerHeber:
### Solution

Part of custom flow effort #550 

1. Added the custom flow resource.
2. Added acceptance tests.
3. Added integration tests.


TomerHeber:
For the next PR (once this is approved):

1. Add the assignment resource.
2. Add examples to the documentation.

env0/resource_custom_flow.go
```diff
@@ -0,0 +1,211 @@
...
+				Description: "name for the custom flow",
+				Required:    true,
+			},
+			"repository": {
+				Type:        schema.TypeString,
+				Description: "repository url for the custom flow source code",
+				Required:    true,
+			},
+			"path": {
+				Type:        schema.TypeString,
```
yarivg, line 40: maybe we should indicate in the `path` description that it should be the full  path including .yaml/.yml

TomerHeber: @yarivg - good point. I will fix.

TomerHeber: added.

HeverFarber (approved):
LGTM

PR #569, opened 2023-01-12
Feat: retry on 404 errors on harness testing

### LINKED ISSUE #568: Retry on 404 errors on harness testing
The harness tests are creating and querying multiple resources and data resources at once.
Due to eventual consistency issues, we sometimes get false positive `404` errors from env0 API.

We should add a conditional to the API client retry mechanism we just added that retries specifically on `404` **only** when running the harness tests (and **not** for "production" usage).

TomerHeber:
resolves #568 

### Solution

Passing an environment variable via the integration tests pipeline.
If the environment variable is set - retry on 404 as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,8 +47,6 @@ jobs:
     runs-on: ubuntu-20.04
     container: golang:1.18-alpine3.16
     timeout-minutes: 20
-    env:
-      INTEGRATION_TESTS: 1
     steps:
       - name: Install Terraform
         run: apk add terraform
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,8 @@ jobs:
     runs-on: ubuntu-20.04
     container: golang:1.18-alpine3.16
     timeout-minutes: 20
+    env:
+      INTEGRATION_TESTS: 1
     steps:
       - name: Install Terraform
         run: apk add terraform
--- a/env0/provider.go
+++ b/env0/provider.go
@@ -2,6 +2,7 @@ package env0
 
 import (
 	"context"
+	"os"
 	"time"
 
 	"github.com/env0/terraform-provider-env0/client"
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -47,6 +47,8 @@ jobs:
     runs-on: ubuntu-20.04
     container: golang:1.18-alpine3.16
     timeout-minutes: 20
+    env:
+      INTEGRATION_TESTS: 1
```
TomerHeber, line 51: I tested it manually, and it passes it to the ubuntu machine.

yaronya: Maybe the harness script can set this environment variable? 
I don't see a reason to execute the tests w/o needing this retry

TomerHeber: What do you mean?
It's just enabling retry with 404 for the integration tests...

I'm probably missing something... or misunderstanding the comment.

yaronya: Your current solution forces the harness executor to configure the environment variable in order to have those retries. 

I'm just saying - why not having the harness test itself configure this environment variable?

TomerHeber: I found it cleaner than setting an environment variable in the harness.go code.
Since harness calls (by command execution) terraform, I was also concerned the environment variable would not pass through.
Not sure what's the scope of the environment when setting in GO and then running a terraform command.

(In other words it's me being a little more cautious). Happy to test it out if you believe it's preferred.

yaronya: We're executing the harness tests also internally in env0. My motivation was to avoid needing to set the environment variable on our end. 

I think it's preferred to have it (or any other indicator that's different than an environment variable) defined in the `harness.go` file so it'd indicate that code is running as a part of the harness/integration tests

yaronya: BUT (forgot to say) if you find it too complex then let's leave it as is

TomerHeber: What I could test is setting the environment variable in harness.go

E.g.: from harness.go
```
	err = os.Setenv("TF_CLI_CONFIG_FILE", path.Join(cwd, "tests", "terraform.rc"))
	if err != nil {
		log.Fatalln("Unable to set env:", err)
	}```

I think this will solve the concern you have.

TomerHeber: @yaronya - modified. Setting environment variable each time terraform command is called.

yaronya: Great, thanks
I'll approve but just making sure you validated that the environment variable is available to the `terraform` bin execution 🙏

yaronya (approved):

PR #574, opened 2023-01-19
Fix: if without_template_settings is defined, validate revision is no…

### LINKED ISSUE #573: If `without_template_settings` is defined, validate `revision` is not set outside of it
If `without_template_settings` block is set in your `env0_environment` resource, then setting the `revision` directly in the `resource` (instead of inside the `without_template_settings` block) wouldn't take effect.

For example:

```
resource "env0_environment" "my_environment" {
  name = "my_env"
  project_id = data.env0_project.my_project.id
  revision = "my_branch"

  without_template_settings {
    repository = "https://url/to/repository"
  }
}
```

This will result in an environment using revision `""`, since the revision is not set in `without_template_settings`. 

We should add a validation, making sure `revision` and `without_template_settings` can't be set at the same time

TomerHeber:
…t set outside of it

fixes #573 

### Solution

1. Added conflicts with to schema.
2. Added a unit test.

### DIFF
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -179,10 +179,11 @@ func resourceEnvironment() *schema.Resource {
 				Computed:    true,
 			},
 			"revision": {
-				Type:        schema.TypeString,
-				Description: "the revision the environment is to be run against",
-				Optional:    true,
-				Computed:    true,
+				Type:          schema.TypeString,
+				Description:   "the revision the environment is to be run against",
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"without_template_settings.0.revision"},
 			},
 			"run_plan_on_pull_requests": {
 				Type:        schema.TypeBool,
--- a/env0/resource_environment_test.go
+++ b/env0/resource_environment_test.go
@@ -1531,6 +1531,63 @@ func TestUnitEnvironmentWithoutTemplateResource(t *testing.T) {
 		})
 	})
 
+	t.Run("Revision conflict", func(t *testing.T) {
+		createEnvironmentResourceConfigWithRevision := func(environment client.Environment, template client.Template, revision string) string {
+			return fmt.Sprintf(`
+			resource "%s" "%s" {
+				name = "%s"
+				revision = "%s"
+				project_id = "%s"
+				workspace = "%s"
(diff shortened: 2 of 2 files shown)


RLRabinowitz (approved):
👍🏻

env0/resource_environment.go
```diff
@@ -179,10 +179,11 @@ func resourceEnvironment() *schema.Resource {
...
 			"revision": {
-				Type:        schema.TypeString,
-				Description: "the revision the environment is to be run against",
-				Optional:    true,
-				Computed:    true,
+				Type:          schema.TypeString,
+				Description:   "the revision the environment is to be run against",
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"without_template_settings.0.revision"},
```
RLRabinowitz, line 186: I think we should conflict with `without_template_settings` in general.

If `without_template_settings` block **exists**, the revision shouldn't exist on the resource directly

If there's a `without_template_settings` but no `without_template_settings.revision`, and there's a `revision: master`, the revision used will be `''`

RLRabinowitz (changes requested):
Ignore my previous approval, I've missed something important
The conflict should be with the `without_template_settings` block in general, not with its `revision`

Actually I missed

env0/resource_environment.go
TomerHeber: df5fe0543593a5e9dc6b0d8b563341cfbae48dbf

env0/resource_environment_test.go
```diff
@@ -1531,6 +1531,63 @@ func TestUnitEnvironmentWithoutTemplateResource(t *testing.T) {
...
+				project_id = "%s"
+				workspace = "%s"
+				terragrunt_working_directory = "%s"
+				force_destroy = true
+				vcs_commands_alias = "%s"
+				without_template_settings {
+					repository = "%s"
+					terraform_version = "%s"
+					type = "%s"
+					revision = "%s"
```
RLRabinowitz, line 1549: IMO preferably remove this "revision" from the test
I think it's more important that this test fails when there's no `revision` in the `without_template_settings` at all

RLRabinowitz (approved):
👍🏻 
Left one small comment on the tests

PR #579, opened 2023-01-25
Feat: add support for subenvironments

### LINKED ISSUE #536: Adding sub-environment workspace, revision and variables for workflow environments
Hi @TomerHeber,
We are working on a new feature for our **workflow** environment.
For each workflow environment we want to be able to configure the following attributes for its sub-environments:
1. workspace names
2. revision
3. configuration variables

We added a new field for the environment creation endpoint `/environments/<environment id>` and the redeploy endpoint `/deployments/<deploy id>, called `subEnvironments`.

the  new environment resource should now look like this:

```terraform

resource "env0_environment" "example_with_sub_environments_configuration" {
  name        = "workflow environment with subenvironments"
  project_id  = data.env0_project.default_project.id
  template_id = data.env0_template.example.id
      alias = "sub_env1"
      revision = "sub_env1_branch"
      workspace = "workspace1"
      configuration {
          name          = "sub_env1_var1"
         type          = "environment"
         value         = "hello"
      }
      configuration {
         name          = "sub_env1_var2"
         type          = "environment"
         value         = "world"
      }
  }
        alias = "sub_env2"
        revision = "sub_env2_branch"
        workspace = "workspace2"
        configuration {
            name          = "sub_env2_var1"
           type          = "environment"
           value         = "hello"
        }
        configuration {
           name          = "sub_env2_var2"
           type          = "environment"
           value         = "world"
        }
    }
}
```

Where the `sub_environment_configuration.configuration` is the same type as `environment.configuration`.
We should aggregate all those `sub_environment_configuration`s into one `map[string]subEnvironment` where the `string` keys are the `sub_environment_configuration.name` and the `subEnvironment` value is:
```go
type SubEnvironment struct {
  revision                         string  `json:"revision,omitempty" tfschema:"revision"`
  workspace                    string  `json:"workspace,omitempty" tfschema:"workspace"`
  configurationChanges *configurationsChanges `json:"configurationChanges,omitempty" tfschema:"-"`
}
```

This new `sub_environment_configuration` map should be added to the `DeployRequest`  as:
```typescript
SubEnvironments map[string]subEnvironment `json:"subEnvironments,omitempty"`
```

**Note**: all those changes only apply for workflow environments, meaning environments that were created from a template of type `workflow`

It's a big change so let me know of you have any question.

TomerHeber:
resolves #536 

### Solution
1. Added code for "Update" case. Precisely when a configuration is updated.
2. Update the acceptance test.
3. Added integration test.

### DIFF
--- a/client/environment_test.go
+++ b/client/environment_test.go
@@ -415,5 +415,8 @@ func TestMarshalEnvironmentCreateWithoutTemplate(t *testing.T) {
 	var environmentCreateFromJSON EnvironmentCreate
 
 	require.NoError(t, json.Unmarshal(b, &environmentCreateFromJSON))
-	require.Equal(t, environmentCreate, environmentCreateFromJSON)
+
+	environmentCreateWithType := environmentCreate
+	environmentCreateWithType.Type = templateCreate.Type
+	require.Equal(t, environmentCreateWithType, environmentCreateFromJSON)
 }
--- a/client/template.go
+++ b/client/template.go
@@ -23,13 +23,6 @@ type TemplateRetry struct {
 	OnDestroy *TemplateRetryOn `json:"onDestroy"`
 }
 
-type TemplateType string
-
-const (
-	TemplateTypeTerraform  TemplateType = "terraform"
-	TemplateTypeTerragrunt TemplateType = "terragrunt"
-)
-
 type TemplateSshKey struct {
 	Id   string `json:"id"`
 	Name string `json:"name"`
@@ -70,7 +63,7 @@ type Template struct {
 type TemplateCreatePayload struct {
 	Retry                TemplateRetry    `json:"retry"`
 	SshKeys              []TemplateSshKey `json:"sshKeys,omitempty"`
-	Type                 TemplateType     `json:"type"`
+	Type                 string           `json:"type"`
 	Description          string           `json:"description"`
 	Name                 string           `json:"name"`
 	Repository           string           `json:"repository"`
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -599,7 +599,7 @@ func shouldUpdateTemplate(d *schema.ResourceData) bool {
 }
 
 func shouldDeploy(d *schema.ResourceData) bool {
-	return d.HasChanges("revision", "configuration")
+	return d.HasChanges("revision", "configuration", "sub_environment_configuration")
 }
 
 func shouldUpdate(d *schema.ResourceData) bool {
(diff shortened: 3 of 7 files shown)


env0/resource_environment.go
```diff
@@ -627,6 +627,31 @@ func updateTemplate(d *schema.ResourceData, apiClient client.ApiClientInterface)
...
+				configurationChanges[i].Scope = client.ScopeEnvironment
+			}
+
+			deployPayload.SubEnvironments[subEnvironment.Alias] = client.SubEnvironment{
+				Revision:             subEnvironment.Revision,
+				ConfigurationChanges: configurationChanges,
+			}
+		}
+	}
+
```
eladmosh, line 654: looks good.
Just one question, where do make sure this code only executes for workflow environments?
As regular environments don't have the `subEnvironments` payload.

TomerHeber: If subenvironments are used, I now pass the "workflow" type:
bfd059632207e15afaff0ab9625fe2545b14255c

I can always check the template as well (that it's the workflow type). I'm unsure if we would want to go that far. (I would hope the API would fail?)

client/template.go
```diff
@@ -23,13 +23,6 @@ type TemplateRetry struct {
 	OnDestroy *TemplateRetryOn `json:"onDestroy"`
 }
 
-type TemplateType string
-
-const (
-	TemplateTypeTerraform  TemplateType = "terraform"
```
TomerHeber, line 29: there is more than just two template types.

client/environment_test.go
```diff
@@ -415,5 +415,8 @@ func TestMarshalEnvironmentCreateWithoutTemplate(t *testing.T) {
 	var environmentCreateFromJSON EnvironmentCreate
 
 	require.NoError(t, json.Unmarshal(b, &environmentCreateFromJSON))
-	require.Equal(t, environmentCreate, environmentCreateFromJSON)
+
+	environmentCreateWithType := environmentCreate
+	environmentCreateWithType.Type = templateCreate.Type
```
TomerHeber, line 420: needed to update this test. When create an environment with a template inside it, the environment "inherits" the type from the template.

E.g.: if template inside environment is workflow, the environment type is also workflow.

env0/resource_environment.go
eladmosh: cool, but I don't see you check it in `deploy` before calling `getSubEnvironments`

eladmosh (approved):

PR #582, opened 2023-01-30
Feat: env0_api_key resource and data source don't store apiKeyId

### LINKED ISSUE #581: env0_api_key resource and data source don't store apiKeyId
In the current implementation the api_key resource/data source do not expose the value of apiKeyId from the api response which renders the created keys not directly usable in terraform.

The api response from env0 for api keys include 4 main values of that are required for operation which are actually present in the tests. https://github.com/env0/terraform-provider-env0/blob/main/env0/data_api_key_test.go#L14-L17
```
		Id:           "id0",
		Name:         "name0",
		ApiKeyId:     "keyid0",
		ApiKeySecret: "secret0",
```
If you look at the state for the resources ApiKeyId is dropped and is not exposed. This is the value that is required when trying to use the credentials. Id itself appears as `auth0|somevalue` which while it uniquely identifies the key is not actually part of the key itself. 

resolution: store and expose apiKeyId in a similar fashion as ApiKeySecret on the resource

TomerHeber:
resolves #581 

### Solution
1. added api_key_id to resource and data schemas.
2. updated acceptance tests.

### DIFF
--- a/env0/resource_api_key.go
+++ b/env0/resource_api_key.go
@@ -46,6 +46,11 @@ func resourceApiKey() *schema.Resource {
 				Optional:    true,
 				ForceNew:    true,
 			},
+			"api_key_id": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "the api key id",
+			},
 		},
 	}
 }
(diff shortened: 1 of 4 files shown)


env0/resource_api_key.go
```diff
@@ -66,14 +71,14 @@ func resourceApiKeyCreate(ctx context.Context, d *schema.ResourceData, meta inte
 		return diag.Errorf("could not create api key: %v", err)
 	}
 
+	if err := writeResourceData(apiKey, d); err != nil {
```
TomerHeber, line 74: Using writeResourceData is preferred (less code and more robust).

env0/resource_api_key.go
```diff
@@ -66,14 +71,14 @@ func resourceApiKeyCreate(ctx context.Context, d *schema.ResourceData, meta inte
...
+		return diag.Errorf("schema resource data serialization failed: %v", err)
+	}
+
 	if omit, ok := d.GetOk("omit_api_key_secret"); ok && omit.(bool) {
 		d.Set("api_key_secret", "omitted")
-	} else {
-		d.Set("api_key_secret", apiKey.ApiKeySecret)
```
TomerHeber, line 72: writeResourceData is doing it instead...

env0/resource_api_key.go
```diff
@@ -66,14 +71,14 @@ func resourceApiKeyCreate(ctx context.Context, d *schema.ResourceData, meta inte
...
 	}
 
-	d.SetId(apiKey.Id)
```
TomerHeber, line 75: writeResourceData is doing it instead...

yaronya (approved):

PR #588, opened 2023-02-13
Feat: custom project and organization TTL policy

### LINKED ISSUE #572: Custom Project and Organization TTL Policy
#### Why
Users want to be able to customize the Max and Default TTL for environment  and not to be limited to the current presets

#### How
For [Organization Policy](https://registry.terraform.io/providers/env0/env0/latest/docs/resources/organization_policy) and [Project Policy](https://registry.terraform.io/providers/env0/env0/latest/docs/resources/project_policy) we should change `max_ttl` and `default_ttl` to be any string of type `<number>-<M/w/d/h>` (still omit for infinity) and it should not be limited to the presets.
Those are only user facing changes, we don't want to make any api changes, BE is ok with those kind of strings.
So I think the only requires change is here is to remove the limitation on the allowed strings and just force validation on the `<number>-<M/w/d/h>` schema.

TomerHeber:
resolves #572 

### Solution

1. Updated both resources `env0_organization_policy` and `env0_project_policy` to support custom TTL values.
2. updated acceptance tests.
3. added conversion functions and unit tests.
4. updated integration tests.

### DIFF
--- a/env0/resource_project_policy.go
+++ b/env0/resource_project_policy.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"math"
-	"time"
 
 	"github.com/env0/terraform-provider-env0/client"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
@@ -135,14 +133,6 @@ func resourcePolicyRead(ctx context.Context, d *schema.ResourceData, meta interf
 	return nil
 }
 
-func getPolicyTtl(value string) (time.Duration, error) {
-	if value == "" {
-		return math.MaxInt64, nil
-	}
-
-	return ttlToDuration(value)
-}
-
 func resourcePolicyUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	apiClient := meta.(client.ApiClientInterface)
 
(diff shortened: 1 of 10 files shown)


env0/utils.go
```diff
@@ -453,3 +457,34 @@ func writeResourceDataEx(prefix string, i interface{}, d *schema.ResourceData) e
...
+	number, err := strconv.Atoi(numberStr)
+	if err != nil {
+		return 0, fmt.Errorf("invalid TTL format %s - not a number: %w", ttl, err)
+	}
+	// M/w/d/h
+
+	var hours int = number
+
+	switch rangeType := match[2]; rangeType {
+	case "M":
```
TomerHeber, line 481: Assumed that 1 Month equals 30 days.

env0/utils.go
```diff
@@ -453,3 +457,34 @@ func writeResourceDataEx(prefix string, i interface{}, d *schema.ResourceData) e
 	}
 	return writeResourceDataSlice([]interface{}{i}, prefix, d)
 }
+
+func ttlToDuration(ttl string) (time.Duration, error) {
+	if ttl == "Infinite" || ttl == "inherit" {
```
TomerHeber, line 462: Inherit is hard to predict.
Worst case scenario, the backend call will fail.

env0/resource_project_policy.go
```diff
@@ -152,10 +158,18 @@ func resourcePolicyUpdate(ctx context.Context, d *schema.ResourceData, meta inte
...
+		return diag.Errorf("invalid default ttl: %v", err)
+	}
+
+	maxTtl, err := getPolicyTtl(payload.MaxTtl)
+	if err != nil {
+		return diag.Errorf("invalid max ttl: %v", err)
+	}
+
+	if maxTtl < defaultTtl {
+		return diag.Errorf("default ttl must not be larger than max ttl: %d %d", defaultTtl, maxTtl)
```
eladmosh, line 172: anyway to avoid the code duplication here and in organization?

env0/utils.go
eladmosh, line 462: In case of `infinite` user will omit the ttl, so it will be `nil` and not `Infinite`

env0/utils.go
eladmosh: Isn't there a way to calc a "real" month using https://pkg.go.dev/time?
This will create problems

TomerHeber: 1-M would mean something different every month.
In other words, 1-M isn't well-defined when it's configured.

env0/resource_project_policy.go
TomerHeber: I think there is. I'll check...

env0/utils.go
eladmosh: nvm, just keep it

env0/utils.go
TomerHeber: Generally not an issue, but made some refactoring.
d12f08e06636e169d9c306a715ceb5f687e37d60

env0/resource_project_policy.go
TomerHeber: Refactored and removed code duplication.
d12f08e06636e169d9c306a715ceb5f687e37d60

yaronya (approved):

PR #595, opened 2023-02-16
Fix: property isArchived is sent in patch request even if it was not …

### LINKED ISSUE #594: property `isArchived` is sent in patch request even if it was not changed
### The Problem
when updating an environment, the `isArchived` property is also included in the request. it is not expected by our backend and causes some side effects, for example in remote backend updating

when you try to configure a remote backend environment, and then, for example trying to change any property, it calls the update with `isArchived = false`, and our backend runs some validations that prevent this update. 

### DoD
properties (in this specific case, `isArchived`) will be included in a patch request only if they were truly changed.

TomerHeber:
…changed

fixes #594 

### Solution
1. modified isArchived to be a pointer and check if it's changed before sending. (Environment is a PATCH request).
2. updated acceptance tests.


tomer-landesman (approved):
Looks good to me.

PR #593, opened 2023-02-15
Feat: stop sending external_id for env0_aws_credentials resource

TomerHeber:
### Issue & Steps to Reproduce / Feature Request
fixes #591 

### Solution
1. Removed external_id from all the locations I found,
2. Updated unit tests, examples, acceptance tests, and integration tests.

PR #598, opened 2023-02-20
Chore: deprecate 'external_id'

TomerHeber:
### Solution

Related to #593.
For now just a deprecation message.

### DIFF
--- a/env0/resource_aws_credentials.go
+++ b/env0/resource_aws_credentials.go
@@ -37,7 +37,6 @@ func resourceAwsCredentials() *schema.Resource {
 				Sensitive:     true,
 				ForceNew:      true,
 				ConflictsWith: []string{"access_key_id"},
-				RequiredWith:  []string{"arn"},
 				Deprecated:    "field will be removed in the near future",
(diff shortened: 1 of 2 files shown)


yaronya (approved):

env0/resource_aws_credentials.go
```diff
@@ -38,6 +38,7 @@ func resourceAwsCredentials() *schema.Resource {
 				ForceNew:      true,
 				ConflictsWith: []string{"access_key_id"},
 				RequiredWith:  []string{"arn"},
+				Deprecated:    "field will be removed in the near future",
```
avnerenv0, line 41: this field should also become optional, don't we need to remove
`RequiredWith: []string{"arn"}` ?

TomerHeber: sure I can do that.

TomerHeber: removed the "requiredWith".
for costs I added optional and removed required.

yaronya (approved):

PR #605, opened 2023-02-23
Chore: change info level to warn level

TomerHeber:
### Solution

Modified from "INFO" to "WARN".
Added another print message.

### DIFF
--- a/env0/provider.go
+++ b/env0/provider.go
@@ -144,18 +144,19 @@ func configureProvider(version string, p *schema.Provider) schema.ConfigureConte
 			AddRetryCondition(func(r *resty.Response, err error) bool {
 				if r == nil {
 					// No response. Possiblly a networking issue (E.g. DNS lookup failure).
+					log.Printf("[WARN] No response, retrying request: %s %s", r.Request.Method, r.Request.URL)
 					return true
 				}
 
 				// When running integration tests 404 may occur due to "database eventual consistency".
 				// Retry when there's a 5xx error. Otherwise do not retry.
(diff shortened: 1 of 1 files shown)


env0/provider.go
```diff
@@ -144,18 +144,19 @@ func configureProvider(version string, p *schema.Provider) schema.ConfigureConte
...
 				// Retry when there's a 5xx error. Otherwise do not retry.
 				if r.StatusCode() >= 500 || isIntegrationTest && r.StatusCode() == 404 {
-					log.Printf("[INFO] Received %d status code, retrying request: %s %s", r.StatusCode(), r.Request.Method, r.Request.URL)
+					log.Printf("[WARN] Received %d status code, retrying request: %s %s", r.StatusCode(), r.Request.Method, r.Request.URL)
 					return true
 				}
 
 				if r.StatusCode() == 200 && isIntegrationTest && r.String() == "[]" {
-					log.Printf("[INFO] Received an empty list for an integration test, retrying request: %s %s", r.Request.Method, r.Request.URL)
+					log.Printf("[WARN] Received an empty list for an integration test, retrying request: %s %s", r.Request.Method, r.Request.URL)
```
yaronya, line 159: Can you please explain how this works exactly? Cause this is just text 😅

TomerHeber: I think (just based on how it's been done so far in the provider) that Terraform knows how to parse it.
Maybe it's wrong? 

E.g.:
https://github.com/hashicorp/terraform-provider-aws/search?q=log.Print

yaronya: I really don't know
Let's try and if it won't work - we'll tackle it again

yaronya (approved):

PR #607, opened 2023-02-26
Fix: env0 variables that are mark is_sensitive = true are shown in pl…

### LINKED ISSUE #600: env0 variables that are mark `is_sensitive = true` are shown in plan/apply as plain text
```
resource "env0_environment" "this" {
  ...
  configuration {
    name = "TF_TOKEN_app_terraform_io"
    value = <plain_token>
    type = "environment"
    description = "Terraform cloud registry token."
    is_sensitive = true
    is_read_only = true
  }
}
Terrafom plan always shows:
      ~ configuration {
            name = "TF_TOKEN_app_terraform_io"
          ~ value = "***************" -> "<plain_token>"
            # (7 unchanged attributes hidden)
        }
It causes false positive drift detections in the environment where this resource belongs.
```

sensitive values shouldn't cause wrong plan output

TomerHeber:
…an/apply as plain text

fixes #600 

### Solution
Fixed several issues/bugs I found while reviewing and testing the code. This should align better with how it's done in the UI.
Made changes to avoid drift for sensitive values.

Acceptance tests updated.


env0/resource_environment.go
```diff
@@ -524,16 +532,20 @@ func resourceEnvironmentCreate(ctx context.Context, d *schema.ResourceData, meta
...
 	}
+
 	if err != nil {
 		return diag.Errorf("could not create environment: %v", err)
 	}
+
 	environmentConfigurationVariables := client.ConfigurationChanges{}
-	if environmentPayload.DeployRequest.ConfigurationChanges != nil {
-		environmentConfigurationVariables = *environmentPayload.DeployRequest.ConfigurationChanges
+	if environmentPayload.ConfigurationChanges != nil {
```
TomerHeber, line 541: The way it was done before isn't correct.
There may be a use-case for `*environmentPayload.DeployRequest.ConfigurationChanges`, but I couldn't find it while playing with the UI.

env0/resource_environment.go
```diff
@@ -849,11 +861,9 @@ func getDeployPayload(d *schema.ResourceData, apiClient client.ApiClientInterfac
 		payload.BlueprintRevision = revision.(string)
 	}
 
-	if configuration, ok := d.GetOk("configuration"); ok {
+	if configuration, ok := d.GetOk("configuration"); ok && isRedeploy {
```
TomerHeber, line 864: Playing with the UI... it's only applicable for a redeploy. In the past, it was different?! Maybe?

env0/resource_environment_test.go
```diff
@@ -431,6 +430,152 @@ func TestUnitEnvironmentResource(t *testing.T) {
 			})
 		})
 
+		t.Run("Create and redeploy configuration variables - sensitive values", func(t *testing.T) {
```
TomerHeber, line 433: There were not tests for sensitive configurations.
Added one. Tests both creation and redeploy/update.

env0/resource_environment.go
sagilaufer1992: we can have configurations for initial deploy, doesn't that change break things in that scenario?

env0/resource_environment_test.go
sagilaufer1992: the test checks that the environment can be created and redeployed with sensitive values? it doesn't check anything regarding the issue at hand, right?

sagilaufer1992 (commented):
looks good, but I have two places I'm not sure about them, would appreciate a clarification, thanks

env0/resource_environment_test.go
TomerHeber: The issue was about sensitive values.
Without the fixes in this PR, the test would fail during `plan`.

env0/resource_environment.go
TomerHeber: This is for "create" use-case.
The "redploy change" I made a bit below would now make this always nil.

Looking at how it works in the UI: the deploy request does not contain the configuration. Just a few fields related to the blueprint/template. This makes sense because all the configurations are already set when creating the environment.

env0/resource_environment.go
TomerHeber: The "update" operation calls the function "deploy" which calls "getDeployPayload" with isRedploy = true.
The "create" operation calls the function "getCreatePayload" which calls "getDeployPayload" with isRedploy = false.

In initial deploy "create" use-case, configuration is passed during the create of the environment, not during the first deploy (as I observed when using the UI + API).

env0/resource_environment.go
sagilaufer1992: looked at the payload, got your point, the variables are set in their own field, not in the field `deployRequest`. thanks

sagilaufer1992 (approved):
my questions had been answered fully, the code looks good, approved 👍

PR #611, opened 2023-03-02
Fix: workflow environment test used bad templates

tomer-landesman:

we need to remove this test for 2 main reasons:

1. the test hasn't worked till now. for the test to work, we need to create a template for the sub-environments, and its template name should be randomized to avoid flakiness issues that we are seeing from time to time. 
 
2. it doesn't really give us confidence. since we performed the deployed lambda split, the provider can get 200 status on a deploy that only entered the queue but was not necessarily deployed successfully. this test, for example, was failing to deploy because it tried to deploy the sub-environments with non-existing templates, yet the test passed. 

this test gave us no confidence as it is passing even 
### Solution
removed workflow tests for now.


tests/integration/012_environment/main.tf
```diff
@@ -19,6 +19,14 @@ resource "env0_template" "template" {
   terraform_version = "0.15.1"
 }
 
+resource "env0_template" "workflow_subenv_null_resource" {
+  name              = "workflow-subenv-null-resource"
```
liranfarage89, line 23: please add a `random` value to the name, to prevent flakiness issues, like if the destruction fails we can't create another template with the same name.

tomer-landesman: we can't do it here. the subenvironment template is hardcoded in the workflow file. can you suggest a workaround for that? we can also remove this test while we figuring our a solution

liranfarage89: I'm going into an interview... won't be available RN

liranfarage89: Ok, I see that it works since #579, let's keep it in meanwhile...

liranfarage89 (approved):

PR #620, opened 2023-03-05
Fix: project's parent_project_id does not work

### LINKED ISSUE #617: Project's `parent_project_id` does not work
### Issue
The project's resource `parent_project_id` does not work. When creating a resource with it, the `parentProjectId` is not sent in the `POST /projects` request

### Steps to reproduce

1. Create a project with a `parent_project_id`
```
data "env0_project" "root" {
  name = "Root Project"
}

resource "env0_project" "sub" {
  name = "Sub Project"
  parent_project_id = data.env0_project.root.id
}
```
2. The "Sub Project" project is created, but not as a sub project of the "Root Project". The `POST /projects` API call does not contain the `parentProjectId`

TomerHeber:
fixes #617 

### Solution
1. Modified create project to pass payload instead of a custom map.
2. Updated HTTP client tests for create project.
3. Tested manually.

### DIFF
--- a/client/project.go
+++ b/client/project.go
@@ -49,8 +49,15 @@ func (client *ApiClient) ProjectCreate(payload ProjectCreatePayload) (Project, e
 		return Project{}, err
 	}
 
-	request := map[string]interface{}{"name": payload.Name, "organizationId": organizationId, "description": payload.Description}
-	err = client.http.Post("/projects", request, &result)
+	payloadWith := struct {
+		ProjectCreatePayload
+		OrganizationId string `json:"organizationId"`
+	}{
+		payload,
+		organizationId,
+	}
+
+	err = client.http.Post("/projects", payloadWith, &result)
 	if err != nil {
 		return Project{}, err
 	}
--- a/client/project_test.go
+++ b/client/project_test.go
@@ -9,6 +9,7 @@ import (
 
 const projectName = "project_test"
 const projectDescription = "project description"
+const parentProjectId = "parent_project_id"
 
 var _ = Describe("Project", func() {
 	var project Project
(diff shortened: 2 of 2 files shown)


client/project.go
```diff
@@ -49,8 +49,15 @@ func (client *ApiClient) ProjectCreate(payload ProjectCreatePayload) (Project, e
 		return Project{}, err
 	}
 
-	request := map[string]interface{}{"name": payload.Name, "organizationId": organizationId, "description": payload.Description}
-	err = client.http.Post("/projects", request, &result)
+	payloadWith := struct {
```
TomerHeber, line 52: Doing it this way will prevent future bugs and is best practice.

client/project.go
```diff
@@ -49,8 +49,15 @@ func (client *ApiClient) ProjectCreate(payload ProjectCreatePayload) (Project, e
...
+		ProjectCreatePayload
+		OrganizationId string `json:"organizationId"`
+	}{
+		payload,
+		organizationId,
+	}
+
+	err = client.http.Post("/projects", payloadWith, &result)
```
TomerHeber, line 60: passing just the payload... instead of a custom map...

client/project.go
RLRabinowitz, line 52: Maybe we should have a better name for it? `payloadWithOrganizationId`, or something of the sort?

RLRabinowitz (approved):

client/project_test.go
```diff
@@ -23,20 +24,28 @@ var _ = Describe("Project", func() {
...
-					"organizationId": organizationId,
-					"description":    projectDescription,
+			payload := struct {
+				ProjectCreatePayload
+				OrganizationId string `json:"organizationId"`
+			}{
+				ProjectCreatePayload{
+					Name:            projectName,
+					Description:     projectDescription,
+					ParentProjectId: parentProjectId,
```
sagilaufer1992, line 34: why not adding a case for a parent project, and keep the one without? both cases are valid, we just need to make sure that the one with parent ALSO works

sagilaufer1992 (commented):
regarding the test - we now only test creation with project id? no checking for a null project id? or did I misunderstand this change?

client/project_test.go
TomerHeber: I'm not sure there will be added value in this.
I added it because the test would have failed without my fix. But with the fix there's little value with way.

PR #625, opened 2023-03-09
chore: recreate workflow environment test

tomer-landesman:

we wanted to recreate the workflow environment test with a new solution that would work correctly

### Solution

fixed the workflow template and introduced `data` to the `null resource` pre created template to be more explicit about what is going on. 

I made sure it is running successfully against BORS stage:
image


tests/integration/012_environment/main.tf
```diff
@@ -113,3 +113,62 @@ resource "env0_environment" "inactive" {
...
+resource "env0_template" "workflow_template" {
+  name              = "Template for workflow environment-${random_string.random.result}"
+  type              = "workflow"
+  repository        = "https://github.com/env0/templates"
+  path              = "misc/workflow-environment-basic"
+  terraform_version = "1.1.5"
+}
+
+data "env0_template" "sub_environment_null_template" {
+  name = "null resource"
```
tomer-landesman, line 128: added that to make it more explicit that we are using a pre created template

tomer-landesman:
[We have no useful provider tests for workflow environment](https://trello.com/c/LIZhaJIR/9110-we-have-no-useful-provider-tests-for-workflow-environment)

TomerHeber (approved):

PR #627, opened 2023-03-22
Feat: workflow trigger new resources

### LINKED ISSUE #580: workflow trigger new resources
We need to add another resource to WorkFlow-Trigger, our workflow trigger today is calling api which update the triggers by overriding the former list, for example - former list = [1,2,3,4] => update list with(5) => new trigger list = [5].

We have users who manage multiple stacks with triggers from the same environment, so this creates concurrency issues for them.

We created new endpoints for "atmoic" updates to the triggers list, and we should create a *new resource* called `workflow_trigger` which will represent a single trigger. 

The relevant endpoints are 
[Subscribe environments as workflow-trigger](https://docs.env0.com/reference/environments-subscribe-downstream-environments)
[Unsubscribe environments as workflow-trigger](https://docs.env0.com/reference/environments-unsubscribe-downstream-environments)

TomerHeber:
resolves #580 

### Solution

1. added the resource.
2. added acceptance tests.
3. added integration tests.
4. added an example.

### DIFF
--- a/env0/resource_workflow_trigger.go
+++ b/env0/resource_workflow_trigger.go
@@ -1,0 +1,90 @@
+package env0
+
+import (
+	"context"
+	"log"
+
+	"github.com/env0/terraform-provider-env0/client"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func resourceWorkflowTrigger() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: resourceWorkflowTriggerCreate,
+		ReadContext:   resourceWorkflowTriggerRead,
+		DeleteContext: resourceWorkflowTriggerDelete,
+		Description:   "cannot be used with env0_workflow_triggers",
+
+		Schema: map[string]*schema.Schema{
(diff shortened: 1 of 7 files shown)


env0/resource_workflow_trigger.go
```diff
@@ -0,0 +1,90 @@
...
+func resourceWorkflowTriggerRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	apiClient := meta.(client.ApiClientInterface)
+
+	environmentId := d.Get("environment_id").(string)
+	downstreamEnvironmentId := d.Get("downstream_environment_id").(string)
+
+	triggers, err := apiClient.WorkflowTrigger(environmentId)
+
+	if err != nil {
+		return diag.Errorf("could not get workflow triggers: %v", err)
```
ItamarMalka, line 45: is it not possible that when there are no workflow triggers we are drifted? 🤔

ItamarMalka (approved):
LGTM left one small question 👍

env0/resource_workflow_trigger.go
ItamarMalka: so if someone deleted the trigger we would not recognize it as drift but add it? 🤔

PR #639, opened 2023-04-21
Feat: ability to filter for parent project name in data project resource

### LINKED ISSUE #630: ability to filter for parent project name in data project resource
example: 
I have the following project structure

"product-a" > "dev"
                     > "stage"
                     > "prod"
"product-b" > "dev"
                     > "stage"
                     > "prod"
                     
The data resource to find "product-a/dev" project

so something like adding a parent project filter would be great.

TomerHeber:
resolves #630 

### Solution
1. Added the new filter to schema.
2. Added acceptance tests.
3. Added integration tests.

### DIFF
--- a/env0/data_project.go
+++ b/env0/data_project.go
@@ -31,7 +31,7 @@ func dataProject() *schema.Resource {
 			},
 			"parent_project_name": {
 				Type:        schema.TypeString,
-				Description: "the name of the parent project. To be used when there are multiple subprojects with the same name under different parent projects",
+				Description: "the name of the parent project. To be used only when there are multiple subprojects with the same name under different parent projects",
 				Optional:    true,
 			},
 			"created_by": {
--- a/env0/data_project.go
+++ b/env0/data_project.go
@@ -87,27 +87,23 @@ func dataProjectRead(ctx context.Context, d *schema.ResourceData, meta interface
 }
 
 func filterByParentProjectName(name string, parentName string, projects []client.Project, meta interface{}) ([]client.Project, error) {
-	if len(parentName) > 0 {
-		filteredProjects := make([]client.Project, 0)
-		for _, project := range projects {
-			if len(project.ParentProjectId) == 0 {
-				continue
-			}
-
-			parentProject, err := getProjectById(project.ParentProjectId, meta)
-			if err != nil {
-				return nil, err
-			}
-
-			if parentProject.Name == parentName {
-				filteredProjects = append(filteredProjects, project)
-			}
+	filteredProjects := make([]client.Project, 0)
+	for _, project := range projects {
+		if len(project.ParentProjectId) == 0 {
+			continue
 		}
 
-		projects = filteredProjects
+		parentProject, err := getProjectById(project.ParentProjectId, meta)
+		if err != nil {
+			return nil, err
+		}
+
+		if parentProject.Name == parentName {
+			filteredProjects = append(filteredProjects, project)
+		}
 	}
 
-	return projects, nil
+	return filteredProjects, nil
 }
 
 func getProjectByName(name string, parentName string, meta interface{}) (client.Project, error) {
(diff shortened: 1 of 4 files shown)


env0/data_project.go
```diff
@@ -81,7 +86,31 @@ func dataProjectRead(ctx context.Context, d *schema.ResourceData, meta interface
 	return nil
 }
 
-func getProjectByName(name interface{}, meta interface{}) (client.Project, error) {
+func filterByParentProjectName(name string, parentName string, projects []client.Project, meta interface{}) ([]client.Project, error) {
+	if len(parentName) > 0 {
```
yaronya, line 90: It'd be cleaner to take this condition out of this function, as this function should be called only if using the parent name attribute

TomerHeber: @yaronya - cleaned it up. It should make more sense now.

env0/data_project_test.go
```diff
@@ -28,6 +29,40 @@ func TestProjectDataSource(t *testing.T) {
...
+	projectWithParent := client.Project{
+		Id:              "id123",
+		Name:            "same_name",
+		CreatedBy:       "env0",
+		Role:            "role0",
+		Description:     "A project's description",
+		ParentProjectId: parentProject.Id,
+	}
+
+	otherProjectWithParent := client.Project{
```
sagilaufer1992, line 57:
```suggestion
	otherProjectWithOtherParent := client.Project{
```

env0/data_project_test.go
sagilaufer1992, line 1: I think there's a case missing - only one result for project by name, but the parent name is also given, but still no match. the case you have that is similar, has 2 candidates, but none with the same parent name given, I think the code will return a single project, that has the name, but not the parent project given in the data_resource, which seems like a bug

sagilaufer1992 (commented):
Looks great! I think there's a case missing in the test, and some clarification is required in another, too many `expects` that are very similar, not sure what each block is for

env0/data_project_test.go
```diff
@@ -96,6 +131,45 @@ func TestProjectDataSource(t *testing.T) {
...
+					mock.EXPECT().Project(otherProjectWithParent.ParentProjectId).Times(1).Return(otherParentProject, nil),
+					mock.EXPECT().Projects().Times(1).Return([]client.Project{projectWithParent, otherProjectWithParent}, nil),
+					mock.EXPECT().Project(projectWithParent.ParentProjectId).Times(1).Return(parentProject, nil),
+					mock.EXPECT().Project(otherProjectWithParent.ParentProjectId).Times(1).Return(otherParentProject, nil),
+					mock.EXPECT().Projects().Times(1).Return([]client.Project{projectWithParent, otherProjectWithParent}, nil),
+					mock.EXPECT().Project(projectWithParent.ParentProjectId).Times(1).Return(parentProject, nil),
+					mock.EXPECT().Project(otherProjectWithParent.ParentProjectId).Times(1).Return(otherParentProject, nil),
+					mock.EXPECT().Projects().Times(1).Return([]client.Project{projectWithParent, otherProjectWithParent}, nil),
+					mock.EXPECT().Project(projectWithParent.ParentProjectId).Times(1).Return(parentProject, nil),
+					mock.EXPECT().Project(otherProjectWithParent.ParentProjectId).Times(1).Return(otherParentProject, nil),
```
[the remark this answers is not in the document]
TomerHeber: it's just the way these type of tests work. I modified it to "anytime" and now it's just 3 lines.

env0/data_project_test.go
sagilaufer1992: In the case that there are two projects with the same name, one with the parent that fits the filter, it will be returned. In case the project is deleted, the other project is returned, as it fits the original name filter, but the parent filter isn't applied, and the user will have an unexpected project as a result. If that's an unreasonable case, leave it, but it seems like an opening for bugs IMO

TomerHeber: sure! Fair enough. I'll fix that.

TomerHeber: This is now fixed.

PR #647, opened 2023-05-07
Fix: Workflow environment creation integration test

Yossi-kerner:
We want to prevent creating sub environments with the same workspace name and template name but since we can't explicitly set workspace name in the sub environment resource

### Solution
have a workflow template with a single environment


tomer-landesman (approved):
Looks good to me, make sure to test with the new template before you merge it

PR #652, opened 2023-05-12
Feat: allow editing sub project parentId so it uses our new move proj…

### LINKED ISSUE #645: Feat: Allow editing sub project parentId so it uses our new move project api
We implemented a new [API](https://docs.env0.com/reference/projects-move-project) to support moving sub projects so now it is possible to edit the sub project parentId so the project would get moved to a different project tree...

### Expected DoD
We should allow editing the parentProjectId and call the API with the project parameters so it gets moved, deleting the parentId should send 'null' (which means the project would be changed to a root project which doesn't have a parentId)

TomerHeber:
…ect api

resolves #645 

### Solution

1. Added the move API call.
2. Updated the project resource schema and the update function.
3. Added unit tests for the move API call.
4. Updated acceptance tests for project resources.
5. Updated the integration tests to cover the 'move' use case.

### DIFF
--- a/client/project_test.go
+++ b/client/project_test.go
@@ -76,7 +76,7 @@ var _ = Describe("Project", func() {
 	Describe("ProjectUpdate", func() {
 		var mockedResponse Project
 		BeforeEach(func() {
-			payload := ProjectCreatePayload{
+			payload := ProjectUpdatePayload{
 				Name:        "newName",
 				Description: "newDesc",
 			}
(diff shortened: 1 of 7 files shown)


env0/resource_project.go
```diff
@@ -110,7 +109,18 @@ func resourceProjectUpdate(ctx context.Context, d *schema.ResourceData, meta int
 	apiClient := meta.(client.ApiClientInterface)
 
 	id := d.Id()
-	var payload client.ProjectCreatePayload
+	var payload client.ProjectUpdatePayload
+
+	if d.HasChange("parent_project_id") {
+		parentProjectId := d.Get("parent_project_id").(string)
+		if parentProjectId == "" {
```
TomerHeber, line 116: This was not supported in the UI.
If it's valid to make a subproject a project (with no parent), let me know, and I will modify this.

ItamarMalka: lmk if that makes sense to you or if there are any further questions 👍

client/project_test.go
```diff
@@ -148,4 +148,21 @@ var _ = Describe("Project", func() {
...
+		}{
+			targetProjectId,
+		}
+
+		BeforeEach(func() {
+			httpCall = mockHttpClient.EXPECT().Post("/projects/"+mockProject.Id+"/move", payload, nil).Times(1)
+			apiClient.ProjectMove(mockProject.Id, targetProjectId)
+		})
+
+		It("Should send POST request with project id and target project id", func() {})
```
ItamarMalka, line 166: empty case here? 🤔

env0/resource_project.go
TomerHeber: makes sense. I'll fix that.

TomerHeber: @ItamarMalka - I fixed it: updated code, tests, and integration tests.
Now support moving a subproject to become a project.

ItamarMalka (approved):
QAed it manually and it worked great
LGTM, Great work 👏

PR #658, opened 2023-06-09
Feat: helm template support + helm repo vcs option support

### LINKED ISSUE #655: helm template support + helm repo vcs option support
Support newly added template type - `helm`
this template requires some extra field which should only be available for this new template type:

Attaching an example when using standard VCS option (no new fields, just a different template type):
```hcl
resource "env0_template" "helm" {
  name               = "example" 
  description        = "Example template with helm" 
  repository         = "https://github.com/env0/templates",
  path: 'misc/helm/dummy',
  type: 'helm',
}
```

Attaching an example when using the new `helm repo` VCS option (different template type, 2 new fields at the bottom):
```hcl
resource "env0_template" "helm" {
  name               = "example" 
  description        = "Example template with helm" 
  repository         = "https://github.com/env0/templates",
  type: 'helm',
  helm_chart_name: 'my-chart'
  is_helm_repository: true
}
```

About `is_helm_repository`:
1. it is similar to `Other VCS`, no new fields for that repo type
2. no `path` property
3. it should only be available when the template type is `helm`, it cannot be used with other types
4. the new `helm_chart_name` should only be passed when `is_helm_repository` is set to true

TomerHeber:
resolves #655 

### Solution

1. Updated API structs.
2. Added required fields in the schema.
3. Added unit tests.
4. Added integration tests.

### DIFF
--- a/client/template.go
+++ b/client/template.go
@@ -144,6 +144,16 @@ func (payload TemplateCreatePayload) Validate() error {
 		return fmt.Errorf("file_name cannot be set when template type is: %s", payload.Type)
 	}
 
+	if payload.IsHelmRepository {
+		if payload.Type != "helm" {
+			return errors.New(`can't set is_helm_repository to "true" for non-helm template`)
+		}
+
+		if payload.HelmChartName == "" {
+			return errors.New("helm_chart_name is required with helm repository")
+		}
+	}
+
 	return nil
 }
 
--- a/client/template.go
+++ b/client/template.go
@@ -37,7 +37,7 @@ type Template struct {
 	Name                 string           `json:"name"`
 	Description          string           `json:"description"`
 	OrganizationId       string           `json:"organizationId"`
-	Path                 string           `json:"path"`
+	Path                 string           `json:"path,omitempty" tfschema:",omitempty"`
 	Revision             string           `json:"revision"`
 	ProjectId            string           `json:"projectId"`
 	ProjectIds           []string         `json:"projectIds"`
@@ -69,7 +69,7 @@ type TemplateCreatePayload struct {
 	Description          string           `json:"description"`
 	Name                 string           `json:"name"`
 	Repository           string           `json:"repository"`
-	Path                 string           `json:"path"`
+	Path                 string           `json:"path,omitempty"`
 	IsGitLab             bool             `json:"isGitLab"`
 	TokenName            string           `json:"tokenName"`
 	TokenId              string           `json:"tokenId,omitempty"`
@@ -144,13 +144,17 @@ func (payload TemplateCreatePayload) Validate() error {
 		return fmt.Errorf("file_name cannot be set when template type is: %s", payload.Type)
 	}
 
-	if payload.IsHelmRepository {
+	if payload.IsHelmRepository || payload.HelmChartName != "" {
 		if payload.Type != "helm" {
 			return errors.New(`can't set is_helm_repository to "true" for non-helm template`)
 		}
 
 		if payload.HelmChartName == "" {
 			return errors.New("helm_chart_name is required with helm repository")
+		}
+
+		if !payload.IsHelmRepository {
+			return errors.New(`is_helm_repositroy set to "true" is required with "helm_chart_name"`)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


env0/resource_template_test.go
```diff
@@ -365,6 +365,50 @@ func TestUnitTemplateResource(t *testing.T) {
...
+			OnDeploy: &client.TemplateRetryOn{
+				Times:      2,
+				ErrorRegex: "RetryMeForDeploy.*",
+			},
+			OnDestroy: &client.TemplateRetryOn{
+				Times:      1,
+				ErrorRegex: "RetryMeForDestroy.*",
+			},
+		},
+		TerraformVersion: "0.12.24",
```
tomporat247, line 388: why `TerraformVersion`? you are testing helm templates

env0/resource_template.go
```diff
@@ -208,6 +211,20 @@ func getTemplateSchema(prefix string) map[string]*schema.Schema {
 			ConflictsWith: allVCSAttributesBut("is_azure_devops", "token_id"),
 			RequiredWith:  requiredWith("token_id"),
 		},
+		"helm_chart_name": {
+			Type:          schema.TypeString,
+			Optional:      true,
+			Description:   "the helm chart name. Required if is_helm_repository is set to 'true'",
+			ConflictsWith: allVCSAttributesBut("helm_chart_name", "is_helm_repository"),
+		},
+		"is_helm_repository": {
```
tomporat247, line 220: the `path` property isn't used when `is_helm_repository` is true.

could we block setting it?

tomporat247, line 220: the `is_helm_repository` could only be set when the `type` is `helm`.

could we enforce that?

env0/resource_template.go
```diff
@@ -208,6 +211,20 @@ func getTemplateSchema(prefix string) map[string]*schema.Schema {
...
+			Type:          schema.TypeBool,
+			Optional:      true,
+			Description:   "true if this template integrates with a helm repository",
+			Default:       "false",
+			ConflictsWith: allVCSAttributesBut("helm_chart_name", "is_helm_repository"),
+			RequiredWith:  requiredWith("helm_chart_name"),
```
tomporat247, line 226: I'm not really in the details about the `RequiredWith` and `ConflictsWith` logic but overall - one cannot be defined without the other, right?

tomporat247 (commented):
Looking good overall.
Asked some questions and wrote a few comments

env0/resource_template.go
TomerHeber: Not through the schema. But I can add a manual test. I'll make sure to add it.

env0/resource_template.go
TomerHeber: correct.
It's even more than that. For example, the other "is_*_" cannot be set with it. (allVCSAttributesBut).

env0/resource_template_test.go
TomerHeber: This is a limitation in the acceptance tests. The use the same flow for all test-cases (Saves a lot of repetitive code).
I can fix it (by adding some 'ifs'), but I'm not sure it's worth the effort.

TomerHeber:
@tomporat247 - thanks for all the feedback. added a new commit 72008cca007d91dae1197902440e5c0687b7dd5b

Regarding your comment here: https://github.com/env0/terraform-provider-env0/pull/658#discussion_r1225734899
I did not make any changes to that. But if you think it's important, I can modify that.

env0/resource_template_test.go
tomporat247: we spoke on slack, we'll open a new issue for this matter

tomporat247 (approved):
Looking good overall, I left one more comment but I think it's not too important as that problem probably happens in other use cases here as well

TomerHeber:
> Note that your unit tests are failing

Yes. I will be pushing more changes.

client/template.go
```diff
@@ -140,6 +144,16 @@ func (payload TemplateCreatePayload) Validate() error {
 		return fmt.Errorf("file_name cannot be set when template type is: %s", payload.Type)
 	}
 
+	if payload.IsHelmRepository {
```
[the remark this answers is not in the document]
TomerHeber: added additional conditionals.

Issue #654, opened 2023-05-24
Environment Role assignments for User/Team


razbensimon:
Terraform Provider **resource**/**data** for the new Environment Role Assignments

Expected Result / DoD ✅:

Terraform Provider resource/data for the new "Environment Role Assignments".

create new OR use an existing resource we have, but just use the new APIs while demand from the user to give environmentId instead of projectId)

There are 2 different assignments - User / Team

Use the new [APIs](https://docs.env0.com/reference/roles-get-user-role-assignments): Get, Upsert, Delete

PR #660, opened 2023-06-16
Feat: user role environment assignment

TomerHeber:
### Solution

Part of #654 

Added user - environment assignment. (team assignment will follow up).

1. Updated DELETE rest call to allow params (required for the new API calls).
2. Added assignment API calls.
3. Added unit tests for new API calls.
4. Added resource schema.
5. Added acceptance tests to new resource.
6. Added integration tests.
7. Added examples.

### DIFF
--- a/env0/resource_user_environment_assignment.go
+++ b/env0/resource_user_environment_assignment.go
@@ -29,9 +29,9 @@ func resourceUserEnvironmentAssignment() *schema.Resource {
 				Required:    true,
 				ForceNew:    true,
 			},
-			"custom_role_id": {
+			"role_id": {
 				Type:             schema.TypeString,
-				Description:      "id of the assigned custom role",
+				Description:      "id of the assigned role",
 				Required:         true,
 				ValidateDiagFunc: ValidateNotEmptyString,
 			},
@@ -74,7 +74,7 @@ func resourceUserEnvironmentAssignmentRead(ctx context.Context, d *schema.Resour
 				return diag.Errorf("schema resource data serialization failed: %v", err)
 			}
 
-			d.Set("custom_role_id", assignment.Role)
+			d.Set("role_id", assignment.Role)
 
 			return nil
 		}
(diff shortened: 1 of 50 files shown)


examples/resources/env0_user_environment_assignment/resource.tf
```diff
@@ -0,0 +1,32 @@
...
+
+resource "env0_api_key" "api_key" {
+  name              = "user-api-key-sample"
+  organization_role = "User"
+}
+
+resource "env0_custom_role" "custom_role" {
+  name = "custom-role-sample"
+  permissions = [
+    "VIEW_PROJECT",
```
yarivg, line 23: I would change the given permission to environment permissions to be more real

env0/resource_user_environment_assignment.go
```diff
@@ -0,0 +1,118 @@
...
+	return &schema.Resource{
+		CreateContext: resourceUserEnvironmentAssignmentCreate,
+		UpdateContext: resourceUserEnvironmentAssignmentUpdate,
+		ReadContext:   resourceUserEnvironmentAssignmentRead,
+		DeleteContext: resourceUserEnvironmentAssignmentDelete,
+
+		Schema: map[string]*schema.Schema{
+			"user_id": {
+				Type:        schema.TypeString,
+				Description: `id of the user. Note: can also be an id of a "User" API key`,
```
yarivg, line 22: are you sure about the note? haven't tried it with 'user' api-keys

yarivg (commented):
Went over most of it and left some comments.. not approving since I don't know go @yaronya 
@TomerHeber

yarivg (approved):
The new API for assignment of role to a user on environment seems to be written well
👍

env0/resource_user_environment_assignment.go
TomerHeber: I'll remove the comment.

PR #665, opened 2023-06-29
Feat: add support for drift detection creation for environment resource

### LINKED ISSUE #664: Add support for drift detection creation for `environment` resource
The API supports the creation of a drift detection setting when creating an environment (See `driftDetectionRequest` in API docs)
We should add this support for the `environment` resource.

TomerHeber:
resolves #664 

### Solution
1. Added a new drift_detection_cron field to the environment schema and implemented create and update use-cases.
2. Updated acceptance tests.
3. Updated integration tests.

### DIFF
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -316,6 +316,12 @@ func resourceEnvironment() *schema.Resource {
 					},
 				},
 			},
+			"drift_detection_cron": {
+				Type:             schema.TypeString,
+				Description:      "cron expression for scheduled drift detection of the environment (cannot be used with resource_drift_detection resource)",
+				Optional:         true,
+				ValidateDiagFunc: ValidateCronExpression,
+			},
 		},
 
 		CustomizeDiff: customdiff.ForceNewIf("template_id", func(ctx context.Context, d *schema.ResourceDiff, meta interface{}) bool {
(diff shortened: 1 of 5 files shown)


env0/resource_environment.go
```diff
@@ -316,6 +316,12 @@ func resourceEnvironment() *schema.Resource {
 					},
 				},
 			},
+			"drift_detection_cron": {
```
TomerHeber, line 319: made more sense to be just one field. No point in making it a complex schema.

env0/resource_environment.go
```diff
@@ -650,6 +665,24 @@ func updateTemplate(d *schema.ResourceData, apiClient client.ApiClientInterface)
 	return nil
 }
 
+func updateDriftDetection(d *schema.ResourceData, apiClient client.ApiClientInterface) diag.Diagnostics {
+	drift_detection_cron, ok := d.GetOk("drift_detection_cron")
+	if !ok || drift_detection_cron.(string) == "" {
+		if err := apiClient.EnvironmentStopDriftDetection(d.Id()); err != nil {
```
TomerHeber, line 671: for updates there's stop drift API and update drift API.
No option to update directly from the environment API.

yaronya: I think that it was fine not supporting update or delete altogether because the UI also doesn't support it, but since this was implemented already and may provide better UX - let's keep it 😄

yaronya (approved):

PR #682, opened 2023-07-20
Fix: cannot create env0_user_team_assignment: validation error

### LINKED ISSUE #681: Cannot create env0_user_team_assignment: validation error
My userId contains an `_` in it and for that reason I'm unable to create an env0_user_team_assignment resource.
The following function fails because two underlines are present in the string.
https://github.com/env0/terraform-provider-env0/blob/9db74b01fa8a46142d8768195a2303fa64c4144c/env0/resource_user_team_assignment.go#L36-L45

TomerHeber:
fixes #681 

### Solution

Some user ids have an "_". Original solution assumed it's not possible.

1. created a custom split called lastSplit: split by the last underscore.
2. use lastSplit instead of using strings.Split().
3. Updated the acceptance tests to cover this use-case.
4. Added unit tests for lastSplit.

### DIFF
--- a/env0/resource_user_team_assignment.go
+++ b/env0/resource_user_team_assignment.go
@@ -33,6 +33,7 @@ func (a *UserTeamAssignment) GetId() string {
 }
 
 func GetUserTeamAssignmentFromId(id string) (*UserTeamAssignment, error) {
+	// lastSplit is used to avoid issues where the user_id has underscores in it.
 	splitUserTeam := lastSplit(id, "_")
 	if len(splitUserTeam) != 2 {
 		return nil, fmt.Errorf("the id %v is invalid must be <user_id>_<team_id>", id)
--- a/env0/resource_user_team_assignment.go
+++ b/env0/resource_user_team_assignment.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"log"
-	"strings"
 	"sync"
 
 	"github.com/env0/terraform-provider-env0/client"
@@ -34,7 +33,7 @@ func (a *UserTeamAssignment) GetId() string {
 }
 
 func GetUserTeamAssignmentFromId(id string) (*UserTeamAssignment, error) {
-	splitUserTeam := strings.Split(id, "_")
+	splitUserTeam := lastSplit(id, "_")
 	if len(splitUserTeam) != 2 {
 		return nil, fmt.Errorf("the id %v is invalid must be <user_id>_<team_id>", id)
 	}
--- a/env0/resource_user_team_assignment_test.go
+++ b/env0/resource_user_team_assignment_test.go
@@ -13,7 +13,7 @@ import (
 )
 
 func TestUnitUserTeamAssignmentResource(t *testing.T) {
-	userId := "uid"
+	userId := "ui_d"
 	teamId := "tid"
 
 	GenerateTeam := func(teamId string, userIds []string) client.Team {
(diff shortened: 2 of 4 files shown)


env0/resource_user_team_assignment.go
```diff
@@ -34,7 +33,7 @@ func (a *UserTeamAssignment) GetId() string {
 }
 
 func GetUserTeamAssignmentFromId(id string) (*UserTeamAssignment, error) {
-	splitUserTeam := strings.Split(id, "_")
+	splitUserTeam := lastSplit(id, "_")
```
TomerHeber, line 36: At the moment this is the only code where import is used with a combination of "user".
If more combinations are added need to be sure lastSplit is used.

env0/resource_user_team_assignment.go
```diff
@@ -34,7 +33,8 @@ func (a *UserTeamAssignment) GetId() string {
 }
 
 func GetUserTeamAssignmentFromId(id string) (*UserTeamAssignment, error) {
-	splitUserTeam := strings.Split(id, "_")
+	// lastSplit is used to avoid issues where the user_id has underscores in it.
+	splitUserTeam := lastSplit(id, "_")
 	if len(splitUserTeam) != 2 {
```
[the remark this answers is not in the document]
TomerHeber: in golang `len(nil)` returns `0`.

env0/utils.go
```diff
@@ -489,3 +489,12 @@ func ttlToDuration(ttl *string) (time.Duration, error) {
 
 	return time.ParseDuration(fmt.Sprintf("%dh", hours))
 }
+
+func lastSplit(s string, sep string) []string {
+	lastIndex := strings.LastIndex(s, sep)
+	if lastIndex == -1 {
+		return nil
```
sagilaufer1992, line 496: why `nil`? the original split would return the original string, no? that might confuse when using that method.
in the case of this fix, the outcome will remain that the length of the desired array won't be 2, so IMO it's better to stick to known outcomes

env0/resource_user_team_assignment.go
sagilaufer1992: just to make sure - team id is some kind of guid, so it won't have `_` in it, so we won't have to consider this case as well?

sagilaufer1992 (commented):
looks good overall, left some comments

env0/utils.go
TomerHeber: fair point. I'll change that.

env0/resource_user_team_assignment_test.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 func TestUnitUserTeamAssignmentResource(t *testing.T) {
-	userId := "uid"
+	userId := "ui_d"
```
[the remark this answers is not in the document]
TomerHeber: I can duplicate the test with a different value.

env0/resource_user_team_assignment.go
TomerHeber: AFAIK yes - a uuid.
userid is not a uuid.

env0/resource_user_team_assignment_test.go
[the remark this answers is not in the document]
TomerHeber: both use-cases are now covered (added another test).

env0/utils.go
TomerHeber: fixed.

sagilaufer1992 (approved):
LGTM, my comments were addressed

sagilaufer1992:
please try and comment your commit hash that resolves a comment, it makes reviewing easier 😃

PR #687, opened 2023-08-05
Feat: approval policy support

### LINKED ISSUE #671: Approval Policy Support
# Add support for Approval Policy configuration

```
resource "env0_approval_policy" "example" {
  name                   = "example"
  repository             = "example"
  path                   = "example"
  revision               = "example"
  token_id               = 0
  ssh_keys               = []
  gitlab_project_id      = 0
  github_installation_id = 0
  bitbucket_client_key   = 0
  is_bitbucket_server    = "true/false"
  is_gitlab_enterprise   = "true/false"
  is_github_enterprise   = "true/false"
  is_git_lab             = "true/false"
  is_azure_devOps        = "true/false"
}

resource "env0_approval_policy_assignment" "assignment_example" {
  depends_on = [env0_approval_policy.example]
  scope      = "PROJECT"
  scopeId    = "example"
  templateId = env0_approval_policy.example.id
}
```

## Approval Policy API
note that create ,update and delete doesn’t require a special endpoint, it is the same as a template, only the type should be `approval-policy`
- [Create](https://docs.env0.com/reference/templates-create-blueprint)
- [Update](https://docs.env0.com/reference/templates-update-blueprint)

*IMPORTANT*
the name should NOT be a user field and should be set on our side as `approval-policy-{scope}-{scopeId}`

### Get approval policy by name
- [Get By Name](https://docs.env0.com/reference/templates-get-approval-policy-by-name)

### assignments
- [Assign](https://docs.env0.com/reference/templates-assign-approval-policy)
- [Unassign](https://docs.env0.com/reference/templates-delete-approval-policy-assignment)

TomerHeber:
resolves #671 

### Solution

1. Added the approval policy client APIs.
2. Added unit tests for the APIs.
3. Added the approval policy resource schema.
4. Added acceptance tests for the approval policy resource.
5. Added the approval policy assignment resource schema.
6. Added acceptance tests for the approval policy assignment resource.
7. Added integration test.
8. Added examples.


env0/common_template.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+type TemplateType string
+
+const (
+	CustomFlow     TemplateType = "custom flow"
+	ApprovalPolicy TemplateType = "approval policy"
+)
+
+func getTemplate(templateType TemplateType) map[string]*schema.Schema {
```
TomerHeber, line 16: common code - moved to a common function.

env0/common_template.go
```diff
@@ -0,0 +1,118 @@
...
+		},
+		"is_azure_devops": {
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: fmt.Sprintf("true if this %s integrates with azure dev ops repository", templateType),
+			Default:     false,
+		},
+	}
+
+	if templateType == CustomFlow {
```
TomerHeber, line 100: name differes between custom flow and approval policy.

env0/common_template.go
```diff
@@ -0,0 +1,118 @@
...
+		s["name"] = &schema.Schema{
+			Type:        schema.TypeString,
+			Description: "name for the custom flow. note: for the UI to render the custom-flow please use `project-<project.id>`",
+			Required:    true,
+		}
+	}
+
+	if templateType == ApprovalPolicy {
+		s["name"] = &schema.Schema{
```
TomerHeber, line 109: I am allowing users to set names.
However, they must do it in the requested format.
During assignment it will fail if they do not do it in the required format and if there are any mismatches.

env0/resource_custom_flow.go
```diff
@@ -20,93 +20,7 @@ func resourceCustomFlow() *schema.Resource {
 
 		Importer: &schema.ResourceImporter{StateContext: resourceCustomFlowImport},
 
-		Schema: map[string]*schema.Schema{
```
TomerHeber, line 23: moved to a common code (above).

client/approval_policy.go
```diff
@@ -0,0 +1,77 @@
+package client
+
+import "fmt"
+
+type ApprovalPolicy struct {
+	Id                   string           `json:"id"`
```
tomer-landesman, line 6: not sure if possible with golang, but can this type be derived from the type of a tamplate? I guess the fields are exactly the same so its basically code duplication. not mandatory if this is the way we are usually doing things

TomerHeber: There are some ways. But generally, there's no inheritance.
Applying that change will add more changes to an already large PR.

env0/resource_approval_policy.go
```diff
@@ -0,0 +1,149 @@
...
+
+func resourceApprovalPolicyCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	apiClient := meta.(client.ApiClientInterface)
+
+	var payload client.TemplateCreatePayload
+	if err := readResourceData(&payload, d); err != nil {
+		return diag.Errorf("schema resource data deserialization failed: %v", err)
+	}
+
+	payload.Type = APPROVAL_POLICY
```
TomerHeber, line 37: implicitly set the template type.

env0/resource_approval_policy.go
```diff
@@ -0,0 +1,149 @@
...
+
+func resourceApprovalPolicyUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	apiClient := meta.(client.ApiClientInterface)
+
+	request, problem := templateCreatePayloadFromParameters("", d)
+	if problem != nil {
+		return problem
+	}
+
+	request.Type = APPROVAL_POLICY
```
TomerHeber, line 78: implicitly set the template type.

client/approval_policy.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func (client *ApiClient) ApprovalPolicyAssign(assignment *ApprovalPolicyAssignment) (*ApprovalPolicyAssignment, error) {
+	var result ApprovalPolicyAssignment
+
+	if err := client.http.Post("/approval-policy/assignment", assignment, &result); err != nil {
+		return nil, err
+	}
+
+	return &result, nil
```
tomer-landesman, line 62: why do you return by reference in this endpoint only?

env0/common_template.go
tomer-landesman: I'm not sure about that part @HeverFarber can you take a look and suggest if its a good product decision in light of your latest changes?

env0/common_template.go
```diff
@@ -0,0 +1,118 @@
...
+			Optional:    true,
+			Description: fmt.Sprintf("true if this %s integrates with azure dev ops repository", templateType),
+			Default:     false,
+		},
+	}
+
+	if templateType == CustomFlow {
+		s["name"] = &schema.Schema{
+			Type:        schema.TypeString,
+			Description: "name for the custom flow. note: for the UI to render the custom-flow please use `project-<project.id>`",
```
tomer-landesman, line 103: name for the `approval policy`

env0/common_template.go
```diff
@@ -0,0 +1,118 @@
...
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+type TemplateType string
+
+const (
+	CustomFlow     TemplateType = "custom flow"
```
tomer-landesman, line 12: please make the types like we do in our own backend
`custom-flow`
`approval-policy`

env0/resource_approval_policy.go
```diff
@@ -0,0 +1,149 @@
...
+	"fmt"
+	"log"
+
+	"github.com/env0/terraform-provider-env0/client"
+	"github.com/google/uuid"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+const APPROVAL_POLICY = "approval-policy"
```
tomer-landesman, line 14: I'd suggest to extract this type in the common template, and use the same one in every place (same for custom flow)

env0/common_template.go
HeverFarber: don't need to force the template name. let the user insert whatever name he wants

env0/resource_approval_policy.go
```diff
@@ -0,0 +1,149 @@
...
+
+func resourceApprovalPolicyRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	apiClient := meta.(client.ApiClientInterface)
+
+	approvalPolicy, err := apiClient.Template(d.Id())
+	if err != nil {
+		return ResourceGetFailure("approval policy", d, err)
+	}
+
+	if approvalPolicy.IsDeleted && !d.IsNewResource() {
```
tomer-landesman, line 57: I'm not sure I's deleted is a property in Template. is it something that is added by the provider / TF?

tests/integration/030_approval_policy/main.tf
```diff
@@ -0,0 +1,27 @@
+provider "random" {}
```
tomer-landesman, line 1: general question - is updating a resource ever tested in the provider?

tomer-landesman, line 1: general question - is updating a resource ever tested in the provider?

env0/common_template.go
tomer-landesman: thanks! so @TomerHeber lets remove this assertion 👍🏼

client/approval_policy.go
TomerHeber: you mean by pointer.
In golang a slice/array is already a pointer.

So in all cases pointers are returned instead of values.

env0/resource_approval_policy.go
TomerHeber: sure!

env0/resource_approval_policy.go
TomerHeber: I copied it from the Template resource (just to be on the safe side). Do you think it's not required?

tests/integration/030_approval_policy/main.tf
TomerHeber: yes. In the acceptance tests.

TomerHeber: yes. In the acceptance tests.
I can add something to the integration test.

tests/integration/030_approval_policy/main.tf
```diff
@@ -16,9 +16,11 @@ data "env0_template" "github_template" {
 }
 
 resource "env0_approval_policy" "test" {
-  name                   = "approval-policy-PROJECT-${env0_project.project.id}"
+  name                   = "approval-policy-${random_string.random.result}"
   repository             = data.env0_template.github_template.repository
   github_installation_id = data.env0_template.github_template.github_installation_id
+  path                   = var.second_run ? "second" : "misc/null-resource"
```
TomerHeber, line 22: for testing updates.

env0/resource_approval_policy_assignment.go
```diff
@@ -56,13 +56,8 @@ func resourceApprovalPolicyAssignmentCreate(ctx context.Context, d *schema.Resou
 		return diag.Errorf("unable to get template with id %s: %v", blueprintId, err)
 	}
 
-	if template.Type != APPROVAL_POLICY {
```
TomerHeber: cna now be any name...

env0/common_template.go
```diff
@@ -9,20 +9,29 @@ import (
...
+	CustomFlow     TemplateType = "custom-flow"
+	ApprovalPolicy TemplateType = "approval-policy"
 )
 
 func getTemplate(templateType TemplateType) map[string]*schema.Schema {
+	var text string
+
+	switch templateType {
+	case CustomFlow:
+		text = "custom flow"
```
tomer-landesman, line 21: forgot a dash - is it on purpose to keep backward compatibility for something?

env0/common_template.go
```diff
@@ -9,20 +9,29 @@ import (
...
+	case ApprovalPolicy:
+		text = "approval-policy"
```
tomer-landesman, line 23: why the word "text"? its very general - whats wrong with template type?
btw - in our FE we decided on a convention that both of those template will be called "Configuration Template" because they have some similarities (they are configuration that the user define through a template)
maybe using this term can benefit you in some places

client/approval_policy.go
tomer-landesman: thanks for the explenation! I just wandered why it is needed only in `ApprovalPolicyAssign` and not in `ApprovalPolicyUnassign` for example.

env0/common_template.go
tomer-landesman: maybe even the name of the file could be "configuration_template_something" because you are using it only for custom-flow and approval-policy and not for a real template

env0/resource_approval_policy.go
tomer-landesman: oh I see. just checked out code, its logic that I wasn't aware of... lets keep is for safe measures 👍🏼

tomer-landesman (approved):
Looks good to me! nice job :)

env0/common_template.go
TomerHeber: It's for documentation.
I actually need to remove the dash from the other one. (In other words, I don't want to the dash in the documentation).

env0/common_template.go
TomerHeber: sure!

PR #700, opened 2023-08-27
Fix: 'endless drift' when setting template path to start with '/'

### LINKED ISSUE #699: BUG: "Endless Drift" when setting template path to start with "/"
Steps to reproduces:

1.  Create a template with path which start with "/"
```
resource "env0_template" "github_template1" {
  name                                    = "Github Test Templates"
  description                             = "template-1"
  type                                    = "terraform"
  repository                              = data.env0_template.github_template.repository
  github_installation_id                  = data.env0_template.github_template.github_installation_id
  path                                    = "/misc/null-resource". **<<<----------------**
  retries_on_deploy                       = 3
  retry_on_deploy_only_when_matches_regex = "abc"
  retries_on_destroy                      = 1
  terraform_version                       = "0.15.1"
}
```
2. plan and apply
3. changed nothing and plan again and see drift

its caused by the fact that the BE removes the "/" from the start of the path.

suggestion solution:
Ignore '/' from the path in the provider remove it on read

TomerHeber:
fixes #699 

### Solution

Keeping the state with "/" to avoid endless drift. Ignoring the value returned by the template if only "/" was omitted.
Added an acceptance test for this use-case and updated the integration test.

### DIFF
--- a/env0/resource_template.go
+++ b/env0/resource_template.go
@@ -393,8 +393,20 @@ func templateCreatePayloadFromParameters(prefix string, d *schema.ResourceData) 
 
 // Reads template and writes to the resource data.
 func templateRead(prefix string, template client.Template, d *schema.ResourceData) error {
+	pathPrefix := "path"
+	if prefix != "" {
+		pathPrefix = prefix + ".0.path"
+	}
+
+	path, pathOk := d.GetOk(pathPrefix)
+
 	if err := writeResourceDataEx(prefix, &template, d); err != nil {
 		return fmt.Errorf("schema resource data serialization failed: %v", err)
+	}
+
+	// https://github.com/env0/terraform-provider-env0/issues/699 - backend removes the "/".
+	if pathOk && path.(string) == "/"+template.Path {
+		d.Set(pathPrefix, path.(string))
 	}
 
 	templateReadRetryOnHelper(prefix, d, "deploy", template.Retry.OnDeploy)
--- a/env0/resource_template_test.go
+++ b/env0/resource_template_test.go
@@ -1006,6 +1006,84 @@ func TestUnitTemplateResource(t *testing.T) {
 		})
 	})
 
+	// https://github.com/env0/terraform-provider-env0/issues/699
+	t.Run("path drift", func(t *testing.T) {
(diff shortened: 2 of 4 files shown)


env0/resource_template.go
```diff
@@ -393,10 +393,22 @@ func templateCreatePayloadFromParameters(prefix string, d *schema.ResourceData)
 
 // Reads template and writes to the resource data.
 func templateRead(prefix string, template client.Template, d *schema.ResourceData) error {
+	pathPrefix := "path"
+	if prefix != "" {
+		pathPrefix = prefix + ".0.path"
+	}
+
+	path, pathOk := d.GetOk(pathPrefix)
```
TomerHeber, line 401: save the current value

env0/resource_template.go
```diff
@@ -393,10 +393,22 @@ func templateCreatePayloadFromParameters(prefix string, d *schema.ResourceData)
...
+
 	if err := writeResourceDataEx(prefix, &template, d); err != nil {
```
TomerHeber, line 403: this updated to the new value

env0/resource_template.go
```diff
@@ -393,10 +393,22 @@ func templateCreatePayloadFromParameters(prefix string, d *schema.ResourceData)
...
 		return fmt.Errorf("schema resource data serialization failed: %v", err)
 	}
 
+	// https://github.com/env0/terraform-provider-env0/issues/699 - backend removes the "/".
+	if pathOk && path.(string) == "/"+template.Path {
```
TomerHeber, line 408: if diff is just in "/" revert to old path - this avoids a drift and keeps state intact with what the user entered.

env0/resource_template_test.go
```diff
@@ -1006,6 +1006,84 @@ func TestUnitTemplateResource(t *testing.T) {
 		})
 	})
 
+	// https://github.com/env0/terraform-provider-env0/issues/699
+	t.Run("path drift", func(t *testing.T) {
```
TomerHeber, line 1010: test fails without the fix.

alonnoga (approved):
LGTM ty for the comments

PR #701, opened 2023-08-27
Fix: do not send terraformVersion unused field

### LINKED ISSUE #659: Do not send terraformVersion unused field
## Problem Overview
Our API is very forgiving in this matter, if you pass the terraform version field while you are in pulumi for example it doesn't have any effect - it's just not being used.

Theoretically you can send `terraformVersion` and `plulumiVersion` at the same request and env0 will accept it and just not use them.
However we shouldn't do that as:
1. That data can confuse us while we are debugging
2. Generally you want to avoid sending wrong data even when unused
3. This can create a bug if someone will introduce a new code that assumes stuff based on those variables (and logically he will be correct)

## Expected Solution
Do not send incorrect unused fields.
`terraformVersion` should not be sent for non terraform/terragrunt templates.

I guess there are more cases which we should fix too, so if you can think of some please fix them as well.

TomerHeber:
fixes #659 

### Solution
1. added omitempty and emptied terraform version if the template type is not "terraform" or "terragrunt".
2. tests updated accordingly.

### DIFF
--- a/client/template.go
+++ b/client/template.go
@@ -49,7 +49,7 @@ type Template struct {
 	IsGitlabEnterprise   bool             `json:"isGitLabEnterprise"`
 	TokenId              string           `json:"tokenId,omitempty" tfschema:",omitempty"`
 	UpdatedAt            string           `json:"updatedAt"`
-	TerraformVersion     string           `json:"terraformVersion"`
+	TerraformVersion     string           `json:"terraformVersion,omitempty"`
 	TerragruntVersion    string           `json:"terragruntVersion,omitempty" tfschema:",omitempty"`
 	IsDeleted            bool             `json:"isDeleted,omitempty"`
 	BitbucketClientKey   string           `json:"bitbucketClientKey" tfschema:",omitempty"`
@@ -111,7 +111,7 @@ type VariablesFromRepositoryPayload struct {
 	Repository           string   `json:"repository"`
 }
 
-func (payload TemplateCreatePayload) Validate() error {
+func (payload *TemplateCreatePayload) Invalidate() error {
 	if payload.OrganizationId != "" {
 		return errors.New("must not specify organizationId")
 	}
@@ -157,6 +157,10 @@ func (payload TemplateCreatePayload) Validate() error {
 		if !payload.IsHelmRepository {
 			return errors.New(`is_helm_repositroy set to "true" is required with "helm_chart_name"`)
 		}
+	}
+
+	if payload.Type != "terragrunt" && payload.Type != "terraform" {
+		payload.TerraformVersion = ""
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


client/template.go
```diff
@@ -78,7 +78,7 @@ type TemplateCreatePayload struct {
 	GitlabProjectId      int              `json:"gitlabProjectId,omitempty"`
 	Revision             string           `json:"revision"`
 	OrganizationId       string           `json:"organizationId"`
-	TerraformVersion     string           `json:"terraformVersion"`
+	TerraformVersion     string           `json:"terraformVersion,omitempty"`
```
sagydr, line 81: Hi @TomerHeber , I'm Sagy from RND, nice to github-meet you  😃 
on purpose you don't have the `tfschema:",omitempty"` syntax here also?

env0/resource_template_test.go
```diff
@@ -625,6 +626,11 @@ func TestUnitTemplateResource(t *testing.T) {
 				HelmChartName:        templateUseCase.template.HelmChartName,
 			}
 
+			if templateUseCase.template.Type != "terraform" && templateUseCase.template.Type != "terragrunt" {
+				templateCreatePayload.TerraformVersion = ""
+				updateTemplateCreateTemplate.TerraformVersion = ""
```
sagydr, line 631: Just out of curiosity as I'm not too familiar with Go, `templateCreatePayload` & `updateTemplateCreateTemplate` have many common fields, is it not common to extract them to something like BaseTemplate struct? Like in Python?

sagydr (approved):
Added some questions, but code looks straight-forward

client/template.go
```diff
@@ -111,7 +111,7 @@ type VariablesFromRepositoryPayload struct {
 	Repository           string   `json:"repository"`
 }
 
-func (payload TemplateCreatePayload) Validate() error {
+func (payload *TemplateCreatePayload) Invalidate() error {
```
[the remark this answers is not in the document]
TomerHeber: Hi @sagydr - ahh that's a golang question (:
without "*" is readonly function - you can't modify "this".
with "*" you can modify "this".

env0/resource_template_test.go
TomerHeber: To add on this: Golang does not have inheritance and is not object oriented.
https://go.dev/doc/faq#:~:text=Types-,Is%20Go%20an%20object%2Doriented%20language%3F,in%20some%20ways%20more%20general.

client/template.go
TomerHeber: Hi @sagydr - nice to meet you (:

tfschema omitempty is only needed when reading from a struct and writing to the terraform schema.
For this struct we read from the terraform schema and write to the struct.

When transforming to a json, the json omitempty will be used and the field will not be sent.
(if string == "" then the field isn't passed in the json).

Happy to explain more if this is still unclear. (Didn't want the explanation to be too long).

PR #708, opened 2023-09-11
Fix: env0_git_token recreating in every tf apply

### LINKED ISSUE #707: env0_git_token recreating in every tf apply
**Describe the bug**
The env0_git_token resource recreates the Git token on each run, even if the value does not change.

**To Reproduce**
Run terraform apply with the following resource:

resource "env0_git_token" "token" {
  name  = "git_token"
  value = <token_value>
}

After the resource is created, run terraform apply again.

**Expected behavior**
The apply must indicate that there is no changes to be applied.

**Provider Version**
1.14.15

**Screenshots**
image

**Additional context**

TomerHeber:
fixes #707 

### Solution

1. "value" returns *****. Therefore, it causes a drift.
2. It is now ignored (except for creation).

### DIFF
--- a/client/git_token.go
+++ b/client/git_token.go
@@ -3,7 +3,7 @@ package client
 type GitToken struct {
 	Id             string `json:"id"`
 	Name           string `json:"name"`
-	Value          string `json:"value"`
+	Value          string `json:"value" tfschema:"-"`
 	OrganizationId string `json:"organizationId"`
 }
 
--- a/env0/resource_git_token_test.go
+++ b/env0/resource_git_token_test.go
@@ -108,10 +108,11 @@ func TestUnitGitTokenResource(t *testing.T) {
 					}),
 				},
 				{
-					ResourceName:      resourceNameImport,
-					ImportState:       true,
-					ImportStateId:     gitToken.Name,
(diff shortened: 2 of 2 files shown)


client/git_token.go
```diff
@@ -3,7 +3,7 @@ package client
 type GitToken struct {
 	Id             string `json:"id"`
 	Name           string `json:"name"`
-	Value          string `json:"value"`
+	Value          string `json:"value" tfschema:"-"`
```
Wassap124, line 6: can you send a doc / explanation as to why setting a field's schema to `-` works? 

tried to find an explanation [here](https://github.com/minamijoyo/tfschema) but couldn't really find any

TomerHeber: tfschema is something internal we wrote.
See https://github.com/env0/terraform-provider-env0/blob/main/DEVELOPMENT.md

`Pay attention to the following caveats:...`

Wassap124: > tfschema is something internal we wrote.

Oh i thought this refers to this https://github.com/minamijoyo/tfschema

Wassap124 (approved):

PR #711, opened 2023-09-19
Feat: add support for Duration for AWS Assume Roles in env0_aws_crede…

### LINKED ISSUE #710: Add support for Duration for AWS Assume Roles in env0_aws_credentials resource
env0 aws assume role credentials now support a `Duration` field.
please add support in TF provider.

see: https://docs.env0.com/reference/configuration-create-credential
duration-in-seconds

I believe default is 5 hours (18,000 secs)

add a variable for duration (in seconds)

🤷 

**Additional context**
n/a

TomerHeber:
…ntials resource

resolves #710 

### Solution

1. Updated the schema.
2. Updated the API.
3. Updated the unit, acceptance and integration tests.

### DIFF
--- a/client/cloud_credentials_test.go
+++ b/client/cloud_credentials_test.go
@@ -80,7 +80,8 @@ var _ = Describe("CloudCredentials", func() {
 			mockOrganizationIdCall(organizationId)
 
 			payloadValue := AwsCredentialsValuePayload{
-				RoleArn: "role",
+				RoleArn:  "role",
+				Duration: 1,
 			}
 
 			httpCall = mockHttpClient.EXPECT().
@@ -120,7 +121,8 @@ var _ = Describe("CloudCredentials", func() {
 			mockOrganizationIdCall(organizationId)
 
 			payloadValue := AwsCredentialsValuePayload{
-				RoleArn: "role",
+				RoleArn:  "role",
+				Duration: 1,
 			}
 
 			httpCall = mockHttpClient.EXPECT().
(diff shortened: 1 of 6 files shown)


client/cloud_credentials_test.go
```diff
@@ -80,7 +80,8 @@ var _ = Describe("CloudCredentials", func() {
 			mockOrganizationIdCall(organizationId)
 
 			payloadValue := AwsCredentialsValuePayload{
-				RoleArn: "role",
+				RoleArn:  "role",
+				Duration: 1,
```
sagilaufer1992, line 84: It's not one of the allowed values, shouldn't the test fail on this?

env0/resource_cost_credentials_test.go
sagilaufer1992, line 1: I see only test that is updated with duration.
Can you test also creation with duration please?
And also create and update with illegal values?

If these cases are tested somewhere else please point me there

sagilaufer1992 (commented):
Looks good. Have some comments about the tests.

client/cloud_credentials_test.go
TomerHeber: This is enforced at the schema/resource level. Not the API level.

env0/resource_cost_credentials_test.go
TomerHeber: Happy to add these.
I'll do it a bit later today.

TomerHeber: @sagilaufer1992 sorry for the delay.
I have added the requested use-cases.

sagilaufer1992: LGTM

sagilaufer1992 (approved):
LGTM, cases added as requested in other comment

PR #712, opened 2023-09-19
Chore: Document how to disallow remote state access

project-kmac:
From customer:

> │ Error: could not create a remote state access configation: 400 Bad Request: /body must have required property 'allowedProjectIds', /body/accessibleFromEntireOrganization must be equal to constant, /body must match a schema in anyOf
seems silly, but the only way i found to disallow remote state access for env0 backend with terraform is to fill the allowed project id list with an string like "none":
resource "env0_environment_state_access" "disallow" {
  accessible_from_entire_organization = false
  allowed_project_ids = ["none"]
skipping allowed_project_ids or setting to [] yields the given error

### DIFF
--- a/docs/resources/environment_state_access.md
+++ b/docs/resources/environment_state_access.md
@@ -42,7 +42,8 @@ resource "env0_environment_state_access" "example_entire_organization" {
 ### Optional
 
 - `accessible_from_entire_organization` (Boolean) when this parameter is 'false', allowed_project_ids should be provided. Defaults to 'false'
-- `allowed_project_ids` (List of String) list of allowed project_ids. Used when 'accessible_from_entire_organization' is 'false'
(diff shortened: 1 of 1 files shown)


docs/resources/environment_state_access.md
```diff
@@ -42,7 +42,8 @@ resource "env0_environment_state_access" "example_entire_organization" {
 ### Optional
 
 - `accessible_from_entire_organization` (Boolean) when this parameter is 'false', allowed_project_ids should be provided. Defaults to 'false'
-- `allowed_project_ids` (List of String) list of allowed project_ids. Used when 'accessible_from_entire_organization' is 'false'
+- `allowed_project_ids` (List of String) list of allowed project_ids. Used when 'accessible_from_entire_organization' is 'false'. To disallow remote state access for env0 backend, fill the allowed_project_id list with string "none". Skipping allowed_project_ids or setting to [] yields a 400 error. (e.g. ```accessible_from_entire_organization = false
```
avnerenv0, line 45: @project-kmac this file is automatically generated and you can't change.
You'll need to make this change in https://github.com/env0/terraform-provider-env0/blob/51748f71e05ea2e52c30cf352a048bcb589de34d/env0/resource_environment_state_access.go#L33C31-L33C31 please.

avnerenv0 (changes requested):

TomerHeber:
will fix the issue instead.

PR #714, opened 2023-09-20
Fix: disallowing remote state access returns 400 bad request

### LINKED ISSUE #713: Disallowing remote state access returns 400 bad request
**Describe the bug**
│ Error: could not create a remote state access configation: 400 Bad Request: /body must have required property 'allowedProjectIds', /body/accessibleFromEntireOrganization must be equal to constant, /body must match a schema in anyOf
seems silly, but the only way i found to disallow remote state access for env0 backend with terraform is to fill the allowed project id list with an string like "none":
resource "env0_environment_state_access" "disallow" {
accessible_from_entire_organization = false
allowed_project_ids = ["none"]
skipping allowed_project_ids or setting to [] yields the given error

TomerHeber:
fixes #713 

### Solution
Removed the omit empty and updated the integration test.

### DIFF
--- a/client/remote_state_access.go
+++ b/client/remote_state_access.go
@@ -8,7 +8,7 @@ type RemoteStateAccessConfiguration struct {
 
 type RemoteStateAccessConfigurationCreate struct {
 	AccessibleFromEntireOrganization bool     `json:"accessibleFromEntireOrganization"`
-	AllowedProjectIds                []string `json:"allowedProjectIds,omitempty"`
+	AllowedProjectIds                []string `json:"allowedProjectIds"`
 }
 
 func (client *ApiClient) RemoteStateAccessConfiguration(environmentId string) (*RemoteStateAccessConfiguration, error) {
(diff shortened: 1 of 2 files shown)


tests/integration/012_environment/main.tf
```diff
@@ -76,10 +76,11 @@ resource "env0_team_environment_assignment" "team_role_environment_assignment" {
   role_id        = var.second_run ? env0_custom_role.custom_role1.id : env0_custom_role.custom_role2.id
 }
 
-/* TODO: need to add an integration test.
-resource "env0_environment_state_access" "state_access" {
-  environment_id      = env0_environment.example.id
-  allowed_project_ids = [env0_project.test_project.id]
+/* TODO
```
TomerHeber, line 79: I uncommented this to check that the error went away.
It still fails because of the environment type used (but the original error is now gone).

client/remote_state_access.go
```diff
@@ -22,6 +22,10 @@ func (client *ApiClient) RemoteStateAccessConfiguration(environmentId string) (*
 }
 
 func (client *ApiClient) RemoteStateAccessConfigurationCreate(environmentId string, payload RemoteStateAccessConfigurationCreate) (*RemoteStateAccessConfiguration, error) {
+	if payload.AllowedProjectIds == nil {
```
TomerHeber, line 25: If an array is not created explicitly, "null" is passed. The API doesn't accept "null". "[]" is needed instead.

PR #697, opened 2023-08-22
Fix: make custom roles data test isolated

GiliFaroEnv0:
### Issue & Steps to Reproduce / Feature Request
[//]: <> (choose “resolves” “fixes” or “closes” and put link for the issue)
since we run our tests in parallel other tests that create custom_role resources interfere with this test
### Solution
Isolate the test and only use resource from the same test

PR #721, opened 2023-10-04
Fix: make cloud_credentials data test isolated

GiliFaroEnv0:
Issue & Steps to Reproduce / Feature Request
since we run our tests in parallel other tests that create `env0_aws_credentials` resources interfere with this test

Solution
Isolate the test and only use resources from the same test

### DIFF
--- a/tests/integration/024_cloud_credentials/main.tf
+++ b/tests/integration/024_cloud_credentials/main.tf
@@ -21,7 +21,7 @@ resource "env0_gcp_credentials" "gcp_cred" {
 }
 
 data "env0_cloud_credentials" "all_aws_credentials" {
-  depends_on = [env0_aws_credentials.aws_cred1, aws_cred2,env0_gcp_credentials.gcp_cred]
+  depends_on = [env0_aws_credentials.aws_cred1, env0_aws_credentials.aws_cred2, env0_gcp_credentials.gcp_cred]
   credential_type = "AWS_ASSUMED_ROLE_FOR_DEPLOYMENT"
 }
 
--- a/tests/integration/024_cloud_credentials/main.tf
+++ b/tests/integration/024_cloud_credentials/main.tf
@@ -24,14 +24,19 @@ data "env0_cloud_credentials" "all_aws_credentials" {
   credential_type = "AWS_ASSUMED_ROLE_FOR_DEPLOYMENT"
 }
 
-data "env0_aws_credentials" "aws_credentials1" {
-  name     = data.env0_cloud_credentials.all_aws_credentials.names[index(data.env0_cloud_credentials.all_aws_credentials.names, env0_aws_credentials.aws_cred1.name)]
-}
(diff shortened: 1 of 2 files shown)


tests/integration/024_cloud_credentials/main.tf
```diff
@@ -19,9 +24,12 @@ data "env0_cloud_credentials" "aws_credentials" {
...
 
-data "env0_aws_credentials" "aws_credentials" {
-  for_each = toset(data.env0_cloud_credentials.aws_credentials.names)
-  name     = each.value
+data "env0_aws_credentials" "aws_credentials1" {
+  name     = data.env0_cloud_credentials.aws_credentials.names[index(data.env0_cloud_credentials.aws_credentials.names, env0_aws_credentials.aws_cred1.name)]
+}
+
+data "env0_aws_credentials" "aws_credentials2" {
+  name     = data.env0_cloud_credentials.aws_credentials.names[index(data.env0_cloud_credentials.aws_credentials.names, env0_aws_credentials.aws_cred2.name)]
```
liranfarage89, line 32: First time seeing this format. does it work like `for_each`?

GiliFaroEnv0: nope its the opposite ,
you take a list and search in the list for a specific value

GiliFaroEnv0: @liranfarage89 same as https://github.com/env0/terraform-provider-env0/pull/697

liranfarage89 (approved):
LGTM

tests/integration/024_cloud_credentials/main.tf
```diff
@@ -4,26 +4,35 @@ resource "random_string" "random" {
...
 
-data "env0_aws_credentials" "aws_credentials" {
-  for_each = toset(data.env0_cloud_credentials.aws_credentials.names)
-  name     = each.value
+data "env0_aws_credentials" "aws_credentials1" {
+  name     = data.env0_cloud_credentials.all_aws_credentials.names[index(data.env0_cloud_credentials.all_aws_credentials.names, env0_aws_credentials.aws_cred1.name)]
+}
+
+data "env0_aws_credentials" "aws_credentials2" {
+  name     = data.env0_cloud_credentials.all_aws_credentials.names[index(data.env0_cloud_credentials.all_aws_credentials.names, env0_aws_credentials.aws_cred2.name)]
```
sagilaufer1992, line 33: why was this part added? it's not used

GiliFaroEnv0: this verify you found the resource you created before,
otherwise it will fail

sagilaufer1992: `aws_credentials2` is not used. in the output we only output `1`, not even `GCP`, but I get why to create `GCP` too. what's the use of `aws2`?

sagilaufer1992: ok, got it, thanks

PR #727, opened 2023-10-14
Feat: add support to Drift Detection on env0_project_policy

### LINKED ISSUE #726: Add support to Drift Detection on env0_project_policy
Drift Detection is available in Project Policies through UI but not available on resource env0_project_policy. 

Add drift_detection_cron as option on env0_project_policy resource.

Proposed solution:
```hcl
resource "env0_project_policy" "project_policy" {
  project_id                    = env0_project.example.id
  ...
  drift_detection_cron          = "0 4 * * *"
}
```

**Additional context**
The option available on portal:
image

TomerHeber:
resolves #726 

### Solution
1. Added cron to the schema.
2. Minor name changes to align with the naming convention.
3. Added acceptance tests.
4. Added integration tests.


env0/provider.go
```diff
@@ -94,7 +94,7 @@ func Provider(version string) plugin.ProviderFunc {
 			},
 			ResourcesMap: map[string]*schema.Resource{
 				"env0_project":                              resourceProject(),
-				"env0_project_policy":                       resourcePolicy(),
+				"env0_project_policy":                       resourceProjectPolicy(),
```
TomerHeber, line 97: naming change to align with convention.

env0/resource_project_policy.go
```diff
@@ -10,14 +10,14 @@ import (
...
 )
 
-func resourcePolicy() *schema.Resource {
+func resourceProjectPolicy() *schema.Resource {
 	return &schema.Resource{
-		CreateContext: resourcePolicyCreate,
-		ReadContext:   resourcePolicyRead,
-		UpdateContext: resourcePolicyUpdate,
-		DeleteContext: resourcePolicyDelete,
+		CreateContext: resourceProjectPolicyCreate,
```
TomerHeber, line 15: naming changes to align with convention.

env0/resource_project_policy.go
```diff
@@ -165,14 +171,18 @@ func resourcePolicyUpdate(ctx context.Context, d *schema.ResourceData, meta inte
 		payload.MaxTtl = ""
 	}
 
+	if payload.DriftDetectionCron != "" {
+		payload.DriftDetectionEnabled = true
```
TomerHeber, line 175: If drift is set by the schema pass "enabled" = "true" in the payload.

env0/resource_project_policy.go
```diff
@@ -165,14 +171,18 @@ func resourcePolicyUpdate(ctx context.Context, d *schema.ResourceData, meta inte
...
+
 	if _, err := apiClient.PolicyUpdate(payload); err != nil {
 		return diag.Errorf("could not update policy: %v", err)
 	}
 
 	return nil
 }
 
-func resourcePolicyDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+func resourceProjectPolicyDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
yaronya, line 185: Shouldn't we disable the cron in case the project policy is deleted?

env0/resource_project_policy.go
yaronya: Does it support the disable of this config upon attribute removal?

TomerHeber: golang has default values. Therefore, if the field is removed 'false' will be sent.

env0/resource_project_policy.go
TomerHeber: I don't know... wouldn't it happen anyway if the policy is removed? How would it be able to run a cron for a policy that doesn't exist?

avnerenv0: what happens when a policy is removed?
I don't think we have an endpoint to "delete a policy"

yaronya: We don't
The policy should be reset using the same API (that's what the UI do, I think - it sends an empty value)

avnerenv0: @yaronya that's a bad idea it means this client will need to know what are the default values of each key.

Why don't we add an endpoint to reset the policy to default?
The backend has `getDefaultPolicies` it would simple and maintainable for the future.

avnerenv0: So @TomerHeber we'll take that, but I think in the mean time can we make it so the destroy sends a 'cron disabled' property, and handle that in a different ticket, sounds right?

TomerHeber: @avnerenv0 - yes. I'll make the required changes. Thanks!

TomerHeber: @avnerenv0 I have validated. No code changes are required.
On a delete, the update payload will be sent with `driftDetectionCron=""` and `driftDetectionEnabled=false`

```
	NumberOfEnvironments        int    `json:"numberOfEnvironments,omitempty"`
	NumberOfEnvironmentsTotal   int    `json:"numberOfEnvironmentsTotal,omitempty"`
	RequiresApprovalDefault     bool   `json:"requiresApprovalDefault"`
	IncludeCostEstimation       bool   `json:"includeCostEstimation"`
	SkipApplyWhenPlanIsEmpty    bool   `json:"skipApplyWhenPlanIsEmpty"`
	DisableDestroyEnvironments  bool   `json:"disableDestroyEnvironments"`
	MaxTtl                      string `json:"maxTtl,omitempty"`
	DefaultTtl                  string `json:"defaultTtl,omitempty"`
	ForceRemoteBackend          bool   `json:"forceRemoteBackend"`
	DriftDetectionCron          string `json:"driftDetectionCron"`
	DriftDetectionEnabled       bool   `json:"driftDetectionEnabled"`
}
```

avnerenv0 (approved):

PR #735, opened 2023-10-25
Fix: integration test 025 is not isolated

sagilaufer1992:

test `025_notifications` uses `all_notifications`, and it fetches data from other tests, such as `017_notification`, that may become unavailable during test 025's run.

### Solution

make the test rely solely on its own data

### QA done

ran the test locally to make sure it works on its own. deliberately failing it is much harder, didn't happen sporadically (that we caught) until a few days prior to the PR

### DIFF
--- a/tests/integration/025_notifications/main.tf
+++ b/tests/integration/025_notifications/main.tf
@@ -6,14 +6,18 @@ resource "random_string" "random" {
   min_lower = 20
 }
 
+locals {
+  notification_name_prefix = "integration-test-025-notification"
+}
+
 resource "env0_notification" "test_notification_1" {
-  name  = "notification123-${random_string.random.result}-1"
+  name  = "${local.notification_name_prefix}-1-${random_string.random.result}"
   type  = "Slack"
   value = "https://someurl1.com"
 }
 
 resource "env0_notification" "test_notification_2" {
-  name  = "notification123-${random_string.random.result}-2"
+  name  = "${local.notification_name_prefix}-2-${random_string.random.result}"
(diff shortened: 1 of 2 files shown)


tests/integration/025_notifications/main.tf
```diff
@@ -18,9 +18,14 @@ resource "env0_notification" "test_notification_2" {
   value = "https://someurl2.com"
 }
 
-data "env0_notifications" "all_notifications" {}
+data "env0_notifications" "all_notifications" {
+  depends_on = [env0_notification.test_notification_1, env0_notification.test_notification_2]
+}
+
+data "env0_notification" "test_notification_1" {
```
avnerenv0, line 25: why not just use 
```
data "env0_notification" "my_notification" {
  name = locals.test_notification_1_name
}
```
here?

sagilaufer1992: we wish to see that the `all_notifications` works as expected too, so I suggest we utilize its data, so if it got broken for some reason the test will fail

chpl: agree with @avnerenv0 .
can be replaced with post condition on the `all_notifications`

sagilaufer1992: [0c08f76](https://github.com/env0/terraform-provider-env0/pull/735/commits/0c08f76aa77d2f10d0a9144aa5caf8222804f19b)

used the resource name, and used output for the `all_notifications` verification, that it got one of the notifications

TomerHeber (approved):

PR #739, opened 2023-11-01
Feat: clarify provider documentation

### LINKED ISSUE #738: clarify provider documentation
add optional? --  (Optional String, Sensitive)   - and "Instead of setting api_key, this can also be set via the ENV0_API_KEY"

TomerHeber:
resolves #738 

### Solution

1. API key and API secret are now optional in the schema.
2. Will still return an error if they're not configured via schema or environment variables.
3. Minor updates to provider documentation to make it clearer.

### DIFF
--- a/env0/provider.go
+++ b/env0/provider.go
@@ -38,16 +38,16 @@ func Provider(version string) plugin.ProviderFunc {
 				},
 				"api_key": {
 					Type:        schema.TypeString,
-					Description: "env0 API key. This can also be set via the ENV0_API_KEY environment variable.",
+					Description: "env0 API key. Instead of this field, it can also be set via the ENV0_API_KEY environment variable.",
 					DefaultFunc: schema.EnvDefaultFunc(apiKeyEnv, nil),
-					Required:    true,
+					Optional:    true,
 					Sensitive:   true,
 				},
 				"api_secret": {
 					Type:        schema.TypeString,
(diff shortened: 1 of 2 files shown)


env0/provider.go
```diff
@@ -195,9 +195,19 @@ func configureProvider(version string, p *schema.Provider) schema.ConfigureConte
 				return false
 			})
 
+		apiKey, ok := d.GetOk("api_key")
```
TomerHeber, line 198: Verify that these fields are set. If not... it's an error.

env0/provider_test.go
```diff
@@ -74,12 +74,12 @@ func testMissingEnvVar(t *testing.T, envVars map[string]string, expectedKey stri
 		defer os.Setenv(key, "")
 	}
 
-	diags := Provider("TEST")().Validate(&terraform.ResourceConfig{})
+	diags := Provider("TEST")().Configure(context.Background(), &terraform.ResourceConfig{})
```
TomerHeber, line 77: Configure is at later stage that returns the error. (Because now schema passes).

env0/provider.go
yaronya: Took me a little time to understand that `EnvDefaultFunc` sets the value from env var 😅

yaronya (approved):

PR #742, opened 2023-11-04
Feat: env0_environment with project_id property

### LINKED ISSUE #737: env0_environment with project_id property
Hi, when calling the env0 terraform provider data source env0_environment with the environment name property that exists more than once (in different projects) we receive the following error:
│ Error: Found multiple environments for name: $name Use ID instead or make sure environment names are unique 
please provide a project_id property for this data source so the data source will be unique

TomerHeber:
resolves #737 

### Solution

1. Updated the schema to make project_id optional and not just computed.
2. Added an acceptance test for the new use-case. (Failed before adding code, passed updating code).
3. Updated code to filter out if project_id is set.

### DIFF
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -1058,7 +1058,7 @@ func getConfigurationVariableFromSchema(variable map[string]interface{}) client.
 	return configurationVariable
 }
 
-func getEnvironmentByName(name interface{}, meta interface{}, excludeArchived bool) (client.Environment, diag.Diagnostics) {
+func getEnvironmentByName(meta interface{}, name string, projectId string, excludeArchived bool) (client.Environment, diag.Diagnostics) {
 	apiClient := meta.(client.ApiClientInterface)
 	environments, err := apiClient.Environments()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


env0/resource_environment.go
```diff
@@ -1058,7 +1058,7 @@ func getConfigurationVariableFromSchema(variable map[string]interface{}) client.
 	return configurationVariable
 }
 
-func getEnvironmentByName(name interface{}, meta interface{}, excludeArchived bool) (client.Environment, diag.Diagnostics) {
+func getEnvironmentByName(meta interface{}, name string, projectId string, excludeArchived bool) (client.Environment, diag.Diagnostics) {
```
TomerHeber, line 1061: added a new field and changed the order of the variables.
In case more variables are added, I'll convert it to a "struct filter".

lironshatbai (approved):

env0/resource_environment.go
lironshatbai: d

razbensimon (approved):

PR #753, opened 2023-11-21
Chore: filter out more redundant scopes from ConfigurationVariablesBy…

### LINKED ISSUE #751: Filter out more redundant scopes from `ConfigurationVariablesByScope`
[This bit of code](https://github.com/env0/terraform-provider-env0/blob/main/client/configuration_variable.go#L124-L130) supposedly filters out variables from scoped which weren't requested (as the request made returns "merged" variables from cascading scopes).

But, the code only filters out organization and template scope variables explicitly.

Filter out all variables from unrequested scopes (using the `scopeId` field).

You could explicitly filter out all the project variables received which were not from the specifically requested project, but at that point i think the code is too messy and specific for no reason.

**Additional context**
This is requested alongside an effort to make variables from parent projects cascade throughout all of their sub-projects.
Since we want the provider to keep its behaviour and not get "new" values from ancestor projects' variables, we need to filter those out as well

TomerHeber:
…Scope

closes #751 

### Solution

1. Updated the filter.
2. Updated the unit tests.

### DIFF
--- a/client/configuration_variable.go
+++ b/client/configuration_variable.go
@@ -85,16 +85,6 @@ func (client *ApiClient) ConfigurationVariablesById(id string) (ConfigurationVar
 	return result, nil
 }
 
-func filterOutConfigurationVariables(variables []ConfigurationVariable, scope Scope) []ConfigurationVariable {
-	filteredVariables := []ConfigurationVariable{}
-	for _, variable := range variables {
-		if variable.Scope != scope {
-			filteredVariables = append(filteredVariables, variable)
-		}
-	}
-	return filteredVariables
-}
-
 func (client *ApiClient) ConfigurationVariablesByScope(scope Scope, scopeId string) ([]ConfigurationVariable, error) {
 	organizationId, err := client.OrganizationId()
 	if err != nil {
@@ -120,15 +110,15 @@ func (client *ApiClient) ConfigurationVariablesByScope(scope Scope, scopeId stri
 		return []ConfigurationVariable{}, err
 	}
 
-	// The API returns global and template scopes for environment (and other) scopes. Filter them out.
-	if scope != ScopeGlobal {
-		result = filterOutConfigurationVariables(result, ScopeGlobal)
-		if scope != ScopeTemplate {
-			result = filterOutConfigurationVariables(result, ScopeTemplate)
+	// The API returns variables of upper scopes. Filter them out.
+	var filteredVariables []ConfigurationVariable
+	for _, variable := range result {
+		if scope == variable.Scope {
+			filteredVariables = append(filteredVariables, variable)
 		}
 	}
 
-	return result, nil
+	return filteredVariables, nil
 }
 
 func (client *ApiClient) ConfigurationVariableCreate(params ConfigurationVariableCreateParams) (ConfigurationVariable, error) {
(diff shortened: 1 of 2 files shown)


client/configuration_variable.go
```diff
@@ -120,15 +110,15 @@ func (client *ApiClient) ConfigurationVariablesByScope(scope Scope, scopeId stri
...
-	if scope != ScopeGlobal {
-		result = filterOutConfigurationVariables(result, ScopeGlobal)
-		if scope != ScopeTemplate {
-			result = filterOutConfigurationVariables(result, ScopeTemplate)
+	// The API returns variables of upper scopes. Filter them out.
+	var filteredVariables []ConfigurationVariable
+	for _, variable := range result {
+		if scope == variable.Scope {
+			filteredVariables = append(filteredVariables, variable)
 		}
```
Wassap124, line 118: that's not good enough 😅 
as per the explanation in the issue https://github.com/env0/terraform-provider-env0/issues/751

There'd be several variables from the same scope "type" from a different specific scope.
you'd get more variables from the API in the `PROJECT` scope, but which are NOT defined on the requested `projectId`.

Does that make sense?

Wassap124: Each variable has a `scope` and a `scopeId`.
The scope can be `ENVIRONMENT`, `PROJECT`, etc.
The scopeIds are uuids fir the specific environment, project, etc. respectfully.

Wassap124 (approved):

client/configuration_variable.go
Wassap124: a

Wassap124 (changes requested):

Wassap124 (changes requested):
Approved by mistake 🤦

weinguy-env0:
[Show Pretty-Plan even if there is a failure in the deployment after the Plan](https://trello.com/c/fTsikGJn/11359-show-pretty-plan-even-if-there-is-a-failure-in-the-deployment-after-the-plan)

client/configuration_variable.go
TomerHeber: so.. should I filter instead based on scope ID? (instead of scope).

Wassap124: YES

TomerHeber: updated.

client/configuration_variable.go
```diff
@@ -120,15 +110,15 @@ func (client *ApiClient) ConfigurationVariablesByScope(scope Scope, scopeId stri
...
 
-	// The API returns global and template scopes for environment (and other) scopes. Filter them out.
-	if scope != ScopeGlobal {
-		result = filterOutConfigurationVariables(result, ScopeGlobal)
-		if scope != ScopeTemplate {
-			result = filterOutConfigurationVariables(result, ScopeTemplate)
+	// The API returns variables of upper scopes. Filter them out.
+	var filteredVariables []ConfigurationVariable
+	for _, variable := range result {
+		if scopeId == variable.ScopeId {
```
RLRabinowitz, line 116: I would also add `scope == variable.Scope` here for safety

TomerHeber: @Wassap124 - do you concur?

Wassap124:
btw @TomerHeber we hadn't forgotten about this PR, we've started QAing this

ronnyorot (approved):
I QAed it and it looks good 🙃

PR #783, opened 2024-01-14
Feat: add new OIDC credentials creation and assigment (vault)

### LINKED ISSUE #743: Add new OIDC credentials creation and assigment
As part of the built-in OIDC authentication epic we added 4 new deployment credentials types that we will need to support in the provider. The new credentials are:
* AWS OIDC:
  - `name` - (String) the name for the credential
  - `arn` - (String) the aws role arn
  - `duration` - (Number, optional) the session duration in seconds. If set must be one of the following: 3600 (1h), 7200 (2h), 14400 (4h), 18000 (5h default), 28800 (8h), 43200 (12h)
* GCP OIDC:
  - `name` - (String) the name for the credential
  - `jwt_configuration_file_content` - (String) should contain the JSON content of the JWT configuration file
* Azure OIDC:
  - `name` - (String) the name for the credential
  - `subscription_id` - (String) the azure subscription id
  - `client_id` - (String) the azure client id
  - `tenant_id` - (String) the azure tenant id
* Vault OIDC:
  - `name` - (String) the name for the credential
  - `vault_addr` - (String) the vault instance address
  - `vault_version` - (String) the vault version to download
  - `vault_role` - (String) the vault role
  -  `jwt_auth_backend_path` - (String) the JWT authentication method path
  - `vault_namespace` - (String, optional) the vault namespace

Currently, when adding an AWS credential, in env0 we have the `externalId` field. For the new credentials we won't need that we will need `JWT OIDC sub`. For example the input form for GCP OIDC credential looks like this:

image

You can get that value from our API `<base-url>/api-keys/oidc-sub?organizationId=<org-id>`

We will also need to be able to assign all those new credentials to projects

TomerHeber:
resolves #743 

### Solution

Last PR for the OIDC effort. This one covers Vault OIDC.

1. Added new resource and data source.
2. Added acceptance tests for resource and data source.
3. Added a new integration test for vault credenitals.
4. Added examples.
5. Added some constants in the client layer.


tests/integration/031_vault_credentials/expected_outputs.json
```diff
@@ -0,0 +1,2 @@
+{
```
Yossi-kerner, line 1: why do we need that?

tests/integration/031_vault_credentials/conf.tf
```diff
@@ -0,0 +1,15 @@
...
+  required_providers {
+    env0 = {
+      source = "terraform-registry.env0.com/env0/env0"
+    }
+  }
+}
+
+provider "env0" {}
+
+variable "second_run" {
```
Yossi-kerner, line 13: why do we need this variable?

Yossi-kerner (approved):
Looking good left some small comments

tests/integration/031_vault_credentials/conf.tf
TomerHeber: I actually want to use it to test updates.
But hadn't applied it to the test. Adding it now.

tests/integration/031_vault_credentials/expected_outputs.json
TomerHeber: This is required for all tests even if not used. Otherwise test fail.
It's used if you want to test with terraform's "output".

tests/integration/031_vault_credentials/conf.tf
TomerHeber: See this is an example:
https://github.com/env0/terraform-provider-env0/pull/783/commits/eef71ed878fdafeb20fd7f6e1fbc945233007470

Yossi-kerner: it looks like we have some failing tests

Yossi-kerner: once you fix those you are good to go

TomerHeber: yeah I just found out update was never tested.
I'm adding the tests, and fixing the cause for it failing.
I'll push once the tests path. Thanks!

TomerHeber: never tested not just for OIDC... but all AWS, GCP and Azure for all credential types...

Yossi-kerner: Can you elaborate on how was it missed? So I will be able to catch it faster in the future

TomerHeber: These are old tests, written before I started working on this project.
I noticed that` var.second_run` wasn't used in any of the tests. If it's not used, "update" isn't tested.
For "update" to work properly I had to add "omitempty" to some of the structures (this is why vault failed to pass).

In other words, there were bugs in the provider that were not detected because tests were missing.

I added the tests and fixed the bugs. Now it should all work as expected.

Yossi-kerner: Alright thanks 😊

PR #785, opened 2024-01-21
Feat: add support for running terragrunt with opentofu

### LINKED ISSUE #784: Add support for running terragrunt with opentofu
We recently added an option to run `terragrunt` with `opentofu` in env0. When a user creates a `terragrunt` template he can choose wether he wants to run `terragrunt` with `opentofu` or `terraform`.
We did it by adding a `terragruntTfBinary` field to the template that defaults to `opentofu`

We would like to have that `terragruntTfBinary` property when creating  a terragrunt template from the provider. The two possible values for this field are `terraform` and `opentofu` and we want `opentofu` to be the default 🔴 **_only for new templates_** 🔴 

**Additional context**
The create template API [docs](https://docs.env0.com/reference/templates-create-blueprint)

TomerHeber:
resolves #784 

### Solution

1. Updated the schema.
2. Added special handling to support the unique behavior.
3. Updated acceptance tests.

### DIFF
--- a/client/template.go
+++ b/client/template.go
@@ -37,7 +37,7 @@ type Template struct {
 	Name                 string           `json:"name"`
 	Description          string           `json:"description"`
 	OrganizationId       string           `json:"organizationId"`
-	Path                 string           `json:"path,omitempty" tfschema:",omitempty"`
+	Path                 string           `json:"path" tfschema:",omitempty"`
 	Revision             string           `json:"revision"`
 	ProjectId            string           `json:"projectId"`
 	ProjectIds           []string         `json:"projectIds"`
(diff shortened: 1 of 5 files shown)


client/template.go
```diff
@@ -128,6 +130,10 @@ func (payload *TemplateCreatePayload) Invalidate() error {
 		return errors.New("must supply opentofu version")
 	}
 
+	if payload.TerragruntTfBinary != "" && payload.Type != "terragrunt" {
```
TomerHeber, line 133: This is to avoid setting `TerragruntTfBinary` without a template of type `terragrunt`.

env0/resource_template.go
```diff
@@ -239,6 +239,12 @@ func getTemplateSchema(prefix string) map[string]*schema.Schema {
 			ConflictsWith: allVCSAttributesBut("helm_chart_name", "is_helm_repository"),
 			RequiredWith:  requiredWith("helm_chart_name"),
 		},
+		"terragrunt_tf_binary": {
```
TomerHeber, line 242: Could not set a deault value to "OpenTofu".
Default will only make sense if this was a "Terragrunt" template.
This is also an issue with existing templates that will use "Terrafrom" and not "OpenTofu".

I tried a few options. At the end I had to manually handle this unique use-case.
This makes the PR a little more complex than usual.

env0/resource_template.go
```diff
@@ -373,12 +379,30 @@ func templateCreatePayloadFromParameters(prefix string, d *schema.ResourceData)
...
+
+	templateType := d.Get(templateTypeKey).(string)
+
+	// If the user has set a value - use it.
+	if terragruntTfBinary := d.Get(terragruntTfBinaryKey).(string); terragruntTfBinary != "" {
+		payload.TerragruntTfBinary = terragruntTfBinary
+	} else {
+		// No value was set - if it's a new template resource of type 'terragrunt' - default to 'opentofu'
+		if templateType == "terragrunt" && isNew {
+			payload.TerragruntTfBinary = "opentofu"
```
TomerHeber, line 404: send opentofu in the creation request if it's a new terragrunt template.

env0/resource_template.go
```diff
@@ -373,12 +379,30 @@ func templateCreatePayloadFromParameters(prefix string, d *schema.ResourceData)
...
 		isAzureDevOpsKey = prefix + "." + isAzureDevOpsKey
+		terragruntTfBinaryKey = prefix + "." + terragruntTfBinaryKey
+		templateTypeKey = prefix + "." + templateTypeKey
+	}
+
+	templateType := d.Get(templateTypeKey).(string)
+
+	// If the user has set a value - use it.
+	if terragruntTfBinary := d.Get(terragruntTfBinaryKey).(string); terragruntTfBinary != "" {
+		payload.TerragruntTfBinary = terragruntTfBinary
```
TomerHeber, line 400: If the user manually configured terragrunt_tf_binary use it.

env0/resource_template.go
```diff
@@ -401,11 +425,22 @@ func templateCreatePayloadFromParameters(prefix string, d *schema.ResourceData)
...
 	if prefix != "" {
-		pathPrefix = prefix + ".0.path"
+		pathPrefix = prefix + ".0." + pathPrefix
+		terragruntTfBinaryPrefix = prefix + ".0." + terragruntTfBinaryPrefix
 	}
 
 	path, pathOk := d.GetOk(pathPrefix)
+	terragruntTfBinary := d.Get(terragruntTfBinaryPrefix).(string)
+
+	// If this value isn't set, ignore whatever is returned from the response.
```
TomerHeber, line 438: Added a long explanation here. Hopefully it's clear enough.
Basically, if the user hasn't set anything in the schema ignore the response. This is to avoid drifts.

env0/resource_environment_test.go
```diff
@@ -1899,9 +1899,11 @@ func TestUnitEnvironmentWithoutTemplateResource(t *testing.T) {
 				ErrorRegex: "RetryMeForDestroy.*",
 			},
 		},
-		Type:                 "terraform",
+		Type:                 "terragrunt",
```
TomerHeber, line 1902: This impacts an environment with a template.
Wanted to test that use-case as well. In this case the user explicitly mentioned here wants to use "terraform" instead of "opentofu".

env0/resource_template_test.go
```diff
@@ -35,9 +35,11 @@ func TestUnitTemplateResource(t *testing.T) {
 				ErrorRegex: "RetryMeForDestroy.*",
 			},
 		},
-		Type:               "terraform",
+		Type:               "terragrunt",
 		IsGitlabEnterprise: true,
 		TerraformVersion:   "0.12.24",
+		TerragruntVersion:  "0.35.1",
+		TerragruntTfBinary: "opentofu",
```
TomerHeber, line 42: In this case, user has set nothing. Defaults to opentofu when the template is created.

kvendingoldo:
btw. you can also integrate [tenv](https://github.com/tofuutils/tenv) that support Terraform as well as OpenTofu (and Terragrunt :) ) in one tool. It allow you to simplify version management.

PR #789, opened 2024-02-05
Feat: module testing support

### LINKED ISSUE #788: FR: Module testing support
Env0 has a new feature - module testing, provider should support it.

[module resource](https://registry.terraform.io/providers/env0/env0/latest/docs/resources/module) should have a few new inputs
* `moduleTestEnabled` - `boolean`
* `runTestsOnPullRequest` - boolean, only allowed if `moduleTestEnabled=true` 
* opentofuVersion - string (same as in [template](https://registry.terraform.io/providers/env0/env0/latest/docs/resources/template#opentofu_version)) , only allowed if `moduleTestEnabled=true`

**Additional context**
see https://docs.env0.com/reference/templates-create-module

TomerHeber:
resolves #788 

### Solution

1. Added the new fields to the API structs.
2. Added the new fields to the schema.
3. Added a condition to fail module test is disabled but opentofu or run_tests_on pull are enalbed.
4. Updated integration tests.
5. Updated acceptance tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - name: Install Go
-        uses: actions/setup-go@v4
+        uses: actions/setup-go@v5
         with:
           go-version: ${{ env.GO_VERSION }}
       - name: Checkout code
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Unshallow
         run: git fetch --prune --unshallow
       - name: Set up Go
-        uses: actions/setup-go@v4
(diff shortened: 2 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - name: Install Go
-        uses: actions/setup-go@v4
+        uses: actions/setup-go@v5
```
chpl, line 21: @TomerHeber WDYT about using [dependabot to update those](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot)

TomerHeber: I'll check it out.
Thanks for the advice.

PR #793, opened 2024-02-14
Fix: changing terragrunt_working_directory for an environment should …

### LINKED ISSUE #792: Changing `terragrunt_working_directory` for an environment should force replacement instead of update in place
**Describe the bug**
When you change the `terragrunt_working_directory` for an existing terragrunt `env0_environment`, it attempts to update the environment in-place. However, the `EnvironmentUpdate` API does not support changing the `terragrunt_working_directory`. Once this value is set for an environment it cannot be updated

So, if anyone attempts to change the `terragrunt_working_directory`, the TF provider will make an attempt to update and will mark it as a success, making the `terraform apply` finish successfully. However, the `terragrunt_working_directory` will not actually be changed

**To Reproduce**
1. Create a terragrunt template, and an environment using said template with a `terragrunt_working_directory` being explicitly set
2. Apply
3. Change the `terragrunt_working_directory` and re-apply
4. The action will succeed, but the `terragrunt_working_directory` will not actually be changed in env0. Also, when the resource is refreshed drift will be detected for the field

**Expected behavior**
Changing the `terragrunt_working_directory` of an existing environment should force replacement, destroying the existing environment and creating a new environment

**Provider Version**
1.16.11

**Additional context**
`terragrunt_working_directory` cannot be changed for an existing environment

TomerHeber:
…force replacement instead of update in place

fixes #792 

### Solution

1. Added forcenew to the terragrunt_working_directory field.
2. Updated tests (avoid changing terragrunt directory).

### DIFF
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -251,7 +251,8 @@ func resourceEnvironment() *schema.Resource {
 			},
 			"terragrunt_working_directory": {
 				Type:        schema.TypeString,
-				Description: "The working directory path to be used by a Terragrunt template. If left empty '/' is used.",
+				Description: "The working directory path to be used by a Terragrunt template. If left empty '/' is used. Note: modifying this field destroys the current environment and creates a new one",
+				ForceNew:    true,
 				Optional:    true,
 			},
 			"vcs_commands_alias": {
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -658,7 +658,7 @@ func shouldDeploy(d *schema.ResourceData) bool {
 }
 
 func shouldUpdate(d *schema.ResourceData) bool {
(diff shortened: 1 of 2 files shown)


env0/resource_environment.go
```diff
@@ -251,7 +251,8 @@ func resourceEnvironment() *schema.Resource {
 			},
 			"terragrunt_working_directory": {
 				Type:        schema.TypeString,
-				Description: "The working directory path to be used by a Terragrunt template. If left empty '/' is used.",
+				Description: "The working directory path to be used by a Terragrunt template. If left empty '/' is used. Note: modifying this field destroys the current environment and creates a new one",
+				ForceNew:    true,
```
TomerHeber, line 255: This will force terraform to create a new environment (if it changes).

RLRabinowitz (commented):
We should probably also remove the field from `shouldUpdate`

TomerHeber:
> We should probably also remove the field from `shouldUpdate`

It wouldn't matter from a correctness perspective. But for completeness I have removed it.

RLRabinowitz (approved):

PR #804, opened 2024-03-12
Feat: add `prevent_auto_deploy` flag to create without deploy

tomer-landesman:
we want to support the new `prevent_auto_deploy` argument of the create envirionment API.

using the property, the user can skip the deployment of the environment in env0 and only create it on resource create. updating and destroying should remain the same

### Solution

- add support for the new `prevent_auto_deploy` api

### Manual QA against dev
image

### DIFF
--- a/client/environment.go
+++ b/client/environment.go
@@ -138,7 +138,7 @@ type EnvironmentCreate struct {
 	IsRemoteBackend             *bool                  `json:"isRemoteBackend,omitempty" tfschema:"-"`
 	Type                        string                 `json:"type,omitempty"`
 	DriftDetectionRequest       *DriftDetectionRequest `json:"driftDetectionRequest,omitempty" tfschema:"-"`
-	PreventAutoDeploy           bool                   `json:"preventAutoDeploy"`
+	PreventAutoDeploy           *bool                  `json:"preventAutoDeploy,omitempty" tfschema:"-"`
 }
 
 // When converted to JSON needs to be flattened. See custom MarshalJSON below.
--- a/client/environment.go
+++ b/client/environment.go
@@ -138,7 +138,7 @@ type EnvironmentCreate struct {
 	IsRemoteBackend             *bool                  `json:"isRemoteBackend,omitempty" tfschema:"-"`
 	Type                        string                 `json:"type,omitempty"`
 	DriftDetectionRequest       *DriftDetectionRequest `json:"driftDetectionRequest,omitempty" tfschema:"-"`
-	PreventAutoDeploy           *bool                  `json:"preventAutoDeploy,omitempty" tfschema:"-"`
+	PreventAutoDeploy           bool                   `json:"preventAutoDeploy"`
 }
 
 // When converted to JSON needs to be flattened. See custom MarshalJSON below.
--- a/client/environment.go
+++ b/client/environment.go
@@ -138,6 +138,7 @@ type EnvironmentCreate struct {
 	IsRemoteBackend             *bool                  `json:"isRemoteBackend,omitempty" tfschema:"-"`
 	Type                        string                 `json:"type,omitempty"`
 	DriftDetectionRequest       *DriftDetectionRequest `json:"driftDetectionRequest,omitempty" tfschema:"-"`
+	PreventAutoDeploy           *bool                  `json:"preventAutoDeploy,omitempty" tfschema:"-"`
 }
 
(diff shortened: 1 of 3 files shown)


client/environment.go
```diff
@@ -138,6 +138,7 @@ type EnvironmentCreate struct {
 	IsRemoteBackend             *bool                  `json:"isRemoteBackend,omitempty" tfschema:"-"`
 	Type                        string                 `json:"type,omitempty"`
 	DriftDetectionRequest       *DriftDetectionRequest `json:"driftDetectionRequest,omitempty" tfschema:"-"`
+	PreventAutoDeploy           *bool                  `json:"preventAutoDeploy,omitempty" tfschema:"-"`
```
TomerHeber, line 141: why is it a pointer, is it okay to always pass false?

env0/resource_environment.go
```diff
@@ -249,6 +249,11 @@ func resourceEnvironment() *schema.Resource {
 				Description: "should use remote backend",
 				Optional:    true,
 			},
+			"prevent_auto_deploy": {
+				Type:        schema.TypeBool,
+				Description: "use this flag to prevent auto deploy on environment creation",
+				Optional:    true,
```
TomerHeber, line 255: if it's okay to always pass false, you can add a default: false.

client/environment.go
tomer-landesman: if it won't be a pointer and the user won't specify it - will it evaluate to false? I think we could always pass it but is it really better?

env0/resource_environment_test.go
```diff
@@ -185,6 +185,72 @@ func TestUnitEnvironmentResource(t *testing.T) {
...
+						Check: resource.ComposeAggregateTestCheckFunc(
+							resource.TestCheckResourceAttr(accessor, "id", environment.Id),
+							resource.TestCheckResourceAttr(accessor, "name", environment.Name),
+							resource.TestCheckResourceAttr(accessor, "project_id", environment.ProjectId),
+							resource.TestCheckResourceAttr(accessor, "template_id", templateId),
+							resource.TestCheckResourceAttr(accessor, "prevent_auto_deploy", "true"),
+						),
+					},
+					{
+						Config: resourceConfigCreate(resourceType, resourceName, map[string]interface{}{
```
TomerHeber, line 221: this step isn't required here...

tomer-landesman: it fails without it - care to explain what it does?

client/environment.go
TomerHeber: it simplifies the code significantly.

But it depends on the API. Some of the API's params do have "3" values: "pass nothing", "false", "true".

env0/resource_environment_test.go
tomer-landesman: sorry my bad, had to fix the assertions as well - the API of those tests is confusing AF 😅 
[removed unnecessary step](https://github.com/env0/terraform-provider-env0/pull/804/commits/e9a6644e1f1d7019b8a055b7e2ff31c40fe99f22)

env0/resource_environment.go
```diff
@@ -799,6 +804,11 @@ func getCreatePayload(d *schema.ResourceData, apiClient client.ApiClientInterfac
 		payload.PullRequestPlanDeployments = boolPtr(val.(bool))
 	}
 
+	//lint:ignore SA1019 reason: https://github.com/hashicorp/terraform-plugin-sdk/issues/817
+	if val, exists := d.GetOkExists("prevent_auto_deploy"); exists {
```
TomerHeber, line 808: you're missing more use-cases.

Search for the string "auto_deploy_on_path_changes_only"

tomer-landesman: why tho? I'm interested only in `Create` flow - update and delete are not interesting to me (even if this property is changed - nothing should be sone because its not really persisted at the backend)

TomerHeber (approved):

client/environment.go
tomer-landesman: I can try passing false as default. let me see

tomer-landesman: thanks!

tomer-landesman: I changed it anyway cause it should be simpler - can you take a quick look?

TomerHeber: unit test is failing...

tomer-landesman: yea I gave up on the `bool` change, it failing to much tests now 😅

TomerHeber (approved):

PR #812, opened 2024-03-21
Feat: support removal_strategy on destroy environment

### LINKED ISSUE #806: Support `removal_strategy` on destroy environment
**Description**
In contingency to `prevent_auto_deploy`, we would also like to have the other way around - when you destroy and environment you can have the option to force its archive instead of trying to destroy, so you won't see a failure in env0. It's already possible via UI / API.

- on environment resource, add a new flag, `removal_strategy: 'destroy' | 'mark_as_archived' ` 
- when the flag is provided as `mark_as_archived`, (default 'destroy') destroy that environment will result in update the archive property with `PUT /environments/:id` endpoint and won't execute the destroy environment.

TomerHeber:
resolves #806 

### Solution

1. Added 'removal_strategy' to the schema (defaults to: 'destroy') and required functionality.
2. Added a PUT API call that marks the environment as isArchived 'true'.
3. Added unit test for the new API call.
4. Added an acceptance test.
5. Added an integration.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,7 +41,7 @@ jobs:
         uses: goreleaser/goreleaser-action@v5
         with:
           version: latest
-          args: release --rm-dist
+          args: release --clean
         env:
           GPG_FINGERPRINT: ${{ steps.import_gpg.outputs.fingerprint }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -335,6 +335,13 @@ func resourceEnvironment() *schema.Resource {
 				Optional:    true,
 				Default:     false,
 			},
+			"removal_strategy": {
+				Type:             schema.TypeString,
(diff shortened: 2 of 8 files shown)


.github/workflows/release.yml
```diff
@@ -41,7 +41,7 @@ jobs:
         uses: goreleaser/goreleaser-action@v5
         with:
           version: latest
-          args: release --rm-dist
+          args: release --clean
```
TomerHeber, line 44: unrelated.
--rm-dist is deprecated.

env0/resource_environment.go
```diff
@@ -335,6 +335,13 @@ func resourceEnvironment() *schema.Resource {
 				Optional:    true,
 				Default:     false,
 			},
+			"removal_strategy": {
+				Type:             schema.TypeString,
+				Description:      "by default when removing an environment, it gets destroyed. Setting this value to 'mark_as_archived' will force the environment to be archived instead of tying to destroy it ('Mark as inactive' in the UI)",
```
TomerHeber, line 340: I hope the description is clear enough.

tomer-landesman (approved):
Looks good to me.

PR #813, opened 2024-03-25
Feat: Organization Role assignments for Team

### LINKED ISSUE #807: Organization Role assignments for Team
The same as we did [here](https://github.com/env0/terraform-provider-env0/issues/654) and [here](https://github.com/env0/terraform-provider-env0/pull/668) for Environment Level assignments - we want to achieve for the new Org level Team assignments.

Note is is just for TEAMS right now. Users is out of scope.

Use the new [APIs](https://docs.env0.com/reference/roles-get-user-role-assignments): Get, Upsert, Delete.
Same API as before, we just add `organizationId`

TomerHeber:
resolves #807 

### Solution

1. Added  API calls.
2. Added unit tests for API calls.
3. Created a new resource.
4. Added acceptance tests for the new resource.
5. Added integration test for the resource.
6. Added examples.

### DIFF
--- a/client/team_organization_assignment.go
+++ b/client/team_organization_assignment.go
@@ -1,0 +1,59 @@
+package client
+
+type AssignTeamRoleToOrganizationPayload struct {
+	TeamId string `json:"teamId"`
+	Role   string `json:"role" tfschema:"role_id"`
+}
+
+type TeamRoleOrganizationAssignment struct {
+	TeamId string `json:"teamId"`
+	Role   string `json:"role" tfschema:"role_id"`
+	Id     string `json:"id"`
+}
+
+func (client *ApiClient) AssignTeamRoleToOrganization(payload *AssignTeamRoleToOrganizationPayload) (*TeamRoleOrganizationAssignment, error) {
+	var result TeamRoleOrganizationAssignment
+
+	organizationId, err := client.OrganizationId()
+	if err != nil {
+		return nil, err
+	}
+
+	payloadWithOrganizationId := struct {
+		*AssignTeamRoleToOrganizationPayload
+		OrganizationId string `json:"organizationId"`
+	}{
+		payload,
+		organizationId,
+	}
+
+	if err := client.http.Put("/roles/assignments/teams", payloadWithOrganizationId, &result); err != nil {
+		return nil, err
+	}
+
+	return &result, nil
+}
+
+func (client *ApiClient) RemoveTeamRoleFromOrganization(teamId string) error {
+	organizationId, err := client.OrganizationId()
+	if err != nil {
+		return err
+	}
+
+	return client.http.Delete("/roles/assignments/teams", map[string]string{"organizationId": organizationId, "teamId": teamId})
+}
+
+func (client *ApiClient) TeamRoleOrganizationAssignments() ([]TeamRoleOrganizationAssignment, error) {
+	organizationId, err := client.OrganizationId()
+	if err != nil {
+		return nil, err
+	}
+
+	var result []TeamRoleOrganizationAssignment
+
+	if err := client.http.Get("/roles/assignments/teams", map[string]string{"organizationId": organizationId}, &result); err != nil {
+		return nil, err
+	}
+
+	return result, nil
+}
--- a/client/team_organization_assignment.go
+++ b/client/team_organization_assignment.go
@@ -1,18 +1,18 @@
 package client
 
-type AssignTeamRoleToOrganizationPayload struct {
+type AssignOrganizationRoleToTeamPayload struct {
 	TeamId string `json:"teamId"`
 	Role   string `json:"role" tfschema:"role_id"`
 }
 
-type TeamRoleOrganizationAssignment struct {
+type OrganizationRoleTeamAssignment struct {
 	TeamId string `json:"teamId"`
 	Role   string `json:"role" tfschema:"role_id"`
 	Id     string `json:"id"`
 }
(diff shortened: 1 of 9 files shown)


client/team_organization_assignment.go
```diff
@@ -0,0 +1,59 @@
...
+type OrganizationRoleTeamAssignment struct {
+	TeamId string `json:"teamId"`
+	Role   string `json:"role" tfschema:"role_id"`
+	Id     string `json:"id"`
+}
+
+func (client *ApiClient) AssignOrganizationRoleToTeam(payload *AssignOrganizationRoleToTeamPayload) (*OrganizationRoleTeamAssignment, error) {
+	var result OrganizationRoleTeamAssignment
+
+	organizationId, err := client.OrganizationId()
```
sagydr, line 17: you extract `organizationId` from the client? I guess that's for security reasons, but wouldn't it be clearer that the user will pass the `organizationId` and we'll verify it is equal to the client's `organizationId`?

TomerHeber: I don't know. I suggest you discuss with the team.
In general every where else in the provider the organization id is implicit (no need to enter it).

client/team_organization_assignment_test.go
```diff
@@ -0,0 +1,142 @@
...
+				httpCall = mockHttpClient.EXPECT().
+					Put("/roles/assignments/teams", assignPayloadWithOrganizationId, gomock.Any()).
+					Do(func(path string, request interface{}, response *OrganizationRoleTeamAssignment) {
+						*response = *expectedResponse
+					}).Times(1)
+				actualResult, err = apiClient.AssignOrganizationRoleToTeam(assignPayload)
+
+			})
+
+			It("Should send POST request with params", func() {})
```
sagydr, line 52: empty on purpose?

env0/resource_team_organization_assignment.go
```diff
@@ -0,0 +1,112 @@
...
+			}
+
+			d.Set("role_id", assignment.Role)
+
+			return nil
+		}
+	}
+
+	tflog.Warn(ctx, "Drift Detected: Terraform will remove id from state", map[string]interface{}{"id": d.Id()})
+	d.SetId("")
```
sagydr, line 79: sorry can you explain these lines?

examples/resources/env0_team_organization_assignment/resource.tf
```diff
@@ -0,0 +1,15 @@
+resource "env0_team" "team" {
+  name = "team"
+}
+
+resource "env0_custom_role" "custom_role" {
+  name = "custom-role-sample"
+  permissions = [
+    "VIEW_ENVIRONMENT"
+  ]
```
sagydr, line 9: Can you please create custom role with organization settings? e.g. view-dashboard? (not sure where the names are taken from in the provider, but I can assist)

tests/integration/009_team/main.tf
```diff
@@ -22,4 +22,24 @@ output "team_resource_description" {
...
+
+resource "env0_custom_role" "custom_role1" {
+  name = "custom-role-${random_string.random.result}"
+  permissions = [
+    "CREATE_PROJECT"
+  ]
+}
+
+resource "env0_custom_role" "custom_role2" {
+  name = "custom-role-${random_string.random.result}2"
```
sagydr, line 35: if it's random-string, why do you need `2` at the end? it will be different than name from custom-role1, no?

tests/integration/009_team/main.tf
```diff
@@ -22,4 +22,24 @@ output "team_resource_description" {
...
+  permissions = [
+    "RUN_PLAN"
+  ]
+}
+
+resource "env0_team_organization_assignment" "organization_role_assignment" {
+  team_id = env0_team.team_resource.id
+  role_id = var.second_run ? env0_custom_role.custom_role1.id : env0_custom_role.custom_role2.id
```
sagydr, line 43: What's the purpose of this `var.second_run`? to show how to change assignments upon 2nd run?

tests/integration/009_team/main.tf
TomerHeber: it generates one random string - that way their names vary.

tests/integration/009_team/main.tf
TomerHeber: yes - this is how updates are tested.
first run is create... second run is an update (covers more use-cases).

examples/resources/env0_team_organization_assignment/resource.tf
TomerHeber: yeah.. I need to fix the example. Will do that.

TomerHeber: example fixed.

env0/resource_team_organization_assignment.go
sagydr: This is a `Read` command and reaching that code means assignment wasn't found among all organization assignments, so clearing the id will force it to be (re)created?
But why is it Drifted? the assignment isn't found in our DB, not in the tf state, no?

sagydr (approved):
looks good

env0/resource_team_organization_assignment.go
TomerHeber: It's in the tfstate but not in the backend.
So it will add it back to the backend.

If the user doesn't want that, he needs to remove it from the terraform file/state. (This will avoid drift).

PR #819, opened 2024-04-03
Fix: Cloudformation Template - Forcing refresh of a TerraformVersion …

### LINKED ISSUE #818: Cloudformation Template - Forcing refresh of a TerraformVersion field
**Describe the bug**
User is using a cloudformation template. 
and updated the provider to the latest version 1.17.2
and now the plan is showing that a "terraformVersion" is being replaced with a default value instead of keeping "" (empty string).
This is causing runs on a lot of unnecessary deployments to get triggered.

**To Reproduce**
Setup a cloudformation environment using older version of provider
update provider and rerun 

**Expected behavior**
No default value for terraformVersion

**Provider Version**
1.17.2

**Screenshots**
CleanShot 2024-04-03 at 08 51 25

TomerHeber:
…field

fixes #818 

### Solution

I'm unsure why it's occurring. But I applied a mitigation.
Tests seem to pass.

### DIFF
--- a/tests/harness.go
+++ b/tests/harness.go
@@ -39,7 +39,7 @@ func main() {
 			} else {
 				success, err := runTest(testName, destroyMode != "NO_DESTROY")
 				if !success {
-					log.Fatalln("Halting due to test failure:", err)
+					log.Fatalf("Halting due to test '%s' failure: %s\n", testName, err.Error())
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


tests/harness.go
```diff
@@ -39,7 +39,7 @@ func main() {
 			} else {
 				success, err := runTest(testName, destroyMode != "NO_DESTROY")
 				if !success {
-					log.Fatalln("Halting due to test failure:", err)
+					log.Fatalf("Halting due to test '%s' failure: %s\n", testName, err.Error())
```
TomerHeber, line 42: unrelated - but was requested to get fixed.

yaronya (approved):

PR #820, opened 2024-04-06
Feat: Refactor for Project Role assignments for Team

### LINKED ISSUE #817: Refactor for Project Role assignments for Team #807
The same as we did 
1. [here](https://github.com/env0/terraform-provider-env0/pull/813) for Organization Level assignments (team only).
2. [here](https://github.com/env0/terraform-provider-env0/pull/668) for Environment Level assignments (team + user).

We want to achieve the same for the new **Project** level **Team** assignments API.

We already implemented this but we now have synced our project team level to the new API too.
meaning we deprecating the old APIs, so we need to adjust the old resources related to Project<-->Team role assignments.

**deprecated APIs:**
1. [updateAssignments](https://docs.env0.com/reference/teams-update-assignments) 
2. [deleteSingleAssignment](https://docs.env0.com/reference/teams-delete-single-assignment) 
3. [createOrUpdateSingleAssignment](https://docs.env0.com/reference/teams-create-or-update-single-assignment) 
4. [getAssignments](https://docs.env0.com/reference/teams-get-assignments)

**new API:**
Use the new Team Role [APIs](https://docs.env0.com/reference/roles-get-team-role-assignments): Get, Upsert, Delete.
Same API as before, we just add `projectId`.
image

TomerHeber:
### Solution

Part of #817. Will create another PR for refactoring the API.

1. Modifed the API code.
2. Updated unit tests.
3. Modified the resource code to use the new API.
4. Updated acceptance tests.

### DIFF
--- a/client/api_client.go
+++ b/client/api_client.go
@@ -61,8 +61,8 @@ type ApiClientInterface interface {
 	TeamCreate(payload TeamCreatePayload) (Team, error)
 	TeamUpdate(id string, payload TeamUpdatePayload) (Team, error)
 	TeamDelete(id string) error
-	TeamProjectAssignmentCreateOrUpdate(payload TeamProjectAssignmentPayload) (TeamProjectAssignment, error)
-	TeamProjectAssignmentDelete(assignmentId string) error
+	TeamProjectAssignmentCreateOrUpdate(payload *TeamProjectAssignmentPayload) (*TeamProjectAssignment, error)
+	TeamProjectAssignmentDelete(projectId string, teamId string) error
 	TeamProjectAssignments(projectId string) ([]TeamProjectAssignment, error)
 	Environments() ([]Environment, error)
 	ProjectEnvironments(projectId string) ([]Environment, error)
--- a/client/api_client_mock.go
+++ b/client/api_client_mock.go
@@ -1851,10 +1851,10 @@ func (mr *MockApiClientInterfaceMockRecorder) TeamDelete(arg0 any) *gomock.Call 
 }
 
 // TeamProjectAssignmentCreateOrUpdate mocks base method.
-func (m *MockApiClientInterface) TeamProjectAssignmentCreateOrUpdate(arg0 TeamProjectAssignmentPayload) (TeamProjectAssignment, error) {
+func (m *MockApiClientInterface) TeamProjectAssignmentCreateOrUpdate(arg0 *TeamProjectAssignmentPayload) (*TeamProjectAssignment, error) {
 	m.ctrl.T.Helper()
 	ret := m.ctrl.Call(m, "TeamProjectAssignmentCreateOrUpdate", arg0)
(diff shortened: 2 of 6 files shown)


sagydr (approved):

razbensimon:
@HeverFarber Do we need a non-patch upgrade here?
i wonder how a new tf plan will look like after apgrading

TomerHeber:
> @TomerHeber @sagydr Do we need a non-patch upgrade here? i wonder how a new tf plan will look like after apgrading

@razbensimon - the plan should remain the same. Is there a way to test it? I can do a new major version if it makes sense. WDYT?

razbensimon:
if we can do it with a `patch` it is awesome, I just bring it up for you to test it.
meaning "upgrading" after having an existing project assignment in the _state_.
then try to:
1. plan/apply with no changes
3. delete assignment
4. change assignment
5. add assignment

client/api_client.go
```diff
@@ -61,8 +61,8 @@ type ApiClientInterface interface {
 	TeamCreate(payload TeamCreatePayload) (Team, error)
 	TeamUpdate(id string, payload TeamUpdatePayload) (Team, error)
 	TeamDelete(id string) error
-	TeamProjectAssignmentCreateOrUpdate(payload TeamProjectAssignmentPayload) (TeamProjectAssignment, error)
-	TeamProjectAssignmentDelete(assignmentId string) error
+	TeamProjectAssignmentCreateOrUpdate(payload *TeamProjectAssignmentPayload) (*TeamProjectAssignment, error)
+	TeamProjectAssignmentDelete(projectId string, teamId string) error
```
razbensimon, line 65: I think that in "api_client" we should have only 3 methods, that serve all scopes for `team` (and additional 3 methods for `users`).
So each method will get the scopeId as an argument (aka `environmentId`, `projectId`, or `organizationId`. They are optional and one of them should exist).
That's because all of those are the same API and routes.

meaning `TeamProjectAssignmentCreateOrUpdate` and `TeamProjectAssignmentDelete` should be removed. And we should not "drag" / keep the existence of the old deprecated APIs names.

WDYS?

razbensimon: the only scopes we haven't refactored yet - should preserve their methods here. Until we will refactor them too.
that's include:
1. org level user assignments
2. project level user assignments

TomerHeber: will be done in a follow-up PR.

TomerHeber:
> if we can do it with a `patch` it is awesome, I just bring it up for you to test it. meaning "upgrading" after having an existing project assignment in the _state_. then try to:
> 
> 1. plan/apply with no changes
> 2. delete assignment
> 3. change assignment
> 4. add assignment
> 
> EDIT: mostly I afraid of the `ProjectRole` & `Role` renames. how it will effect the state/plans?

As discussed on Slack. Only concern is if someone had mistakenly applied multiple assignments on the same project and team. (Not supported in the UI).

Issue #816, opened 2024-03-31
Referencing predefined "Project Viewer"/"Project Planner"/"Project Deployer"/"Project Admin" in a env0_team_environment_assignment


ankon:
**Is your feature request related to a problem? Please describe.**

For a business/enterprise plan one can create env0_custom_role resources, or use a data source to find the id that one uses in this snippet:

```hcl
data "env0_custom_role" "team role" {
  name = "Known Role"
}

resource "env0_team_environment_assignment" "assignment" {
  team_id        = var.team_id
  environment_id = env0_environment.environment.id
  role_id        = data.env0_custom_role.role.team_role.id
}
```

But: How would I reference the existing roles that exist also on the pro plan?

I tried in vain:
```hcl
data "env0_custom_role" "team_deployer" {
  name = "Project Deployer"
}
```

**Describe the solution you'd like**

I guess a couple:
* Have a data source for the predefined roles?
* Have a way to see the ids in the UI so I can hard-code them?
* Make the above `data "env0_custom_role"` work?
 
**Describe alternatives you've considered**

* Buy the business/enterprise plan, and make my own roles.
* Assign the roles manually in the UI, if needed with some `ignore_changes` magic

**Additional context**

The `env0_custom_role` produces:
```
│ Error: schema resource data serialization failed: id is empty
│ 
│   with module.XXXX.data.env0_custom_role.team_role,
│   on modules/env0/environment/main.tf line 80, in data "env0_custom_role" "team_role":
│   80: data "env0_custom_role" "team_role" {
│ 
```

tomporat247:
Hey, thank you bringing this up - we are looking into it 🙏

tomporat247:
Hey @ankon, we've looked into this request and we have an answer.

The `env0_custom_role` data resource is designed to be used with custom roles (which are eligible for higher tiers).
If you wish to use the default roles (the ones that exist on the "pro" tier) you can simply pass the literal id of the role you wish to use, here is an example:
```hcl
resource "env0_team_environment_assignment" "assignment" {
  team_id        = var.team_id
  environment_id = env0_environment.environment.id
  role_id        = "User"
}
```

Here is the "mapping" of the roles you can pass:
Organization roles - `User` / `Admin`
Project roles - `Viewer` / `Planner` / `Deployer` / `Admin`
Environment roles - `Viewer` / `Planner` / `Deployer` / `Admin`

tomporat247:
Yes of course, that's a very welcomed suggestion :)

PR #832, opened 2024-04-20
Feat: add role validator

TomerHeber:
Part of #816 

### Solution

Applied more refactoring. In this PR I added a validator and added a test for it.
Made some security fixes in go.mod.

Keeping this ticket open. last part is to deprecate the custom_role_id and role, and use role_id instead (for resourceTeamProjectAssignment).


env0/resource_team_organization_assignment.go
```diff
@@ -29,7 +29,7 @@ func resourceTeamOrganizationAssignment() *schema.Resource {
 				Type:             schema.TypeString,
 				Description:      "id of the assigned custom role. The following built-in roles can be passed as well: `User`, `Admin`",
 				Required:         true,
-				ValidateDiagFunc: ValidateNotEmptyString,
+				ValidateDiagFunc: NewRoleValidator([]string{"User", "Admin"}),
```
razbensimon, line 32: I feel like `[]string{"User", "Admin"}` should be a constant instance that can be imported from `builtin_roles.go`
same for other scopes

client/builtin_roles.go
```diff
@@ -7,6 +7,6 @@ const (
 	Viewer   string = "Viewer"
 )
 
-func IsBuiltinProjectRole(role string) bool {
+func IsBuiltinRole(role string) bool {
```
razbensimon, line 10: umm but `IsBuiltinProjectRole` is actually more accurate

env0/validators.go
```diff
@@ -129,3 +117,29 @@ func ValidateTtl(i interface{}, path cty.Path) diag.Diagnostics {
 
 	return nil
 }
+
+func NewRoleValidator(supportedBuiltInRoles []string) schema.SchemaValidateDiagFunc {
```
razbensimon, line 121: I don't thing `New` as a prefix in the name is a good choice.
New until when? 🙃

can it be just `RoleValidator`?

env0/validators.go
```diff
@@ -129,3 +117,29 @@ func ValidateTtl(i interface{}, path cty.Path) diag.Diagnostics {
...
+	return func(i interface{}, p cty.Path) diag.Diagnostics {
+		role := i.(string)
+
+		if role == "" {
+			return diag.Errorf("may not be empty")
+		}
+
+		if !client.IsBuiltinRole(role) {
+			// Custom role.
```
razbensimon, line 130: even better to create `IsCustomRole` for explicit readability.
inside the common file

env0/validators.go
TomerHeber: It creates a call back function that acts as the validator (based on the given input).
That's why a validator is used.

env0/resource_team_organization_assignment.go
TomerHeber: Sure.

env0/validators.go
TomerHeber: Sure.

TomerHeber:
@razbensimon see commit for additional fixes:
https://github.com/env0/terraform-provider-env0/pull/832/commits/6ed9a03d683f8af5be2fd031448322b5d925f318

razbensimon (approved):
Approving but still think the comment about the "new" word in the name should be addressed.

env0/validators.go
razbensimon: I didnt get your response, I am talking about the function name

TomerHeber: Notice it returns a new function.
In other words, it creates something.
Some validators do not create anything and are without the "new" some do and have the "new".

TomerHeber: I'll merge - if it's still an issue.
We can discuss more - the issue will remain open anyways, as I plan another final PR.

PR #842, opened 2024-05-01
Fix: resource 'env0_environment_scheduling' does not pass destroy_cro…

### LINKED ISSUE #841: resource "env0_environment_scheduling" does not pass destroy_cron=null on updates
**Describe the bug**
When using env0_environment_scheduling resource with a 
`destroycron=null` on a subsequent update to the resource we get this error:

```
Error: could not create or update environment scheduling: 400 Bad Request: /body/destroy must have required property 'cron', /body/destroy must be null, /body/destroy must match a schema in anyOf
```

**How To Reproduce**

1. tofu apply env0_environment_scheduling on an existing environment.
```
terraform {
  required_providers {
    env0 = {
      source = "env0/env0"
      version = "~> 1.18.5"
    }
  }
}

locals {
    environment_id = "63a4a4e1-501f-4617-8b41-a46a7238c73c"
}

resource "env0_environment_scheduling" "environment_scheduling" {
  environment_id = local.environment_id
  deploy_cron    = "45 21 * * 0,2,7"
  destroy_cron   = null
}
```

2. modify the env0_environment_scheduling deploy_cron to something else.

```
resource "env0_environment_scheduling" "environment_scheduling" {
  environment_id = local.environment_id
  deploy_cron    = "45 21 * * *"
  destroy_cron   = null
}
```

3. Run tofu apply again ^
4. Get error from backend.

**Expected behavior**
Resource should get updated without errors.

**Provider Version**
1.18.5

**Screenshots**
CleanShot 2024-04-30 at 14 32 24

**Additional context**
n/a

TomerHeber:
…n=null on updates

fixes #841 

### Solution

1. Passing null instead of a "disabled" cron.
2. Updated an integration test.

### DIFF
--- a/env0/resource_environment_scheduling.go
+++ b/env0/resource_environment_scheduling.go
@@ -69,6 +69,14 @@ func resourceEnvironmentSchedulingCreateOrUpdate(ctx context.Context, d *schema.
 		return diag.Errorf("schema resource data deserialization failed: %v", err)
 	}
 
+	if payload.Deploy != nil && !payload.Deploy.Enabled {
+		payload.Deploy = nil
+	}
+
+	if payload.Destroy != nil && !payload.Destroy.Enabled {
+		payload.Destroy = nil
+	}
+
 	if _, err := apiClient.EnvironmentSchedulingUpdate(environmentId, payload); err != nil {
 		return diag.Errorf("could not create or update environment scheduling: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


env0/resource_environment_scheduling.go
```diff
@@ -69,6 +69,14 @@ func resourceEnvironmentSchedulingCreateOrUpdate(ctx context.Context, d *schema.
 		return diag.Errorf("schema resource data deserialization failed: %v", err)
 	}
 
+	if payload.Deploy != nil && !payload.Deploy.Enabled {
```
TomerHeber, line 72: This basically passes a null instead of a disabled cron.
Would have been better to fix it in the backend. However, for the sake of time I've fixed it here.

yaronya (approved):

Issue #829, opened 2024-04-17
Add support to new ConfigurationSet functionality


GiliFaroEnv0:
We are adding a new Feature called `ConfigurationSet` which will provide the option to group `Configuration properties` with a display name and a description and assign/unassign them to the different entities as one unit.

We need to create resources to create `ConfigurationSet` and assign it to an entity.

`resource ConfigurationSet` {
[create using create API
](https://docs.env0.com/reference/configuration-create-configuration-set)
[edit using edit API](https://docs.env0.com/reference/configuration-update-configuration-set)
[delete using delete API
](https://docs.env0.com/reference/configuration-delete-configuration-set)}
}

`resource AssignConfigurationSet` {
[assign using assign API
](https://docs.env0.com/reference/configuration-assign-configuration-set-to-scope)}
unassign using API with similar props but DELETE (will be merged later today)
}

GiliFaroEnv0:
![](https://github.trello.services/images/mini-trello-icon.png) [Open ticket for the provider](https://trello.com/c/HONJg9Ij/12680-open-ticket-for-the-provider)

TomerHeber:
@GiliFaroEnv0 

A couple of questions/comments:
1. The GET response does not return the configurations? It is needed. (https://docs.env0.com/reference/configuration-find-configuration-set-by-id)
2. To which resource is a configuration set assigned to?
3. The update operation is delta changes? or is it an upsert?

I'm sure most of these questions can be answered by myself if the feature is available in the UI.
Is it available and ready to be used in the UI?

GiliFaroEnv0:
> @GiliFaroEnv0
> 
> A couple of questions/comments:
> 
> 1. The GET response does not return the configurations? It is needed. (https://docs.env0.com/reference/configuration-find-configuration-set-by-id)
> 2. To which resource is a configuration set assigned to?
> 3. The update operation is delta changes? or is it an upsert?
> 
> I'm sure most of these questions can be answered by myself if the feature is available in the UI. Is it available and ready to be used in the UI?

1. yes you need to also query [this API](https://docs.env0.com/reference/configuration-find-variables-by-scope) with queryParam `setId`
2. currently you can only use it with the extra resource (you will create) `AssignConfigurationSet`
3.delta changes like we have in variables of the environment resource

PR #853, opened 2024-05-21
Feat: add API calls for configuration set

TomerHeber:
### Solution

Part of #829

1. Added the API calls required for configuration set.
2. Added unit tests for configuration set.

### DIFF
--- a/client/configuration_set.go
+++ b/client/configuration_set.go
@@ -1,0 +1,101 @@
+package client
+
+type CreateConfigurationSetPayload struct {
+	Name        string `json:"name"`
+	Description string `json:"description"`
+	// if Scope is "organization", scopeId will be calculated in the functions.
+	Scope                   string                  `json:"scope"`   // "project" or "organization".
+	ScopeId                 string                  `json:"scopeId"` // project id or organization id.
+	ConfigurationProperties []ConfigurationVariable `json:"configurationProperties"`
+}
+
+type UpdateConfigurationSetPayload struct {
+	Name                           string                  `json:"name"`
+	Description                    string                  `json:"description"`
+	ConfigurationPropertiesChanges []ConfigurationVariable `json:"configurationPropertiesChanges"` // delta changes.
+}
+
+type ConfigurationSet struct {
+	Id string `json:"id"`
+}
+
+func (client *ApiClient) ConfigurationSetCreate(payload *CreateConfigurationSetPayload) (*ConfigurationSet, error) {
+	var result ConfigurationSet
+	var err error
+
+	if payload.Scope == "organization" {
+		payload.ScopeId, err = client.OrganizationId()
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if err := client.http.Post("/configuration-sets", payload, &result); err != nil {
+		return nil, err
+	}
+
+	return &result, nil
+}
+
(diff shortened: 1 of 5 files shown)


client/configuration_set_test.go
```diff
@@ -0,0 +1,200 @@
...
+		BeforeEach(func() {
+			httpCall = mockHttpClient.EXPECT().
+				Delete("/configuration-sets/"+id, nil).
+				Do(func(path string, request interface{}) {}).
+				Times(1)
+
+			apiClient.ConfigurationSetDelete(id)
+		})
+
+		It("Should call delete once", func() {})
```
GiliFaroEnv0, line 166: its empty...

GiliFaroEnv0: don't you prefer to do in in the test not in the before each?

client/configuration_set.go
```diff
@@ -0,0 +1,101 @@
...
+	var result []ConfigurationVariable
+
+	organizationId, err := client.OrganizationId()
+	if err != nil {
+		return nil, err
+	}
+
+	if err := client.http.Get("/configuration", map[string]string{
+		"setId":          setId,
+		"organizationId": organizationId,
```
GiliFaroEnv0, line 96: don't send the `organizationId` it will return both the variables of the set and the organization...
You should send only the `setId`

client/configuration_set.go
```diff
@@ -0,0 +1,101 @@
...
+
+	if err := client.http.Get("/configuration-sets", map[string]string{
+		"scope":               scope,
+		"scopeId":             scopeId,
+		"includeHigherScopes": "false",
+	}, &result); err != nil {
+		return nil, err
+	}
+
+	return result, nil
```
GiliFaroEnv0, line 69: I'm not sure why you need this API...
In the FE we want to list the sets 
do you plan to add` all_sets data` ?

client/configuration_set.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+type UpdateConfigurationSetPayload struct {
+	Name                           string                  `json:"name"`
+	Description                    string                  `json:"description"`
+	ConfigurationPropertiesChanges []ConfigurationVariable `json:"configurationPropertiesChanges"` // delta changes.
+}
+
+type ConfigurationSet struct {
+	Id string `json:"id"`
```
GiliFaroEnv0, line 19: the return also contain the `name` and the `description`

GiliFaroEnv0 (commented):
Left 1 comment which is a must the rest is recommendations
Do you plan to add the assignment/unassignment API in a separate PR?

client/configuration_set.go
TomerHeber: got it! thanks!

client/configuration_set.go
TomerHeber: sure I can remove it for now.

client/configuration_set.go
TomerHeber: yes I know. Not really needed. But I'll add it.

client/configuration_set_test.go
TomerHeber: this will test that delete was called once.
Without it, it test's nothing...

GiliFaroEnv0 (approved):
NICE work as always

PR #857, opened 2024-05-21
Feat: add API calls for configuration set assignment

TomerHeber:
### Solution

Part of #829 

1. Added API calls configuration sets assignments.
2. Added unit tests.


GiliFaroEnv0 (approved):
NICE

PR #859, opened 2024-05-27
Feat: add the resource configuration_set - WIP

TomerHeber:
### Solution

Part of #857 

1. Added the new resource.
2. Implemented create, read and delete (update not started).
3. Added one sanity test (more tests is WIP).

### DIFF
--- a/client/configuration_set_test.go
+++ b/client/configuration_set_test.go
@@ -152,3 +152,5 @@ var _ = Describe("Configuration Set", func() {
 		})
 	})
 })
+
+// TODO add more tests...
--- a/env0/resource_variable_set.go
+++ b/env0/resource_variable_set.go
@@ -1,0 +1,376 @@
+package env0
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/env0/terraform-provider-env0/client"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+var variableSetVariableSchema *schema.Resource = &schema.Resource{
+	Schema: map[string]*schema.Schema{
+		"name": {
+			Type:        schema.TypeString,
+			Description: "variable name",
+			Required:    true,
+		},
+		"value": {
+			Type:        schema.TypeString,
+			Description: "variable value for 'hcl', 'json', or 'text' format",
+			Optional:    true,
+		},
+		"dropdown_values": {
+			Type:        schema.TypeList,
+			Description: "a list of variable values for 'dropdown' format",
+			Optional:    true,
+			Elem: &schema.Schema{
+				Type: schema.TypeString,
+			},
+		},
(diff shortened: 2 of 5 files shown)


client/configuration_set_test.go
```diff
@@ -152,3 +155,5 @@ var _ = Describe("Configuration Set", func() {
 		})
 	})
 })
+
+// TODO add more tests...
```
GiliFaroEnv0, line 159: todo?

env0/resource_variable_set.go
```diff
@@ -0,0 +1,451 @@
...
+			Optional:    true,
+			Default:     false,
+		},
+		"format": {
+			Type:             schema.TypeString,
+			Description:      "the value format: 'text' (free text), 'dropdown' (dropdown list), 'hcl', 'json'. Note: 'hcl' and 'json' can only be used in terraform variables.",
+			Optional:         true,
+			Default:          "text",
+			ValidateDiagFunc: NewStringInValidator([]string{"text", "dropdown", "hcl", "json"}),
+		},
```
GiliFaroEnv0, line 52: is there a way not to duplicate it between all the usages, environment / regular variable??

not need to be in this PR just for the future

env0/resource_variable_set.go
```diff
@@ -0,0 +1,451 @@
...
+	return nil
+}
+
+type mergedVariables struct {
+	currentVariables []client.ConfigurationVariable
+	newVariables     []client.ConfigurationVariable
+	deletedVariables []client.ConfigurationVariable
+}
+
+func mergeVariables(schema []client.ConfigurationVariable, api []client.ConfigurationVariable) *mergedVariables {
```
GiliFaroEnv0, line 293: again here the same logic should also be in the environment resource can we extract it into a common and use the same un both?

GiliFaroEnv0 (approved):
Amazing work as always!

env0/resource_variable_set.go
TomerHeber: The current configuration variables in the environment has more features, but that's not the problem...
The implementation is lacking (bugs/limitations) and the schema is confusing (doesn't align with the UI).
But yes, once we add variables_sets to environment,  common code will be used.

About environment configuration - we could deprecate current implementation... and replace it  but that's a separate (large) effort...

env0/resource_variable_set.go
TomerHeber: The current configuration variables in the environment has more features, but that's not the problem...
The implementation is lacking (bugs/limitations) and the schema is confusing (doesn't align with the UI).
But yes, once we add variables_sets to environment, common code will be used.

About environment configuration - we could deprecate current implementation... and replace it but that's a separate (large) effort...

client/configuration_set_test.go
TomerHeber: that's leftovers... forgot to remove it...

TomerHeber: will remove it on the next PR.

Issue #856, opened 2024-05-21
A new resource for environment importing


alonnoga:
**Is your feature request related to a problem? Please describe.**
We are adding a feature to allow importing tofu/terraform stacks to env0 environments 
We need a new resource for it
**Describe the solution you'd like**
We need a new resource, these are the entities endpoints
( get ) https://docs.env0.com/reference/environment-import-find-staging-environment-by-id
( create ) https://docs.env0.com/reference/environment-import-create-staging-environment
( update ) https://docs.env0.com/reference/environment-import-update-staging-environment

When the resource is removed from the stack it should not call delete ( it should just get removed from tf state )
You are not allowed to change an environment import's workspace ( you have to create a new import )

TomerHeber:
@alonnoga is this done?
Can we close it?

PR #864, opened 2024-05-28
Feat: add environment import id scope to configuration variable

alonnoga:
Part of #856
We added a new scope for environment import
It should not delete the variable when removed ( like how we don't delete environment imports when the resource is removed from the configuration )


alonnoga:
We decided to reuse environment variable scope instead, I'll create another pr to allow not deleting a variable when removed from config

PR #866, opened 2024-05-30
Chore: add soft delete to configuration variable

alonnoga:
Part of :#856

We need to be able to remove configuration variables from code without actually deleting them

In environment import we initialize variables for the user to review and create an environment from
These resources are not expected to be committed, managed other then applying them once, so we need them to not be removed once the configuration is removed

Edit:
After adding a soft delete flag to config variable it I realized we should have added one to env import resource to make this strange behavior explicit 👍🏻 ( So I added that as well )


env0/resource_configuration_variable.go
```diff
@@ -110,6 +110,11 @@ func resourceConfigurationVariable() *schema.Resource {
 				Description: "the value of this variable must match provided regular expression (enforced only in env0 UI)",
 				Optional:    true,
 			},
+			"soft_delete": {
```
alonnoga, line 113: @TomerHeber now that I think about it we could have added it to "environment import resource" as well so that it won't be implicit that its not actually deleted

env0/resource_configuration_variable.go
```diff
@@ -110,6 +110,11 @@ func resourceConfigurationVariable() *schema.Resource {
...
+				Type:        schema.TypeBool,
+				Description: "soft delete the configuration variable, once removed from the configuration it won't be deleted from env0",
+				Optional:    true,
```
TomerHeber, line 116: Please add a default value. This prevents many issues with bool types.

env0/resource_configuration_variable.go
```diff
@@ -237,6 +242,11 @@ func resourceConfigurationVariableUpdate(ctx context.Context, d *schema.Resource
 }
 
 func resourceConfigurationVariableDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	// don't delete if soft delete is set
+	if softDelete, ok := d.GetOk("soft_delete"); ok && softDelete.(bool) {
```
TomerHeber, line 246: If you put a default you can just get the softDelete.
GetOk does not work properly with booleans...

env0/resource_environment_import.go
```diff
@@ -58,6 +58,11 @@ func resourceEnvironmentImport() *schema.Resource {
 				Description: "iac version of the environment",
 				Optional:    true,
 			},
+			"soft_delete": {
+				Type:        schema.TypeBool,
+				Description: "soft delete the configuration variable, once removed from the configuration it won't be deleted from env0",
+				Optional:    true,
```
TomerHeber, line 64: same here - please add a default.

env0/resource_environment_import.go
```diff
@@ -126,9 +131,17 @@ func resourceEnvironmentImportUpdate(ctx context.Context, d *schema.ResourceData
 	return nil
 }
 
-// should not actually delete the environment import
-// this resource is used to populate the environment import wizard data
-// we don't want to delete the environment import after it's been created outside the wizard
 func resourceEnvironmentImportDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	// don't delete if soft delete is set
+	if softDelete, ok := d.GetOk("soft_delete"); ok && softDelete.(bool) {
```
TomerHeber, line 136: same here.. once you have a default.. just get the value...

TomerHeber (changes requested):
LGTM - just a few small changes required.

env0/resource_configuration_variable.go
alonnoga: what are the issues?

env0/resource_configuration_variable.go
alonnoga: [eed48af](https://github.com/env0/terraform-provider-env0/pull/866/commits/eed48af5b5c5d63a6c376f7d7d746f992270db90)

env0/resource_environment_import.go
alonnoga: [eed48af](https://github.com/env0/terraform-provider-env0/pull/866/commits/eed48af5b5c5d63a6c376f7d7d746f992270db90)

env0/resource_environment_import.go
alonnoga: [eed48af](https://github.com/env0/terraform-provider-env0/pull/866/commits/eed48af5b5c5d63a6c376f7d7d746f992270db90)

env0/resource_environment_import.go
```diff
@@ -58,6 +58,12 @@ func resourceEnvironmentImport() *schema.Resource {
 				Description: "iac version of the environment",
 				Optional:    true,
 			},
+			"soft_delete": {
+				Type:        schema.TypeBool,
+				Description: "soft delete the configuration variable, once removed from the configuration it won't be deleted from env0",
```
TomerHeber, line 63: Is this comment correct? default is false...

alonnoga: yeah if they set the field ( to true ) it won't delete
I just defaulted it to false so it won't be set by default ( we want soft delete to be explicit )

we generate the code that uses these flags so we'll just set `soft_delete = true` we don't really expect customers to use those flags, but we just didn't want the resource to act in unexpected ways without it being explicit

alonnoga:
@TomerHeber please also see this commit [22a1e2e](https://github.com/env0/terraform-provider-env0/pull/866/commits/22a1e2e6f5a0033dc3e12a29f157087edabc806c)

TomerHeber (approved):

PR #870, opened 2024-06-03
Feat: add support for variable set assignment resource

TomerHeber:
Part of #829 

### Solution

1. Added the assignment resource.
2. Added an API call.
3. Added unit tests.
4. Added acceptance tests.
5. Added integration tests.
6. Updated examples.

### DIFF
--- a/env0/resource_variable_set_assignment.go
+++ b/env0/resource_variable_set_assignment.go
@@ -24,9 +24,9 @@ func resourceVariableSetAssignment() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"scope": {
 				Type:             schema.TypeString,
-				Description:      "the resource(scope) type to assign to. Valid values: 'template', 'environment', 'workflow', 'organization', 'project'",
+				Description:      "the resource(scope) type to assign to. Valid values: 'template', 'environment', 'module', 'organization', 'project', 'deployment'",
 				Required:         true,
-				ValidateDiagFunc: NewStringInValidator([]string{"template", "environment", "workflow", "organization", "project"}),
+				ValidateDiagFunc: NewStringInValidator([]string{"template", "environment", "module", "organization", "project", "deployment"}),
 				ForceNew:         true,
 			},
 			"scope_id": {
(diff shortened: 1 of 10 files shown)


sagilaufer1992:
there are usages of both `variable` and `configuration` sets. can we make use of a single term? preferably the one used for single variables

env0/resource_variable_set_assignment.go
```diff
@@ -0,0 +1,204 @@
...
+	return &schema.Resource{
+		CreateContext: resourceVariableSetAssignmentCreate,
+		UpdateContext: resourceVariableSetAssignmentUpdate,
+		ReadContext:   resourceVariableSetAssignmentRead,
+		DeleteContext: resourceVariableSetAssignmentDelete,
+
+		Schema: map[string]*schema.Schema{
+			"scope": {
+				Type:             schema.TypeString,
+				Description:      "the resource(scope) type to assign to. Valid values: 'template', 'environment', 'workflow', 'organization', 'project'",
```
sagilaufer1992, line 27: the allowed values are: organization, project, template, module, environment, deployment
https://docs.env0.com/reference/configuration-find-set-assignments-by-scope see what the assignmentScope is allowed to be.

sagilaufer1992 (commented):
looks great, left some comments

TomerHeber:
> there are usages of both `variable` and `configuration` sets. can we make use of a single term? preferably the one used for single variables

yeah for the printing of messages. Sure thing.

TomerHeber:
> looks great, left some comments

@sagilaufer1992 - thanks!
This is still work in progress. I will add the rest today.

env0/resource_variable_set_assignment.go
TomerHeber: thank you. will fix.

tests/integration/033_variable_set/main.tf
```diff
@@ -70,3 +70,9 @@ resource "env0_variable_set" "project_scope" {
     format = "text"
   }
 }
+
+resource "env0_variable_set_assignment" "assignment" {
+  scope    = "project"
+  scope_id = env0_project.project.id
+  set_ids  = var.second_run ? [env0_variable_set.org_scope.id] : [env0_variable_set.project_scope.id]
```
sagilaufer1992, line 77: the test seems a bit missing... can we check the outcome of that? using output or some other method, to see that the assignment doesn't exist on the org after the second run, and exists where it should after on each run?

sagilaufer1992 (approved):
LGTM
requested to update the integration test to make sure the expected outcome is achieved, not just alter the resource

tests/integration/033_variable_set/main.tf
TomerHeber: This is tested in the acceptance tests.
The integration tests are limited in their functionality and are more of a sanity just to make sure things really work...
I don't think there's an easy way to test what you're asking with the current integration tests... /:

sagilaufer1992: Alright. As long as there is a test for it

PR #872, opened 2024-06-05
Feat: importing templateless environment causes the configuration to …

### LINKED ISSUE #865: Bug - Importing templateless environment causes the configuration to be unplannable
**Describe the bug**
When importing an environment that does not use a template, into the configuration, the configuration then becomes unplannable

**To Reproduce**
1. Manually create an environment in env0, without using a template (requires "Create Environment from VCS" permission)
2. Locally use the following configuration (See bottom of the reproduction steps)
3. Run `tofu init`
4. Run `tofu import env0_environment.my_env <ENV_ID>`
5. Run `tofu plan`
6. Get the error `template_id may not be modified, create a new environment instead`

I used an OpenTofu environment using "Other VCS" in the UI, and the following configuration:
```hcl
terraform {
  required_providers {
    env0 = {
      source = "env0/env0"
    }
  }
}

provider "env0" {
  ...
}

resource "env0_environment" "e" {
  name                       = "<ENVIRONMENT_NAME>"
  project_id                 = "<PROJECT_ID>"

  without_template_settings {
    repository = "<REPOSITORY_URL>"
    path       = "<PATH>"
    type       = "opentofu"
  }
}
```

**Expected behavior**
Plan should work, and should not update the configuration

**Provider Version**
1.18.9

**Screenshots**
None

**Additional context**
Seems to either be an issue with wrongfully importing some data with a templateless environment (we probably don't need to import `template_id` field and only import `without_template_settings` block), or be an issue with our validation

TomerHeber:
…be unplannable

resolves #865 

### Solution

1. Added code to import the template information for templatless use-case.
2. Added acceptance tests.
3. Added integration test.

### DIFF
--- a/env0/resource_template.go
+++ b/env0/resource_template.go
@@ -301,7 +301,7 @@ func resourceTemplateRead(ctx context.Context, d *schema.ResourceData, meta inte
 		return nil
 	}
 
-	if err := templateRead("", template, d); err != nil {
+	if err := templateRead("", template, d, false); err != nil {
 		return diag.Errorf("%v", err)
 	}
 
@@ -422,7 +422,7 @@ func templateCreatePayloadFromParameters(prefix string, d *schema.ResourceData) 
 }
 
 // Reads template and writes to the resource data.
-func templateRead(prefix string, template client.Template, d *schema.ResourceData) error {
+func templateRead(prefix string, template client.Template, d *schema.ResourceData, isImport bool) error {
 	pathPrefix := "path"
 	terragruntTfBinaryPrefix := "terragrunt_tf_binary"
 	terraformVersionPrefix := "terraform_version"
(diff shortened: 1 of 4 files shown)


env0/resource_template.go
```diff
@@ -440,12 +440,15 @@ func templateRead(prefix string, template client.Template, d *schema.ResourceDat
...
 	// This helps avoid drifts when defaulting to 'opentofu' for new 'terragrunt' templates, and 'terraform' for existing 'terragrunt' templates.
 	// 'template.TerragruntTfBinary' field is set to 'omitempty'. Therefore, the state isn't modified if `template.TerragruntTfBinary` is an empty string.
-	if terragruntTfBinary == "" {
-		template.TerragruntTfBinary = ""
-	}
-	// Same explanation as above.
-	if terraformVersion == "" {
-		template.TerraformVersion = ""
+	// This is not true for imports - because the shcema is empty irrespective in that case.
+	if !isImport {
```
RLRabinowitz, line 444: Is this condition necessary? If the schema is empty anyway
What would be the effect of removing this condition?

RLRabinowitz (approved):

env0/resource_template.go
TomerHeber: it will not import terraform_version properly.
If the schema is empty (which it is in import), it will remove these values from the struct, afterwards when updates the schema based on the steuct, terraform_version will not be set.

This occurred to me running the test. So I added this condition.

PR #893, opened 2024-07-10
Feat: support new apply-all environment property

### LINKED ISSUE #890: Support new apply-all environment property
We have added a new "apply-all" feature (docs will be updated soon) to be able to run more commands thru PR comments.
However we would like to enable this feature only on environments which explicitly enabled this feature.
This is possible thru the UI, a new `Run PR Comments Commands using an Alias` checkbox added in the environment's Continuous Deployment section.
Also this checkbox is found on the Project Policies configuration, to be applied automatically to new environments created.
So feature request is to be able to support this feature when creating environments thru the provider

When creating environment, support having the `vcsPrCommentsEnabled` field like we support in `services/environment-service/api.d.ts`
Same with project policies, support the new field `vcsPrCommentsEnabledDefault`

**Additional context**
This is the new environment's CD checkbox:
The "alias" is now optional, i.e. the `vcsPrCommentsEnabled` field may be true without providing any alias, but if alias is provided, the `vcsPrCommentsEnabled` must be checked.
image

This is the new project policies checkbox:
image

** Notes **
The UI will allow setting alias only when `vcsPrCommentsEnabled=true`. We are not adding validations so we won't break older providers, so we will set `vcsPrCommentsEnabled=true` whenever alias is given in the API/provider request.

TomerHeber:
resolves #890 

### Solution

1. Added the new field in the APIs for project policy and environment.
2. Updated the schemas for project policy and environment.
3. Updated the acceptance tests.
4. Updated the integration tests.


env0/resource_environment.go
```diff
@@ -266,6 +266,12 @@ func resourceEnvironment() *schema.Resource {
 				Description: "set an alias for this environment in favor of running VCS commands using PR comments against it. Additional details: https://docs.env0.com/docs/plan-and-apply-from-pr-comments",
 				Optional:    true,
 			},
+			"vcs_pr_comments_enabled": {
+				Type:        schema.TypeBool,
+				Description: "set to 'true' to enable running VCS PR commands using PR comments. This can be set to 'true' (enabled) without setting alias in 'vcs_commands_alias'.",
+				Optional:    true,
+				Default:     false,
```
TomerHeber, line 273: note: that the default is false.

env0/resource_environment.go
```diff
@@ -380,6 +386,12 @@ func setEnvironmentSchema(ctx context.Context, d *schema.ResourceData, environme
 		return fmt.Errorf("schema resource data serialization failed: %v", err)
 	}
 
+	if val := d.Get("vcs_pr_comments_enabled").(bool); !val && environment.VcsPrCommentsEnabled {
```
TomerHeber, line 389: this is important. Although the default is 'false' it's possible vcs_pr_comments_enabled is set to 'true' (E.g. alias was used).
To avoid drifts - it's ignored in this case.

env0/resource_environment.go
```diff
@@ -882,6 +894,11 @@ func getCreatePayload(d *schema.ResourceData, apiClient client.ApiClientInterfac
 		return client.EnvironmentCreate{}, diag.Errorf("schema resource data deserialization failed: %v", err)
 	}
 
+	if val := d.Get("vcs_commands_alias").(string); len(val) > 0 {
+		// if alias is set - VcsPrCommentsEnabled must be true.
+		payload.VcsPrCommentsEnabled = true
```
TomerHeber, line 899: alias is used - pass 'true' in the API call.

env0/resource_project_policy.go
```diff
@@ -111,6 +111,12 @@ func resourceProjectPolicy() *schema.Resource {
 				Optional:         true,
 				ValidateDiagFunc: ValidateCronExpression,
 			},
+			"vcs_pr_comments_enabled_default": {
+				Type:        schema.TypeBool,
+				Description: "if 'true' all environments created in this project via the UI will be created with an 'enabled' running VCS PR commands using PR comments. Note: in the provider when creating an environment this is ignored, and must be configured explicitly by setting 'vcs_pr_comments_enabled' to 'true' or setting an alias in 'vcs_commands_alias'.",
```
TomerHeber, line 116: See the note.
This is a design decision for the provider.

env0/resource_environment.go
sagydr: yes awesome

env0/resource_environment.go
sagydr: We actually have code in the Backend that does that - sets to true if alias is given. Maybe it's not necessary here too.. WDYT?

env0/resource_project_policy.go
sagydr: nice. But what do you mean by `design decision for the provider.`?
This is the purpose of the project policies, no?

tests/integration/011_policy/main.tf
```diff
@@ -18,17 +18,19 @@ resource "env0_project_policy" "test_policy" {
...
-  skip_redundant_deployments    = true
+  project_id                      = env0_project.test_project.id
+  number_of_environments          = 1
+  number_of_environments_total    = 1
+  requires_approval_default       = true
+  include_cost_estimation         = true
+  skip_apply_when_plan_is_empty   = true
+  disable_destroy_environments    = true
+  skip_redundant_deployments      = true
+  vcs_pr_comments_enabled_default = true
```
sagydr, line 33: This tests the project policy configuration?

sagydr (approved):
looks good, approved with minor comments/questions

client/environment.go
```diff
@@ -122,6 +122,7 @@ type Environment struct {
 	LatestDeploymentLog         DeploymentLog `json:"latestDeploymentLog"`
 	TerragruntWorkingDirectory  string        `json:"terragruntWorkingDirectory,omitempty"`
 	VcsCommandsAlias            string        `json:"vcsCommandsAlias"`
+	VcsPrCommentsEnabled        bool          `json:"vcsPrCommentsEnabled" tfschema:"-"`
```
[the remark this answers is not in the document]
razbensimon: @TomerHeber can you explain more?
we indeed need that- 
if a customer
1. use the the provider version
2. do not set `vcsPrCommentsEnabled`
3. do set `VcsCommandsAlias`

then the BE response may return `vcsPrCommentsEnabled` `true` in the response.
we should not make a drift for that.

use case - a customer that used the alias in old provider version, then did an upgrade.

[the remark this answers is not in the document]
TomerHeber: yeah - in the schema it will be false.
But in the API it will pass true... because alias is used in the schema.
There's a part where this drift is ignored.

tests/integration/011_policy/main.tf
TomerHeber: just sanity really.
The better test is the acceptance testing.

client/project_policy.go
```diff
@@ -18,6 +18,7 @@ type Policy struct {
 	ForceRemoteBackend          bool    `json:"forceRemoteBackend"`
 	DriftDetectionCron          string  `json:"driftDetectionCron"`
 	DriftDetectionEnabled       bool    `json:"driftDetectionEnabled"`
+	VcsPrCommentsEnabledDefault bool    `json:"vcsPrCommentsEnabledDefault"`
```
razbensimon, line 21: can we put this one bellow `ContinuousDeploymentDefault`?
this is just my OCD yelling, and want those 3 to be together...

env0/resource_environment.go
TomerHeber: oh! interesting...

env0/resource_environment.go
```diff
@@ -266,6 +266,12 @@ func resourceEnvironment() *schema.Resource {
 				Description: "set an alias for this environment in favor of running VCS commands using PR comments against it. Additional details: https://docs.env0.com/docs/plan-and-apply-from-pr-comments",
 				Optional:    true,
 			},
+			"vcs_pr_comments_enabled": {
+				Type:        schema.TypeBool,
+				Description: "set to 'true' to enable running VCS PR commands using PR comments. This can be set to 'true' (enabled) without setting alias in 'vcs_commands_alias'.",
```
razbensimon, line 271: also can add docs link https://docs.env0.com/docs/plan-and-apply-from-pr-comments

```suggestion
			"vcs_pr_comments_enabled": {
				Type:        schema.TypeBool,
				Description: "set to 'true' to enable running VCS PR plan/apply commands using PR comments. This can be set to 'true' (enabled) without setting alias in 'vcs_commands_alias'.",
```

env0/resource_environment.go
sagydr: but... I kinda like having this logic here... @razbensimon WDYT?

env0/resource_environment.go
sagydr: Maybe even specifically to the Configuration section:
https://docs.env0.com/docs/plan-and-apply-from-pr-comments#configuration

env0/resource_project_policy.go
TomerHeber: going to make changes based on our discussion.

env0/resource_environment.go
TomerHeber: sure!

env0/resource_project_policy.go
razbensimon: @TomerHeber Not sure I got this too.
it should be something like this:
```suggestion
				Description: "if 'true' all new environments created in this project will be created with an 'enabled' VCS PR commands using PR comments.",
```

razbensimon: for a user to enable this feature, he should either provide 
'vcs_pr_comments_enabled' to 'true' or setting an alias in 'vcs_commands_alias', or set both which is explicit.

we can add a warning log when the user uses only `vcs_commands_alias`, this is just a suggestion. 
Anyways the BE takes care of it and forcibly set `vcs_pr_comments_enabled` if needed (not depending on client UI/provider/API)

razbensimon: let me know if we didnt understand you

env0/resource_environment.go
razbensimon: what it adds to the provider user if this logic exists already in BE?

env0/resource_project_policy.go
razbensimon: Indeed if the user enabled the project policy - but didn't add `vcs_pr_comments_enabled` on the environment resource IaC. this is a problem.

how we should handle this?

client/environment.go
```diff
@@ -144,6 +145,7 @@ type EnvironmentCreate struct {
 	TTL                         *TTL                     `json:"ttl,omitempty" tfschema:"-"`
 	TerragruntWorkingDirectory  string                   `json:"terragruntWorkingDirectory,omitempty"`
 	VcsCommandsAlias            string                   `json:"vcsCommandsAlias"`
+	VcsPrCommentsEnabled        *bool                    `json:"vcsPrCommentsEnabled,omitempty" tfschema:"-"`
```
TomerHeber, line 148: bool ptr to support tri-state - true/false/undefined.

client/environment.go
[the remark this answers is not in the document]
TomerHeber: The answer has changed.
the return value will now be ignored if nothing was configured in the schema. Not a perfect solution, but should work well for customers.

client/environment.go
```diff
@@ -193,6 +195,7 @@ type EnvironmentUpdate struct {
 	AutoDeployByCustomGlob      string `json:"autoDeployByCustomGlob"`
 	TerragruntWorkingDirectory  string `json:"terragruntWorkingDirectory,omitempty"`
 	VcsCommandsAlias            string `json:"vcsCommandsAlias,omitempty"`
+	VcsPrCommentsEnabled        *bool  `json:"vcsPrCommentsEnabled,omitempty" tfschema:"-"`
```
TomerHeber, line 198: bool ptr to support tri-state - true/false/undefined. (this is for update)

env0/resource_environment.go
```diff
@@ -266,6 +266,11 @@ func resourceEnvironment() *schema.Resource {
 				Description: "set an alias for this environment in favor of running VCS commands using PR comments against it. Additional details: https://docs.env0.com/docs/plan-and-apply-from-pr-comments",
 				Optional:    true,
 			},
+			"vcs_pr_comments_enabled": {
+				Type:        schema.TypeBool,
+				Description: "set to 'true' to enable running VCS PR plan/apply commands using PR comments. This can be set to 'true' (enabled) without setting alias in 'vcs_commands_alias'. Additional details: https://docs.env0.com/docs/plan-and-apply-from-pr-comments#configuration",
+				Optional:    true,
```
TomerHeber, line 272: generally it's good practice to set a default value.
But we need to support a tri-state. Setting a default, will enforce it to be only 2 states.

env0/resource_environment.go
```diff
@@ -380,6 +385,13 @@ func setEnvironmentSchema(ctx context.Context, d *schema.ResourceData, environme
 		return fmt.Errorf("schema resource data serialization failed: %v", err)
 	}
 
+	//lint:ignore SA1019 reason: https://github.com/hashicorp/terraform-plugin-sdk/issues/817
+	if _, exists := d.GetOkExists("vcs_pr_comments_enabled"); !exists {
```
TomerHeber, line 389: if not configured in the schema ignore whatever is returned in the response.
This is not perfect, but will work well for customers.

env0/resource_environment.go
```diff
@@ -724,7 +736,20 @@ func shouldDeploy(d *schema.ResourceData) bool {
 }
 
 func shouldUpdate(d *schema.ResourceData) bool {
-	return d.HasChanges("name", "approve_plan_automatically", "deploy_on_push", "run_plan_on_pull_requests", "auto_deploy_by_custom_glob", "auto_deploy_on_path_changes_only", "vcs_commands_alias", "is_remote_backend", "is_inactive", "is_remote_apply_enabled")
+	if d.HasChanges("name", "approve_plan_automatically", "deploy_on_push", "run_plan_on_pull_requests", "auto_deploy_by_custom_glob", "auto_deploy_on_path_changes_only", "vcs_commands_alias", "is_remote_backend", "is_inactive", "is_remote_apply_enabled", "vcs_pr_comments_enabled") {
+		return true
+	}
+
+	//lint:ignore SA1019 reason: https://github.com/hashicorp/terraform-plugin-sdk/issues/817
+	if val, exists := d.GetOkExists("vcs_pr_comments_enabled"); exists {
```
TomerHeber, line 744: hasChanges won't work in some scenarios. I had to use GetOkExists.
This would mean that some cases (and only if the schema is modified!) it may return 'true' for this when it shouldn't.
This would just mean an extra API call. But everything will work as expected.

env0/resource_environment.go
```diff
@@ -724,7 +736,20 @@ func shouldDeploy(d *schema.ResourceData) bool {
...
+		// if this field is set to 'false' will return that there is change each time.
+		// this is because the terraform SDK is unable to detecred changes between 'unset' and 'false' (sdk limitation).
+		if !val.(bool) {
```
TomerHeber, line 747: hasChanges works well when it's 'true'. So just the 'false' use-case.

env0/resource_environment.go
```diff
@@ -884,6 +909,11 @@ func getCreatePayload(d *schema.ResourceData, apiClient client.ApiClientInterfac
 		return client.EnvironmentCreate{}, diag.Errorf("schema resource data deserialization failed: %v", err)
 	}
 
+	//lint:ignore SA1019 reason: https://github.com/hashicorp/terraform-plugin-sdk/issues/817
+	if val, exists := d.GetOkExists("vcs_pr_comments_enabled"); exists {
```
TomerHeber, line 913: if the schema is set with 'vcs_pr_comments_enabled' that's when drifts are detected.
This will work well for customers. (But not perfect).

env0/resource_environment.go
```diff
@@ -1004,6 +1034,12 @@ func getUpdatePayload(d *schema.ResourceData) (client.EnvironmentUpdate, diag.Di
 		return client.EnvironmentUpdate{}, diag.Errorf("schema resource data deserialization failed: %v", err)
 	}
 
+	// Because the terraform SDK is unable to detecred changes between 'unset' and 'false' (sdk limitation): always set a value here (even if there's no change).
+	//lint:ignore SA1019 reason: https://github.com/hashicorp/terraform-plugin-sdk/issues/817
+	if val, exists := d.GetOkExists("vcs_pr_comments_enabled"); exists {
```
TomerHeber, line 1039: if the schema is set with 'vcs_pr_comments_enabled' use the value for update.
(Ignore the global policy).

env0/resource_environment_test.go
```diff
@@ -193,6 +215,84 @@ func TestUnitEnvironmentResource(t *testing.T) {
 			})
 		})
 
+		t.Run("vcs pr comments enabled", func(t *testing.T) {
```
yaronya, line 218: Can we specifically also test the scenario where `vcs_commands_alias` is sent without specifying `vcs_pr_comments_enabled` ?

TomerHeber: I don't think it really matters in these type of tests (since the tests are mocked).
I was asked not handle this use-case in the provider. Instead the backend will handle it.

I we already have a harness test passing an alias without vcs_pr_comments_enabled.

```
resource "env0_environment" "example" {
  depends_on    = [env0_template_project_assignment.assignment]
  force_destroy = true
  name          = "environment-${random_string.random.result}"
  project_id    = env0_project.test_project.id
  template_id   = env0_template.template.id
  configuration {
    name  = "environment configuration variable"
    value = "value"
  }
  approve_plan_automatically = true
  revision                   = "master"
  vcs_commands_alias         = "alias"
  drift_detection_cron       = var.second_run ? "*/5 * * * *" : "*/10 * * * *"
}
```

PR #907, opened 2024-07-25
Fix: env0_environment does not redeploy when revision is changed in w…

### LINKED ISSUE #906: `env0_environment` does not redeploy when revision is changed in `without_template_settings`
**Describe the bug**
When you change the `revision` field of an `env0_environment` resource - you will trigger a deployment. This is good. But when you have `without_template_settings`, you can't specify the `revision` directly on the `environment`, only nested within the `without_template_settings`, and when you change it - it **does not trigger a redeploy**.

**Expected behavior**
When changing the `revision` inside a `without_template_settings` block of an `env0_environment` resource, I expect the update to trigger a redeploy.

TomerHeber:
…ithout_template_settings

fixes #906 

### Solution

1. Update shouldDeploy to test that use-cases.
2. Some other small changes.
3. Updated the acceptance tests.

### DIFF
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -732,6 +732,12 @@ func shouldUpdateTemplate(d *schema.ResourceData) bool {
 }
 
 func shouldDeploy(d *schema.ResourceData) bool {
+	if _, ok := d.GetOk("without_template_settings.0"); ok {
+		if d.HasChange("without_template_settings.0.revision") {
+			return true
+		}
+	}
+
 	return d.HasChanges("revision", "configuration", "sub_environment_configuration", "variable_sets")
 }
 
@@ -1126,8 +1132,7 @@ func getDeployPayload(d *schema.ResourceData, apiClient client.ApiClientInterfac
 	var err error
 
 	if isTemplateless(d) {
-		templateId, ok := d.GetOk("without_template_settings.0.id")
-		if ok {
+		if templateId, ok := d.GetOk("without_template_settings.0.id"); ok {
 			payload.BlueprintId = templateId.(string)
 		}
 	} else {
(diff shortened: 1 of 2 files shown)


env0/resource_environment.go
```diff
@@ -732,6 +732,12 @@ func shouldUpdateTemplate(d *schema.ResourceData) bool {
 }
 
 func shouldDeploy(d *schema.ResourceData) bool {
+	if _, ok := d.GetOk("without_template_settings.0"); ok {
```
TomerHeber, line 735: added the required use-case.

env0/resource_environment.go
```diff
@@ -1126,8 +1132,7 @@ func getDeployPayload(d *schema.ResourceData, apiClient client.ApiClientInterfac
 	var err error
 
 	if isTemplateless(d) {
-		templateId, ok := d.GetOk("without_template_settings.0.id")
-		if ok {
+		if templateId, ok := d.GetOk("without_template_settings.0.id"); ok {
```
TomerHeber, line 1135: no change... just one less line.

env0/resource_environment.go
```diff
@@ -1139,6 +1144,10 @@ func getDeployPayload(d *schema.ResourceData, apiClient client.ApiClientInterfac
 	}
 
 	if isRedeploy {
+		if revision, ok := d.GetOk("without_template_settings.0.revision"); ok {
```
TomerHeber, line 1147: set the revision in the deploy request only if it's a redeploy.

env0/resource_environment_test.go
```diff
@@ -2541,6 +2541,13 @@ func TestUnitEnvironmentWithoutTemplateResource(t *testing.T) {
 
 				// Update
 				mock.EXPECT().TemplateUpdate(template.Id, templateUpdatePayload).Times(1).Return(updatedTemplate, nil),
+				mock.EXPECT().ConfigurationSetsAssignments("ENVIRONMENT", environment.Id).Times(1).Return(nil, nil),
+				mock.EXPECT().EnvironmentDeploy(environment.Id, client.DeployRequest{
```
TomerHeber, line 2545: Now deploy is called as part of the test.

env0/resource_environment.go
avnerenv0: this syntax is super strange ... do we use it a lot in this codebase / is it common in Go?

avnerenv0 (approved):
Thank you

env0/resource_environment.go
TomerHeber: Yes.
This is common go.

PR #920, opened 2024-07-31
Revert "Fix: Drift Notification for Variable Sets (due to ordering) in env0_e…"

RLRabinowitz:
Reverts env0/terraform-provider-env0#910

That PR has introduced a regression, breaking the configuration of any `env0_environment` that had `variable_sets` in it


RLRabinowitz:
This actually does not fix the bug. It seems the bug has existed ever since the addition of `variable_sets` to the resource

PR #925, opened 2024-08-05
Fix: Organization Policy can't be updated if max_ttl and default_ttl …

### LINKED ISSUE #924: Organization Policy can't be updated if max_ttl and default_ttl are nil
**Describe the bug**
A customer that had no `max_ttl` nor `default_ttl` set on his org policy resource tried to update his policy and faced an issue with the API. Upon inspection it seems that we are sending the value as `""` (empty string) and not null, same for the state itself. 
the API accepts null but empty sting is not good for us.

**To Reproduce**
create an org policy resource 
```
resource "env0_organization_policy" "policy_example" {
  do_not_consider_merge_commits_for_pr_plans = true
  enable_oidc                                = true
  enforce_pr_commenter_permissions           = true
  do_not_report_skipped_status_checks        = true

}
```
update whatever field you like

**Expected behavior**
the API should be sent with nil and not empty string

**Provider Version**
latest

**Screenshots**
image

**Additional context**
The solution we want to go for is sanitize the payload before sending to the api, something like 
```
	if payload.DefaultTtl != nil && *payload.DefaultTtl == "" {
		payload.DefaultTtl = nil
	}
	if payload.MaxTtl != nil && *payload.MaxTtl == "" {
		payload.MaxTtl = nil
	}
```
this is so that customers with broken state would be unblocked as well.

TomerHeber:
…are nil

fixes #924 

### Solution

When applying the API call: defaultTtl or maxTtl are an empty string, convert them to nil.

### DIFF
--- a/client/organization.go
+++ b/client/organization.go
@@ -72,6 +72,14 @@ func (client *ApiClient) OrganizationPolicyUpdate(payload OrganizationPolicyUpda
 		return nil, err
 	}
 
+	if payload.DefaultTtl != nil && *payload.DefaultTtl == "" {
+		payload.DefaultTtl = nil
+	}
+
+	if payload.MaxTtl != nil && *payload.MaxTtl == "" {
+		payload.MaxTtl = nil
+	}
+
 	var result Organization
 	if err := client.http.Post("/organizations/"+id+"/policies", payload, &result); err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


tomer-landesman:
can you add a UT for that?

tomer-landesman (approved):
Looks good to me.

PR #927, opened 2024-08-05
Fix: removing agent_project_assignment does not auto re-assign to the default agent

### LINKED ISSUE #914: removing `agent_project_assignment` does not auto re-assign to the default agent
**Describe the bug**
when using `agent_project_assignment` to assign a project to a specific agent, and then removing the `agent_project_assignment` resource, the project is not properly re-assigned to the default agent.

**To Reproduce**
* have org with at least 2 agents.
* use `agent_project_assignment` to assign project to 2nd agent (non-default)
* remove `agent_project_assignment`
* see project assignment - still stuck on 2nd agent.

**Expected behavior**
* after removing `agent_project_assignment` project should use default agent.

**Provider Version**
1.19.7

**Screenshots**
CleanShot 2024-07-30 at 09 16 59

**Additional context**

TomerHeber:
fixes #914 

### Solution

The previous implementation was incorrect. The implementation was re-written.
1. Updated the API calls.
2. Updated the unit tests for the API calls.
3. Rewritten the resource implementation.
4. Rewritten the acceptance tests for the resource.
5. Updated the import example to use "project_id".


client/agent_project_assignment.go
```diff
@@ -2,11 +2,11 @@ package client
 
 // Key is project id.
 // Value is agent id.
-type AssignProjectsAgentsAssignmentsPayload map[string]interface{}
+type AssignProjectsAgentsAssignmentsPayload map[string]string
 
 type ProjectsAgentsAssignments struct {
 	DefaultAgent   string                 `json:"defaultAgent"`
-	ProjectsAgents map[string]interface{} `json:"ProjectsAgents"`
+	ProjectsAgents map[string]interface{} `json:"projectsAgents"`
```
RLRabinowitz, line 9: nit

Could be changed to `map[string]interface{}` as well, right?
```suggestion
	ProjectsAgents map[string]string `json:"projectsAgents"`
```

env0/resource_agent_project_assignment.go
```diff
@@ -57,138 +43,99 @@ func resourceAgentProjectAssignment() *schema.Resource {
...
+	}
 
-	if _, err := apiClient.AssignAgentsToProjects(assignments.ProjectsAgents); err != nil {
-		return diag.Errorf("could not update project-agent assignments: %v", err)
+	if _, err := apiClient.AssignAgentsToProjects(payload); err != nil {
+		return diag.Errorf("failed to assign project '%s' to agent '%s': %v", assignment.ProjectId, assignment.AgentId, err)
 	}
 
-	d.SetId(newAssignment.GetId())
+	d.SetId(assignment.ProjectId)
```
RLRabinowitz, line 62: Does it matter much if the ID is the project ID or assignment ID?

env0/resource_agent_project_assignment.go
```diff
@@ -57,138 +43,99 @@ func resourceAgentProjectAssignment() *schema.Resource {
...
-	}
 
-	for projectId, agentId := range assignments.ProjectsAgents {
-		if projectId == newAssignment.ProjectId && agentId == newAssignment.AgentId {
-			return diag.Errorf("assignment for project id %v and agent id %v already exist", projectId, agentId)
-		}
+	var assignment AgentProjectAssignment
+	if err := readResourceData(&assignment, d); err != nil {
+		return diag.Errorf("schema resource data deserialization failed: %v", err)
 	}
```
RLRabinowitz, line 52: We might still want this validation that was removed, but only making sure that there is no assignment, or that the assignment is to the default agent.

Removing this could mean that someone using a different Terraform/OpenTofu stack can attempt to manage the same resource there, and cause conflicts

env0/resource_agent_project_assignment.go
```diff
@@ -57,138 +43,99 @@ func resourceAgentProjectAssignment() *schema.Resource {
...
-		return nil
+	for projectId, agent := range assignments.ProjectsAgents {
+		if projectId == assignment.ProjectId {
+			assignment.AgentId = agent.(string)
+			break
+		}
 	}
 
-	if err := writeResourceData(assignment, d); err != nil {
+	if err := writeResourceData(&assignment, d); err != nil {
```
RLRabinowitz, line 89: Consider extracting this logic, which is shared with another action

env0/resource_agent_project_assignment_test.go
```diff
@@ -11,219 +11,215 @@ import (
...
-						"agent_id":   agentId,
-					}),
-					Check: resource.ComposeAggregateTestCheckFunc(
-						resource.TestCheckResourceAttr(accessor, "id", agentId+"_"+projectId),
-						resource.TestCheckResourceAttr(accessor, "project_id", projectId),
-						resource.TestCheckResourceAttr(accessor, "agent_id", agentId),
-					),
-					PlanOnly:           true,
-					ExpectNonEmptyPlan: true,
+					Config: getConfig(&defaultAssignment),
```
RLRabinowitz, line 109: I don't think that `ENV0_DEFAULT` is an actual agentId that is returned from API ever. I think it's just a value that is accepted when updating assignments, to indicate that this is the default agent

client/agent_project_assignment.go
TomerHeber: or "string" if it's always a string...

env0/resource_agent_project_assignment.go
TomerHeber: I don't think an assignment has an id.
So AFAIK it has to be the project_id (assuming one assignment per project).

env0/resource_agent_project_assignment_test.go
TomerHeber: Yeah I got that. that's what I did. Or am I misunderstanding?

env0/resource_agent_project_assignment.go
TomerHeber: need to be careful with over re-using code.
Golang (unlike NodeJS) is a strongly typed language, readability is more important.

Ill have another look, but if I decided not re-use code, it was probably for a good reason.

env0/resource_agent_project_assignment.go
TomerHeber: This is really what Terraform is for.. it will report a drift.
In any case, maintaining two states in Terraform is never good practice.

env0/resource_agent_project_assignment.go
TomerHeber: I could do a small reuse here... I'll push something in a couple of minutes.

TomerHeber: Made the required changes. Thanks for pointing it out!

env0/resource_agent_project_assignment.go
RLRabinowitz: Oh OK. So this error wasn't really necessary before, IIUC

env0/resource_agent_project_assignment.go
RLRabinowitz: Up until now this was `GetId`, which is a combo of agentKey and project ID. And now it has changed. Wouldn't this affect existing configurations, causing drift or other issues?

client/agent_project_assignment.go
RLRabinowitz: Yeah, sorry, that's what I meant

env0/resource_agent_project_assignment_test.go
RLRabinowitz: Not sure why I commented that here 😅

env0/resource_agent_project_assignment.go
RLRabinowitz: I guess ProjectId makes more sense here as the ID. What would be the implications of this change on existing configurations?

TomerHeber: yes you are correct.

The feature did not work as expected earlier, so fixing it now may cause a drift.
I think the best thing we can do is just add a note in the release notes about this breaking change. (That's what we've done in the past).

RLRabinowitz: Alright. Could you please test the behaviour locally so we'd know exactly what users might see once they upgrade?

RLRabinowitz (approved):
As discussed, this is a breaking change for anyone with an existing state managing `agent_project_assignment`s. Anyone planning after upgrading to a new version containing this fix would get a drift, that would unassign and reassign the agent to the project.

The diff itself will look odd, sayin that the `agentKey` and `projectId` will be changed, and the "from" part of the change will be wrong.

We should make sure to mention that in the release notes, when detailing the breaking change

PR #939, opened 2024-08-22
Feat: add support for 'move environment' backend

TomerHeber:
### Issue & Steps to Reproduce / Feature Request
resolves #938 

### Solution

1. Added a new API call for moving environments.
2. Added a unit test for the new API call.
3. Removed 'forceNew' from 'project_id' in the environment schema.
4. On environment update, if project_id is updated, call the move API.
5. Added an acceptance tests.
6. Updated harness tests.

PR #943, opened 2024-08-26
Chore: move env test is flaky - use never deployed env

chpl:
the test is flaky - often fails with 
```
│ Error: failed to move environment to project id 'e46ce3e9-79b5-4e4d-bba9-1d9b9f252155': 400 Bad Request: cannot move environment while in progress. id: 4cc847e7-c45d-4389-b49d-4e9bdf94f80a
```
(was added in https://github.com/env0/terraform-provider-env0/pull/939)

### Solution
use never deployed environment

### DIFF
--- a/tests/integration/012_environment/main.tf
+++ b/tests/integration/012_environment/main.tf
@@ -40,7 +40,7 @@ resource "env0_template_project_assignment" "assignment2" {
 }
 
 resource "env0_environment" "auto_glob_envrironment" {
-  depends_on                       = [env0_template_project_assignment.assignment]
+  depends_on = [env0_template_project_assignment.assignment]
   name                             = "environment-auto-glob-${random_string.random.result}"
   project_id                       = env0_project.test_project.id
   template_id                      = env0_template.template.id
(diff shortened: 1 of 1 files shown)


tests/integration/012_environment/main.tf
```diff
@@ -55,7 +55,7 @@ resource "env0_environment" "example" {
   depends_on    = [env0_template_project_assignment.assignment]
```
yarivg, line 55: Is this environment still necessary?

yarivg (approved):
seems good

tests/integration/012_environment/main.tf
chpl: yes. tomer just added the move part [here](https://github.com/env0/terraform-provider-env0/pull/939)

tomer-landesman: I think the original purpose of the test was to test different types / settings of environments, and here there was an attempt to kill 2 birds with one stone by testing both regular deployment and move environment. so I think it is still needed for the original purpose

PR #952, opened 2024-09-06
Feat: add "parent_project_path" or "hierarchy_by_name" to data "env0_…

### LINKED ISSUE #948: add "parent_project_path" or "hierarchy_by_name" to `data "env0_project"`

Given this project structure:
App -> Dev -> us-east-1 -> service
    -> Stage -> us-east-1 -> service 
App2 -> Dev -> service  

It is very difficult to get the "right" `service` project. 
And with the current `data "env0_project"` resource, I have to implement a relatively complicated data source to find the "right" `service` project id.

A very naive implementation:

data "env0_project" "app" {
  name = "App"
}

data "env0_project" "dev" {
  name = "Dev"
  parent_project_id = data.env0_project.app
}

... and so forth.

The challenge is that we do not always know how many "parent" levels there, and thus making this repeatable and "programmatic" is difficult.

It would be nice to specify an entire path as a project filter.

```
data "env0_project" "service1" {
  name = "service"
  parent_project_path = "App|Dev|us-east-1"
}

data "env0_project" "service3" {
  name = "service"
  parent_project_path = "App2|Dev"
}
```

I could not figure out how to write a dynamic data structure that could depend on N number of parent projects.  

**Additional context**
And if there could be a filter using `hierarchy` ("d02e5d45-7e02-41ed-a999-bc51a2dabb8e|80d66fc8-0157-46d8-8a5c-5abdac2e98b8") to consume as well

TomerHeber:
…project"

resolves #948 

### Solution
1. Updated the project schema and added the new value.
2. Add required code to search based on path.
3. Added acceptance tests.
4. Added harness tests.

### DIFF
--- a/env0/data_project.go
+++ b/env0/data_project.go
@@ -30,9 +30,16 @@ func dataProject() *schema.Resource {
 				Computed:     true,
 			},
 			"parent_project_name": {
-				Type:        schema.TypeString,
-				Description: "the name of the parent project. Can be used as a filter when there are multiple subprojects with the same name under different parent projects",
-				Optional:    true,
+				Type:          schema.TypeString,
+				Description:   "the name of the parent project. Can be used as a filter when there are multiple subprojects with the same name under different parent projects",
+				Optional:      true,
+				ConflictsWith: []string{"parent_project_path"},
+			},
+			"parent_project_path": {
+				Type:          schema.TypeString,
+				Description:   "a path of ancestors projects divided by the prefix '|'. Can be used as a filter when there are multiple subprojects with the same name under different parent projects. For example: 'App|Dev|us-east-1' will search for a project with the hierarchy 'App -> Dev -> us-east-1' ('us-east-1' being the parent)",
+				Optional:      true,
+				ConflictsWith: []string{"parent_project_name"},
 			},
 			"parent_project_id": {
 				Type:        schema.TypeString,
@@ -79,7 +86,8 @@ func dataProjectRead(ctx context.Context, d *schema.ResourceData, meta interface
 		if !ok {
 			return diag.Errorf("either 'name' or 'id' must be specified")
 		}
-		project, err = getProjectByName(name.(string), d.Get("parent_project_id").(string), d.Get("parent_project_name").(string), meta)
+
+		project, err = getProjectByName(name.(string), d.Get("parent_project_id").(string), d.Get("parent_project_name").(string), d.Get("parent_project_path").(string), meta)
 		if err != nil {
 			return diag.Errorf("%v", err)
 		}
@@ -94,6 +102,7 @@ func dataProjectRead(ctx context.Context, d *schema.ResourceData, meta interface
 
 func filterByParentProjectId(parentId string, projects []client.Project) []client.Project {
 	filteredProjects := make([]client.Project, 0)
+
 	for _, project := range projects {
 		if len(project.ParentProjectId) == 0 {
 			continue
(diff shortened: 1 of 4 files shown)


env0/data_project.go
```diff
@@ -30,9 +31,16 @@ func dataProject() *schema.Resource {
...
+				Type:          schema.TypeString,
+				Description:   "the name of the parent project. Can be used as a filter when there are multiple subprojects with the same name under different parent projects",
+				Optional:      true,
+				ConflictsWith: []string{"parent_project_path"},
+			},
+			"parent_project_path": {
+				Type:          schema.TypeString,
+				Description:   "a path of ancestors projects divided by the prefix '|'. Can be used as a filter when there are multiple subprojects with the same name under different parent projects. For example: 'App|Dev|us-east-1' will search for a project with the hierarchy 'App -> Dev -> us-east-1' ('us-east-1' being the parent)",
+				Optional:      true,
+				ConflictsWith: []string{"parent_project_name"},
```
ItamarMalka, line 43: does it not also conflict with the parent_project_id?

env0/data_project.go
```diff
@@ -162,14 +182,64 @@ func getProjectByName(name string, parentId string, parentName string, meta inte
...
+	filteredProjects := make([]client.Project, 0)
+
+	path := strings.Split(parentPath, "|")
+
+	for _, project := range projectsByName {
+		parentIds := strings.Split(project.Hierarchy, "|")
+		// right most element is the project itself, remove it.
+		parentIds = parentIds[:len(parentIds)-1]
+
+		matches, err := pathMatches(path, parentIds, meta)
```
ItamarMalka, line 218: mmmm that feels a bit complicated, couldn't we have filtered the projects by their hierarchy seeing if `hierarchy.contains(searchedPath)` ?
Or am I missing something?

env0/data_project.go
```diff
@@ -162,14 +182,64 @@ func getProjectByName(name string, parentId string, parentName string, meta inte
 	return projectsByName[0], nil
 }
 
+func pathMatches(path, parentIds []string, meta interface{}) (bool, error) {
+	if len(path) != len(parentIds) {
```
ItamarMalka, line 186: so if we have a1|a2|a3|searchedProject
and we put a1|a2 as the path we would not find it, is that intended?

env0/data_project.go
ItamarMalka: oh oh nvm I am wrong since those are IDs not names so it won't work that way 👌

ItamarMalka (commented):
Looks good, just ping me when you answered the questions I left 👍

env0/data_project.go
TomerHeber: you're correct...
I think I did not do that because the parent_project_id is also a computed field, and I was worried it will cause issues.
But looks like the SDK knows how to handle that use-case, so I'll push the change. (Sometimes the Terraform SDK is buggy).

env0/data_project.go
TomerHeber: I tend to agree. We could support prefixes as well.
I don't know if that's what @away168 had in mind... but why not...

modifying to:
```
	if len(path) > len(parentIds) {
		return false, nil
	}
```

TomerHeber:
> Looks good, just ping me when you answered the questions I left 👍

@ItamarMalka thank you.
I have updated the PR. For the prefix match, I've added a couple of new tests.
All changes are in 40dcdef3a92ce46374dc5864a16145f96342b4d8

ItamarMalka (approved):
LGTM 🚀 👍 
@away168 let us know if there is anything else that you think is needed here 🙏

PR #954, opened 2024-09-11
Fix: approve_plan_automatically attribute fails for sub_environment

### LINKED ISSUE #947: approve_plan_automatically attribute fails for sub_environment
**Describe the bug**
When adding `approve_plan_automatically` attribute to a sub_environment, using either `true` or `false` values, a failure occurs with the following error:

```
│ Error: could not create environment: 400 Bad Request: approval flow can't be enabled on workflow environments
│ 
│   with env0_environment.example_with_hcl_configuration,
│   on main.tf line 15, in resource "env0_environment" "example_with_hcl_configuration":
│   15: resource "env0_environment" "example_with_hcl_configuration" {
│ 
```

**To Reproduce**
1. Create an `env0_environment` resource with `sub_environment_configuration`
2. Add `approve_plan_automatically` attribute to one of the sub environments
3. Run apply

```
terraform {
  required_providers {
    env0 = {
      source = "env0/env0"
    }
  }
}

provider "env0" {
  api_key    = var.env0_api_key
  api_secret = var.env0_api_secret
}

resource "env0_environment" "example_with_hcl_configuration" {
  name        = "test workflow approval policy"
  project_id  = var.project_id
  template_id = var.template_id

  is_remote_backend = "true"


    alias    = "vpc"
    revision = "main"
    approve_plan_automatically = "true"
  }

    alias    = "eks"
    revision = "main"
    approve_plan_automatically = "true"
  }
}
```

**Expected behavior**
Automatic approval plan should be enabled/disabled on the sub environment, as per the documentation:
[https://registry.terraform.io/providers/env0/env0/latest/docs/resources/environment#nested-schema-for-sub_environment_configuration](https://registry.terraform.io/providers/env0/env0/latest/docs/resources/environment#nested-schema-for-sub_environment_configuration)

**Provider Version**
latest (1.20.6) and 1.19.3

**Screenshots**
CleanShot 2024-09-03 at 17 33 48@2x

TomerHeber:
fixes #947 

### Solution

1. For workflow - always pass `RequiredUserApproval` as `false` in the deploy payload (for the root env).
2. Updated acceptance tests.
3. Updated harness tests.

### DIFF
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -1179,6 +1179,11 @@ func getDeployPayload(d *schema.ResourceData, apiClient client.ApiClientInterfac
 		payload.BlueprintRevision = revision.(string)
 	}
 
+	// For 'Workflows', the 'root' environment should never require a user approval.
+	if _, ok := d.GetOk("sub_environment_configuration"); ok {
+		payload.UserRequiresApproval = boolPtr(false)
+	}
+
 	if isRedeploy {
 		if revision, ok := d.GetOk("without_template_settings.0.revision"); ok {
 			payload.BlueprintRevision = revision.(string)
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -976,9 +976,20 @@ func getCreatePayload(d *schema.ResourceData, apiClient client.ApiClientInterfac
 		payload.AutoDeployOnPathChangesOnly = boolPtr(val.(bool))
 	}
 
+	_, isWorkflow := d.GetOk("sub_environment_configuration")
+
 	//lint:ignore SA1019 reason: https://github.com/hashicorp/terraform-plugin-sdk/issues/817
 	if val, exists := d.GetOkExists("approve_plan_automatically"); exists {
 		payload.RequiresApproval = boolPtr(!val.(bool))
+
+		if isWorkflow && *payload.RequiresApproval {
+			return client.EnvironmentCreate{}, diag.Errorf("approve_plan_automatically cannot be 'false' for workflows")
+		}
+	}
+
+	// For 'Workflows', the 'root' environment should never require an approval.
+	if _, ok := d.GetOk("sub_environment_configuration"); ok {
+		payload.RequiresApproval = boolPtr(false)
 	}
 
 	//lint:ignore SA1019 reason: https://github.com/hashicorp/terraform-plugin-sdk/issues/817
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -989,7 +989,7 @@ func getCreatePayload(d *schema.ResourceData, apiClient client.ApiClientInterfac
 	}
 
 	// For 'Workflows', the 'root' environment should never require an approval.
-	if _, ok := d.GetOk("sub_environment_configuration"); ok {
+	if isWorkflow {
 		payload.RequiresApproval = boolPtr(false)
 	}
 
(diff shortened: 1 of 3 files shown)


env0/resource_environment.go
```diff
@@ -1179,6 +1190,11 @@ func getDeployPayload(d *schema.ResourceData, apiClient client.ApiClientInterfac
 		payload.BlueprintRevision = revision.(string)
 	}
 
+	// For 'Workflows', the 'root' environment should never require a user approval.
+	if _, ok := d.GetOk("sub_environment_configuration"); ok {
```
TomerHeber, line 1194: this shouldn't be needed...
This might be a bug in the backend...

env0/resource_environment.go
```diff
@@ -415,7 +415,8 @@ func setEnvironmentSchema(ctx context.Context, d *schema.ResourceData, environme
 		d.Set("output", string(environment.LatestDeploymentLog.Output))
 	}
 
-	if environment.RequiresApproval != nil {
+	// Don't update this value for workflow environments - this value should always be 'false'.
+	if _, isWorkflow := d.GetOk("sub_environment_configuration"); !isWorkflow && environment.RequiresApproval != nil {
```
TomerHeber, line 419: don't modify the schema for workflow environment...
generally this would not be required... however due to the change in line 1195 (which I believe is a backend issue) - I had to add this here. 

(see my comment in line 1195).

env0/resource_environment.go
TomerHeber: it seems that the backend will error if this is 'undefined'.

weinguy-env0: this could be overridden by a project policy, this is a good extra step anyways.

env0/resource_environment.go
```diff
@@ -976,9 +977,20 @@ func getCreatePayload(d *schema.ResourceData, apiClient client.ApiClientInterfac
...
 		payload.RequiresApproval = boolPtr(!val.(bool))
+
+		if isWorkflow && *payload.RequiresApproval {
+			return client.EnvironmentCreate{}, diag.Errorf("approve_plan_automatically cannot be 'false' for workflows")
+		}
+	}
+
+	// For 'Workflows', the 'root' environment should never require an approval.
+	if _, ok := d.GetOk("sub_environment_configuration"); ok {
+		payload.RequiresApproval = boolPtr(false)
```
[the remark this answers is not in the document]
TomerHeber: line 986 - is the use-case when someone excplicitly set 'approve_plan_automatically' to 'false' and it's a workflow - in that case we can quickly error out as bad schema set by the user.

line 992 - is the use-case where it's a workflow - the value is forced to be 'false'.

env0/resource_environment.go
```diff
@@ -976,9 +977,20 @@ func getCreatePayload(d *schema.ResourceData, apiClient client.ApiClientInterfac
...
 	if val, exists := d.GetOkExists("approve_plan_automatically"); exists {
 		payload.RequiresApproval = boolPtr(!val.(bool))
+
+		if isWorkflow && *payload.RequiresApproval {
+			return client.EnvironmentCreate{}, diag.Errorf("approve_plan_automatically cannot be 'false' for workflows")
+		}
+	}
+
+	// For 'Workflows', the 'root' environment should never require an approval.
+	if _, ok := d.GetOk("sub_environment_configuration"); ok {
```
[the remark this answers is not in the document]
TomerHeber: thanks! fixed!

env0/resource_environment_test.go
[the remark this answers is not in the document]
TomerHeber: added b1d22bb0e1b05c6cf2cf089bddf808fa0a6d72fd

weinguy-env0 (approved):

Issue #894, opened 2024-07-10
bug - configuration variable that overrides a template variable in env0_environment shows drift


away168:
**Describe the bug**
A variable defined in the env0_template is being overwritten | specified in env0_environment that uses the template. 

the plan shows drift of parameters that are not meant to be re-configured (e.g. `regex`, `is_required`, `description`

```
  ~ resource "env0_environment" "default_projects" {
        id                               = "f29bfdca-1a77-49dd-b197-f2f0c4913a99"
        name                             = "Skunkworks Project"
      ~ variable_sets                    = [
          - "69cfcac0-5135-4967-a9d5-c6166f216f4c",
        ]
        # (13 unchanged attributes hidden)

      ~ configuration {
          - description  = "main project name for team" -> null
          ~ is_required  = true -> false
            name         = "team_name"
          - regex        = "[a-zA-Z0-9-_]*" -> null
            # (6 unchanged attributes hidden)
        }
    }
```

**To Reproduce**
1. Create an env0_template and env0_configuration_variable attached to template.
2. Create an env0_environment that is deployed from template with a `configuration` variable that overrides the template variable.

**Expected behavior**
the env0_environment should not show drift.

**Provider Version**
v1.19.1

**Screenshots**

**Additional context**
Sample code:

```
resource "env0_template" "projects" {
  name                   = "Team Projects"
  type                   = "opentofu"
  description            = "Project Factory for onboarding new teams"
  opentofu_version       = "latest"
  repository             = data.env0_template.this.repository
  path                   = "env0/projects"
  github_installation_id = data.env0_template.this.github_installation_id
}

resource "env0_template_project_assignment" "projects" {
  template_id = env0_template.projects.id
  project_id  = data.env0_environment.this.project_id
}

resource "env0_configuration_variable" "team_name" {
  template_id = env0_template.projects.id
  name        = "team_name"
  description = "main project name for team"
  is_required = true
  regex       = "[a-zA-Z0-9-_]*"
  type        = "terraform"
}

resource "env0_environment" "default_projects" {
  count = var.create_projects ? 1 : 0

  name                       = "${var.default_team_name} Project"
  project_id                 = data.env0_environment.this.project_id
  template_id                = env0_template.projects.id
  approve_plan_automatically = true
  is_remote_backend          = true
  removal_strategy           = "mark_as_archived"
  workspace                  = "${var.default_team_name}_project"

  configuration {
    name        = "team_name"
    value       = var.default_team_name
[...]

PR #955, opened 2024-09-20
Fix: configuration variable that overrides a template variable in env…

TomerHeber:
…0_environment shows drift

Part of #894

### Solution

Since the change is large and complex, this PR is the first part of fixing the issue.
For now I've just added the Overwrites field to the struct.

### DIFF
--- a/client/configuration_variable.go
+++ b/client/configuration_variable.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"encoding/json"
 	"errors"
 )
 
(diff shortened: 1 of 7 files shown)


client/configuration_variable.go
```diff
@@ -39,22 +40,30 @@ func (c *ConfigurationVariableSchema) ResourceDataSliceStructValueWrite(values m
...
+	Id             string                           `json:"id,omitempty"`
+	Name           string                           `json:"name"`
+	Description    string                           `json:"description,omitempty"`
+	Type           *ConfigurationVariableType       `json:"type,omitempty" tfschema:",omitempty"`
+	Schema         *ConfigurationVariableSchema     `json:"schema,omitempty"`
+	ToDelete       *bool                            `json:"toDelete,omitempty"`
+	IsReadOnly     *bool                            `json:"isReadonly,omitempty"`
+	IsRequired     *bool                            `json:"isRequired,omitempty"`
+	Regex          string                           `json:"regex,omitempty"`
+	Overwrites     *ConfigurationVariableOverwrites `json:"overwrites,omitempty"` // Is removed when marhseling to a JSON.
```
TomerHeber, line 66: This struct is also sent in a POST requests.
Overwrites field should **not** be sent in a request. (We are only intrested of this value in a response).
To solve this issue without making too many code changes, I have decided to write a custom marshal.

client/configuration_variable.go
```diff
@@ -77,6 +86,17 @@ type ConfigurationVariableUpdateParams struct {
 	Id           string
 }
 
+func (v ConfigurationVariable) MarshalJSON() ([]byte, error) {
+	v.Overwrites = nil
+
+	// This is done to prevent an infinite loop.
+	type ConfigurationVariableDummy ConfigurationVariable
```
TomerHeber, line 93: without doing the following an infinite loop will happen, because inside marshal we are calling marshal again on the same object.

client/configuration_variable.go
```diff
@@ -77,6 +86,17 @@ type ConfigurationVariableUpdateParams struct {
 	Id           string
 }
 
+func (v ConfigurationVariable) MarshalJSON() ([]byte, error) {
+	v.Overwrites = nil
```
TomerHeber, line 90: avoid sending Overwrites when marshaling a json to bytes.

client/configuration_variable.go
```diff
@@ -77,6 +86,17 @@ type ConfigurationVariableUpdateParams struct {
...
+
+	// This is done to prevent an infinite loop.
+	type ConfigurationVariableDummy ConfigurationVariable
+
+	dummy := ConfigurationVariableDummy(v)
+
+	return json.Marshal(&dummy)
```
TomerHeber, line 97: this would have caused loops...

client/configuration_variable_test.go
```diff
@@ -291,3 +295,37 @@ var _ = Describe("Configuration Variable", func() {
 		})
 	})
 })
+
+func TestConfigurationVariableMarshelling(t *testing.T) {
```
TomerHeber, line 299: this test was added to verify that the marshal and unmarshal are working as expected.

env0/configuration_variable.go
```diff
@@ -0,0 +1,143 @@
+package env0
```
TomerHeber, line 1: moved these functions to here from env0/resource_environment.go
(These are mostly commom functions).

client/configuration_variable_test.go
```diff
@@ -187,9 +190,10 @@ var _ = Describe("Configuration Variable", func() {
 	})
 
 	Describe("ConfigurationVariableDelete", func() {
+
 		BeforeEach(func() {
 			httpCall = mockHttpClient.EXPECT().Delete("configuration/"+mockConfigurationVariable.Id, nil)
-			apiClient.ConfigurationVariableDelete(mockConfigurationVariable.Id)
+			_ = apiClient.ConfigurationVariableDelete(mockConfigurationVariable.Id)
```
sagilaufer1992, line 196: why these changes?

sagilaufer1992 (approved):

client/configuration_variable_test.go
TomerHeber: using a new linter - fixes a linting issue.
It tells the linter, I know that this can return an error - but I want to ignore it.

PR #965, opened 2024-10-21
Fix: vcs_command_alias should be set to "" when vcs_pr_comments_enabl…

### LINKED ISSUE #964: `vcs_command_alias` should be set to `""` when `vcs_pr_comments_enabled` is disabled
**Describe the bug**
In a way while this could be a backend bug, the RnD team has asked to file this with the Provider.

When [`vcs_pr_comments_enabled`](https://registry.terraform.io/providers/env0/env0/latest/docs/resources/environment#vcs_pr_comments_enabled) is disabled; and the vcs_command_alias is still set - the environment service still thinks that the vcs_pr_comments is enabled and configured. Therefore it generates a "summary" in the git PR

**To Reproduce**
Create an environment with the provider; set vcs_command_alias="foo" and vcs_pr_comments_enabled=true

Then set `vcs_pr_comments_enabled=false` 

Create a PR.
you'll see a pr_comment "header / summary" in the PR.

**Expected behavior**
PR Comment header or summary should not be created/added.

**Provider Version**
1.20.12

**Screenshots**
image

**Additional context**

TomerHeber:
…ed is disabled

fixes #964 

### Solution

Forces alias to be empty if vcs_pr_comments_enabled is unset or false.
Updated acceptance tests.

### DIFF
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -1062,6 +1062,11 @@ func getCreatePayload(d *schema.ResourceData, apiClient client.ApiClientInterfac
 		}
 	}
 
+	// Send an empty vcs_commands_alias if vcs_pr_comments_enabled is 'false' or unset - https://github.com/env0/terraform-provider-env0/issues/964
+	if payload.VcsPrCommentsEnabled == nil || !*payload.VcsPrCommentsEnabled {
+		payload.VcsCommandsAlias = ""
+	}
+
 	payload.DeployRequest = &deployPayload
 
 	return payload, nil
@@ -1127,6 +1132,11 @@ func getUpdatePayload(d *schema.ResourceData) (client.EnvironmentUpdate, diag.Di
 
 	if d.HasChange("is_inactive") {
 		payload.IsArchived = boolPtr(d.Get("is_inactive").(bool))
+	}
+
+	// Send an empty vcs_commands_alias if vcs_pr_comments_enabled is 'false' or unset - https://github.com/env0/terraform-provider-env0/issues/964
+	if payload.VcsPrCommentsEnabled == nil || !*payload.VcsPrCommentsEnabled {
+		payload.VcsCommandsAlias = ""
 	}
 
 	if err := assertEnvironment(d); err != nil {
--- a/env0/resource_environment_test.go
+++ b/env0/resource_environment_test.go
@@ -59,7 +59,7 @@ func TestUnitEnvironmentResource(t *testing.T) {
 		},
 		TerragruntWorkingDirectory: "/terragrunt/directory/",
 		VcsCommandsAlias:           "alias2",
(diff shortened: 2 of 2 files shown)


env0/resource_environment.go
```diff
@@ -1062,6 +1062,11 @@ func getCreatePayload(d *schema.ResourceData, apiClient client.ApiClientInterfac
 		}
 	}
 
+	// Send an empty vcs_commands_alias if vcs_pr_comments_enabled is 'false' or unset - https://github.com/env0/terraform-provider-env0/issues/964
+	if payload.VcsPrCommentsEnabled == nil || !*payload.VcsPrCommentsEnabled {
```
TomerHeber, line 1066: create use-case.

env0/resource_environment.go
```diff
@@ -1129,6 +1134,11 @@ func getUpdatePayload(d *schema.ResourceData) (client.EnvironmentUpdate, diag.Di
 		payload.IsArchived = boolPtr(d.Get("is_inactive").(bool))
 	}
 
+	// Send an empty vcs_commands_alias if vcs_pr_comments_enabled is 'false' or unset - https://github.com/env0/terraform-provider-env0/issues/964
+	if payload.VcsPrCommentsEnabled == nil || !*payload.VcsPrCommentsEnabled {
```
TomerHeber, line 1138: update use-case.

Yossi-kerner (commented):
I see that you had to add the `VcsPrCommentsEnabled` field set to `true`. Does that mean that users that don't have that prop and depends on having an alias or not having it to enable and disable pr comments will have issues no?

TomerHeber:
> I see that you had to add the `VcsPrCommentsEnabled` field set to `true`. Does that mean that users that don't have that prop and depends on having an alias or not having it to enable and disable pr comments will have issues no?

Hi @Yossi-kerner - valid question!
The "true" was set in tests, many of the tests were broken becuase of the change in the code. For the alias to be "set" I have to also set VcsPrCommentsEnabled to 'true'. So this was a minimum change to fix the tests (Which unfortunatly, took some time... because most tests got broken...).

To be honest, the fix could have been in the backend and not in the provider. This is an anti-pattern for the provider code. However, I was asked to make the fix in the provider. So that's what I did (:

Yossi-kerner (approved):
Sorry for the late approve. We decided we are ok with that change, just make sure to change the major version when we release because it's a breaking change and make it clear in the change log what should be done 🙏 Thanks

TomerHeber:
> Sorry for the late approve. We decided we are ok with that change, just make sure to change the major version when we release because it's a breaking change and make it clear in the change log what should be done 🙏 Thanks

Feel free to edit the description if it's unclear.
https://github.com/env0/terraform-provider-env0/releases/tag/v1.22.0

Thanks!

PR #980, opened 2024-11-16
Feat: Breaking Change - default terragrunt_tf_binary to 'opentofu'

### LINKED ISSUE #979: Breaking Change - default terragrunt_tf_binary to 'opentofu'
Up until now ' terragrunt_tf_binary' defaulted to "opentofu" strictly for new resources.
Moving forward we would like to dafault it for any resource - existing or new.

These means that terragrunt_tf_binary must now be explicitly set to 'terraform'.

TomerHeber:
resolves #979 

### Solution

1. For terragrunt_tf_binary - implicitly send "opentofu" when terragrunt is enabled and terragrunt_tf_binary has not been set.
2. In the template body http respone ignore the "opentofu" value in the field terragruntTfBinary if terragrunt_tf_binary has not been set - to avoid drifts.
3. Updated acceptance tests.

--- breaking change for terraform + terragrunt_tf_binary (must excplicitly state 'terraform' value).

Discussion thread:
https://env0.slack.com/archives/C01QBBEGBEH/p1731500435184849


client/template.go
```diff
@@ -98,7 +98,7 @@ type TemplateCreatePayload struct {
 	IsAzureDevOps        bool             `json:"isAzureDevOps" tfschema:"is_azure_devops"`
 	IsHelmRepository     bool             `json:"isHelmRepository"`
 	HelmChartName        string           `json:"helmChartName,omitempty"`
-	TerragruntTfBinary   string           `json:"terragruntTfBinary,omitempty"`
```
TomerHeber, line 101: in the schema default is "" - so it no longer matters if an empty response is returned.

client/template.go
```diff
@@ -131,18 +131,31 @@ func (payload *TemplateCreatePayload) Invalidate() error {
...
+		if payload.TerragruntVersion == "" {
+			return errors.New("must supply terragrunt version")
+		}
+
+		// The provider implicitly defaults to "opentofu".
+		if payload.TerragruntTfBinary == "" {
+			payload.TerragruntTfBinary = OPENTOFU
+		}
+
+		if payload.TerragruntTfBinary == OPENTOFU && payload.OpentofuVersion == "" {
```
TomerHeber, line 144: enforce opentofu or terraform versions (when applicible) in the provider side.
(This is a nice to have).

client/template.go
```diff
@@ -131,18 +131,31 @@ func (payload *TemplateCreatePayload) Invalidate() error {
...
+			return errors.New("must supply opentofu version")
+		}
+
+		if payload.TerragruntTfBinary == TERRAFORM && payload.TerraformVersion == "" {
+			return errors.New("must supply terraform version")
+		}
+	} else {
+		payload.TerragruntTfBinary = ""
```
TomerHeber, line 152: in any case - if it's not terragrunt always send an empty terragrunt_tf_binary/

env0/resource_template.go
```diff
@@ -410,25 +410,6 @@ func templateCreatePayloadFromParameters(prefix string, d *schema.ResourceData)
 		return payload, diag.Errorf("schema resource data serialization failed: %v", err)
 	}
 
-	isNew := d.IsNewResource()
```
TomerHeber, line 413: this was the code to handle only for new resources... this is the breaking change...

client/template.go
```diff
@@ -131,18 +131,31 @@ func (payload *TemplateCreatePayload) Invalidate() error {
...
 	}
 
 	if payload.Type == OPENTOFU && payload.OpentofuVersion == "" {
 		return errors.New("must supply opentofu version")
 	}
 
-	if payload.TerragruntTfBinary != "" && payload.Type != TERRAGRUNT {
```
TomerHeber, line 142: not needed anymore.
We just ignore the value by doing:
`payload.TerragruntTfBinary = ""` (line 152 in the new code).

env0/resource_template_test.go
```diff
@@ -1376,25 +1389,6 @@ func TestUnitTemplateResource(t *testing.T) {
 		runUnitTest(t, testCase, func(mock *client.MockApiClientInterface) {})
 	})
 
-	t.Run("terragrunt_tf_binary set with a non terragrunt template type", func(t *testing.T) {
```
TomerHeber, line 1379: relevant code was removed. Therefore, I removed the test.

env0/resource_template_test.go
```diff
@@ -1376,18 +1389,38 @@ func TestUnitTemplateResource(t *testing.T) {
...
+					ExpectError: regexp.MustCompile("must supply opentofu version"),
+				},
+			},
+		}
+
+		runUnitTest(t, testCase, func(mock *client.MockApiClientInterface) {})
+	})
+
+	t.Run("run terragrunt with terraform binary and no terraform version", func(t *testing.T) {
 		testCase := resource.TestCase{
```
TomerHeber, line 1412: removed this test... relevant code was removed.

client/template.go
razbensimon: are you sure? i think it is better to be explicit and tell the user to delete it

TomerHeber: sure I don't mind either way.

env0/resource_environment.go
```diff
@@ -1443,7 +1443,7 @@ func resourceEnvironmentImport(ctx context.Context, d *schema.ResourceData, meta
 			return nil, fmt.Errorf("failed to get template with id %s: %w", templateId, err)
 		}
 
-		if err := templateRead("without_template_settings", template, d, true); err != nil {
+		if err := templateRead("without_template_settings", template, d); err != nil {
```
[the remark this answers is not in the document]
TomerHeber: There was special handling related to import and TerragruntTfBinary - it's no longer needed since we will implicitly default to "opentofu".

If there is a value returned from the http response (when askign for the template) it will be assigned to the schema.

In other words, the code that was handling a special use case related to terraform import and TerragruntTfBinary was removed. So it's no longer required to pass a true/false flag.

env0/resource_template.go
```diff
@@ -440,33 +421,24 @@ func templateCreatePayloadFromParameters(prefix string, d *schema.ResourceData)
...
-	// If this value isn't set, ignore whatever is returned from the response.
-	// This helps avoid drifts when defaulting to 'opentofu' for new 'terragrunt' templates, and 'terraform' for existing 'terragrunt' templates.
-	// 'template.TerragruntTfBinary' field is set to 'omitempty'. Therefore, the state isn't modified if `template.TerragruntTfBinary` is an empty string.
-	// This is not true for imports - because the shcema is empty irrespective in that case.
-	if !isImport {
-		if terragruntTfBinary == "" {
+
+	// This is done to avoid drifts in case the backend returns "opentofu", but non is configured in the provider.
+	// (The provider implicitly defaults to "opentofu").
+	if template.TerragruntTfBinary == client.OPENTOFU {
```
razbensimon, line 437: 1. But this PR will do a breaking change on purpose, so why do we need that? 
2. if BE returns "opentofu" and non is configured in the provider, it means this template was "terraform" before. So we should actually set "terraform", or actually show the drift. Am I missing something? i

client/template.go
razbensimon: so lets keep it, with test

env0/resource_template.go
TomerHeber: If the BE returns opentofu - then it's opentofu.
The backend is the source of truth.

However, if nothing is set in the provider - then - the defaul is "opentofu".
For various hashicorp reasons, I can't set default as "opentofu" in the terraform schema. But still want avoid the drift for the "opentofu" case. `'' -> 'opentofu'`.

If in the backend the value is "terraform" - and nothing was set in the provider - the user will get a drift message `terraform -> null`. To avoid the drift they will have to explicitly add `terragrunt_tf_binary = 'terraform'`

Now this is assuming that the backend (source of truth) will always return the correct value.

client/template.go
TomerHeber: I actually already updated it. See latest commit.

env0/resource_template.go
TomerHeber: If the BE returns opentofu - then it's opentofu.
The backend is the source of truth.

However, if nothing is set in the provider - then - the defaul is "opentofu".
For various hashicorp reasons, I can't set default as "opentofu" in the terraform schema. But still want avoid the drift for the "opentofu" case. `'' -> 'opentofu'`.

If in the backend the value is "terraform" - and nothing was set in the provider - the user will get a drift message `terraform -> null`. To avoid the drift they will have to explicitly add `terragrunt_tf_binary = 'terraform'`

Now this is assuming that the backend (source of truth) will always return the correct value.

razbensimon: > terraform -> null

isn't it `'' -> 'terraform'`? I didnt get this part

and thanks for the explanation

client/template.go
razbensimon: [6abd865](https://github.com/env0/terraform-provider-env0/pull/980/commits/6abd865082dad28d3a9787dd791a3a9e4747cbfb)

razbensimon (approved):
See the last question, let's resolve that before the merge
Overall looks good to me 👍🏼 

Let's do it in a new minor version, and doc it.

env0/resource_template.go
TomerHeber: yeah - I think you're right. Just flip the examples.

TomerHeber:
> See the last question, let's resolve that before the merge Overall looks good to me 👍🏼
> 
> Let's do it in a new minor version, and doc it.

Yes. That's the plan.
Let me know if I should merge it...

razbensimon:
yea lets

razbensimon:
Maybe release it after other open PRs, an isolated version of a single change @TomerHeber

PR #983, opened 2024-11-19
Fix: add pagination for teams api

### LINKED ISSUE #981: Pagination for `env0_teams` data source
**Describe the bug**
My organization has huge amount of teams (1176 at the moment). The `env0_teams` fails with

```
╷
│ Error: Could not get teams: 502 Bad Gateway: {"message": "Internal server error"}
│ 
│   with data.env0_teams.all_teams,
│   on env0_project_policy.tf line 1, in data "env0_teams" "all_teams":
│    1: data "env0_teams" "all_teams" {}
│ 
╵
```

**To Reproduce**

```
data "env0_teams" "all_teams" {}
```

**Expected behavior**

This should work without error.

**Provider Version**
1.22.1

**Screenshots**
N/A

**Additional context**

It looks like https://github.com/env0/terraform-provider-env0/blob/v1.22.1/env0/data_teams.go#L31 calling this code https://github.com/env0/terraform-provider-env0/blob/v1.22.1/client/team.go#L72-L87 and I am not seeing it attempting to do the pagination. Unless it is buried somewhere in the middleware, as I can see https://github.com/env0/terraform-provider-env0/blob/v1.22.1/client/pagination.go but I don't see where it is being injected into the HTTP client, so I am assuming it is not enabled for this call.

TomerHeber:
fixes #981 

### Solution

Added pagination for teams() api call.
Minor updates to code to add the pagination.
Updated unit tests.

### DIFF
--- a/client/team.go
+++ b/client/team.go
@@ -22,6 +22,11 @@ type Team struct {
 	Users          []User `json:"users"`
 }
 
+type PaginatedTeamsResponse struct {
+	Teams       []Team `json:"teams"`
+	NextPageKey string `json:"nextPageKey"`
+}
+
 func (client *ApiClient) TeamCreate(payload TeamCreatePayload) (Team, error) {
 	if payload.Name == "" {
 		return Team{}, errors.New("must specify team name on creation")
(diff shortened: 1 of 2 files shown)


client/environment.go
```diff
@@ -233,24 +233,20 @@ type EnvironmentDeployResponse struct {
 	Id string `json:"id"`
 }
 
-func (Environment) getEndpoint() string {
```
TomerHeber, line 236: not needed... I do it in a different way.

client/pagination.go
```diff
@@ -29,7 +24,7 @@ func (p *Pagination) next(currentPageSize int) bool {
 }
 
 // params - additional params. may be nil.
-func getAll[P Paginated](client *ApiClient, params map[string]string) ([]P, error) {
+func getAll[P any](client *ApiClient, endpoint string, params map[string]string) ([]P, error) {
```
TomerHeber, line 27: modified to support all types.

client/team.go
```diff
@@ -84,20 +84,28 @@ func (client *ApiClient) GetTeams(params map[string]string) ([]Team, error) {
...
 func (client *ApiClient) TeamsByName(name string) ([]Team, error) {
-	return client.GetTeams(map[string]string{"name": name})
+	organizationId, err := client.OrganizationId()
+	if err != nil {
+		return nil, err
+	}
+
+	var result []Team
+
+	if err := client.http.Get("/teams/organizations/"+organizationId, map[string]string{
```
TomerHeber, line 104: name param doesn't work with limit and offset... (bug in backend?)

razbensimon (approved):
NIce 👍🏼

PR #984, opened 2024-11-23
Feat: add the option to add a regex filter for list of teams

### LINKED ISSUE #982: Need to be able to lookup teams based on the list of names considering that not all of them may exists yet

Consider SSO enabled Env0 org. Env0 to my knowledge does not use SCIM at the moment, so the Teams are being populated to Env0 only when a first member of corresponding IdP group log in for the first time. Until then the team does not exists.

Consider a TF module that's used to create Env0 projects and assign permissions to corresponding teams. Such a module would accept a list of group names, which may include groups names that doesn't yet exists as Env0 Teams.

Trying to do something like this in that case would fail:

```
data "env0_team" "compartment_admins" {
  for_each = toset(local.okta_admin_groups)
  name     = each.key
}
```

The error would be something like

```
╷
│ Error: Could not find an env0 team with name <redacted>
│ 
│   with data.env0_team.compartment_admins["<redacted>"],
│   on env0_project_policy.tf line 1, in data "env0_team" "compartment_admins":
│    1: data "env0_team" "compartment_admins" {
│ 
╵
```

There needs to be a way to account for a fact that the team might not exist yet, so the code would not fail and process other groups that do. Maybe `env0_teams` data source needs to be able to provide some kind of filter as an input, but there needs to be corresponding API endpoint otherwise if TF providers would do the filtering client-side - that wouldn't actually solve anything.

As a workaround I considered the following approach

```
data "env0_teams" "all_teams" {}

data "env0_team" "compartment_admins" {
  for_each = toset([
    for group in data.env0_teams.all_teams.names : group
    if contains(local.okta_admin_groups, group)
  ])
  name = each.value
}
```

The downside is that `env0_teams` data source trying to get a list of all teams in the org, which is excessive and unnecessary. Additionally, it fails agains another issue described in https://github.com/env0/terraform-provider-env0/issues/981

**Additional context**

TomerHeber:
resolves #982 

### Solution

1. Added a regex filter to the data_teams schema, and implemented the code.
2. Added additional acceptance tests.


liranfarage89:
@TomerHeber  Im not sure how exactly it resolves #982 
[dee-kryvenko](https://github.com/dee-kryvenko) suggested a workdaround

```
data "env0_teams" "all_teams" {}

data "env0_team" "compartment_admins" {
  for_each = toset([
    for group in data.env0_teams.all_teams.names : group
    if contains(local.okta_admin_groups, group)
  ])
  name = each.value
}
```

>The downside is that env0_teams data source trying to get a list of all teams in the org, which is excessive and unnecessary. Additionally, it fails agains another issue described in https://github.com/env0/terraform-provider-env0/issues/981


This PR's implementation filters the full backend response.

TomerHeber:
> @TomerHeber Im not sure how exactly it resolves #982 [dee-kryvenko](https://github.com/dee-kryvenko) suggested a workdaround
> 
> ```
> data "env0_teams" "all_teams" {}
> 
> data "env0_team" "compartment_admins" {
>   for_each = toset([
>     for group in data.env0_teams.all_teams.names : group
>     if contains(local.okta_admin_groups, group)
>   ])
>   name = each.value
> }
> ```
> 
> > The downside is that env0_teams data source trying to get a list of all teams in the org, which is excessive and unnecessary. Additionally, it fails agains another issue described in #981
> 
> This PR's implementation filters the full backend response.

Hi Liran:

Have you observed the discussion in:
https://github.com/env0/terraform-provider-env0/issues/982

liranfarage89:
@TomerHeber I read it and still don't get why this filter is necessary as long as the user suggested a workaround that basically does the same

TomerHeber:
> @TomerHeber I read it and still don't get why this filter is necessary as long as the user suggested a workaround that basically does the same 

Got it!
Are you okay with closing this issue?

liranfarage89 (approved):
👍 

Writing here for future reference:
Although the filter happens in the client, the teams are globally cached.

PR #1006, opened 2025-01-24
Fix: support more vcs-type fields in module creation

### LINKED ISSUE #1002: Support vcs-type fields in module creation
**Describe the bug**
When someone creates a module, using token_id, we can't determine what kind of vcs-type is used

**Expected behavior**
I expect the user to be able and provide one of these fields:
```
'isGitLab',  'isAzureDevOps',  'isBitbucketServer'
'isGitHubEnterprise',  'isGitLabEnterprise'
```
and add these fields (false by default) to the request

**Screenshots**
Current module doc:

Image

**Additional context**

### LINKED ISSUE #1004: env0_module resource does not detect changes in ENV0 UI
**Describe the bug**
When the ENV0 module is created with terraform (exists in the state) and later is removed from UI or with API, terraform does not detect changes. 

**To Reproduce**
1. Create ENV0 module with terraform and env0_module resource
2. Remove that module using ENV0 UI
3. Run terraform plan again, there will be no changes

**Expected behavior**
env0_module detects that the module was removed and will show the diff that the resource will be created again

**Provider Version**
1.24.2

**Screenshots**
-

**Additional context**
ENV0 provider uses [this](https://github.com/env0/terraform-provider-env0/blob/main/client/module.go#L104) function to read the module from the ENV0 API. This will return the response even if the module was deleted, `"isDeleted":true` will be in the response JSON. IMHO that function should check if "isDeleted" is true or false in the response and return the module object when "isDeleted" is false

TomerHeber:
fixes #1002 
fixes #1004 

### Solution

1. Ignoring modules with isDeleted set to true (fixes #1004).
2. Updated the acceptance tests.
3. Added the missing is*** fields to support additional vcs types.
4. Updated the structs to support the additional fields.


env0/resource_module.go
```diff
@@ -12,12 +12,6 @@ import (
 )
 
 func resourceModule() *schema.Resource {
-	vcsExactlyOneOf := []string{
```
TomerHeber, line 15: using conflictsWith instrad.
Due to this comment: 
`when isGitLabEnterprise is true the token_id and token_name should not be required`

Wassap124 (approved):
Looks good!

env0/resource_module.go
```diff
@@ -119,6 +113,30 @@ func resourceModule() *schema.Resource {
 				Optional:    true,
 				Default:     false,
 			},
+			"is_gitlab": {
```
sagilaufer1992, line 116: why not using conflictsWith for these as well?
GL can't be BBS for example, seems like a good validation

TomerHeber: sure!

PR #1013, opened 2025-02-19
Feat: env0 Organization creation via provider

### LINKED ISSUE #1012: env0 Organization creation via provider
There's no way to create an env0 organization using the env0 provider, making it harder to automate and manage multiple organizations. A customer requested this, and we need to demo it in two weeks.

The env0 provider should support creating organizations, similar to [this API](https://docs.env0.com/reference/organization-create-organization).

Manually creating organizations or via the API using scripts outside of Terraform

**Additional context**
This would help manage multiple organizations with their own policies and teams in a structured and segregated way.

TomerHeber:
resolves #1012 

### Solution

1. Add new API calls.
2. Added unit tests for new API calls.
3. Added new resource.
4. Added acceptance tests for new resource.
5. Added an example.

Did not add a harness test for creating an organization! There's no option to delete an organization.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ tf
 
 # macosx
 .DS_Store
+test/
--- a/client/api_client.go
+++ b/client/api_client.go
@@ -23,6 +23,9 @@ type ApiClientInterface interface {
 	OrganizationId() (string, error)
 	OrganizationPolicyUpdate(OrganizationPolicyUpdatePayload) (*Organization, error)
 	OrganizationUserUpdateRole(userId string, roleId string) error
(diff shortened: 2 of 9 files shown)


TomerHeber:
The API key doesn't have the permission to view/use/create the  organization even though it's admin scope.

yaronya (approved):

omry-hay:
@yaronya I think we don't allow creating organizations using an API key for a reason.
@HeverFarber I think you talked about it with @Constantine19.
So I think we should revert this PR. 

cc @TomerHeber ☝️

yaronya:
@TomerHeber apologies for the confusion. I totally forgot that this is not allowed via API key.
I'll update the GH issue as well.

PR #1020, opened 2025-03-08
Feat: Environment Resource with properties from Project gets created …

### LINKED ISSUE #1014: Environment Resource with properties from Project gets created with a drift
**Describe the bug**
When I create a project with prPlan and apply settings for the environments like the following code:
```
#################### Create Project and environment with CD enabled via project ####################
resource "env0_project" "project_with_cd" {
  name = "test_my_project_with_cd"
}

  project_id = env0_project.project_with_cd.id

  continuous_deployment_default = true
  run_pull_request_plan_default = true
}

resource "env0_template_project_assignment" "assignment" {
  template_id = "96938846-481a-4e98-8f9a-dd5ac14c42c2"
  project_id  = env0_project.project_with_cd.id
}

resource "env0_environment" "environment_with_cd" {
  name        = "environment_with_cd"
  project_id  = env0_project.project_with_cd.id
  template_id = "96938846-481a-4e98-8f9a-dd5ac14c42c2"

  depends_on = [env0_project_policy.example, env0_template_project_assignment.assignment]
}
```

Since the environments inherits from the project:
```
  continuous_deployment_default = true
  run_pull_request_plan_default = true
```

It creates the environment with a drift, thus rerunning an apply drops these properties
```
  # env0_environment.environment_with_cd will be updated in-place
  ~ resource "env0_environment" "environment_with_cd" {
      - auto_deploy_on_path_changes_only = true -> null
      - deploy_on_push                   = true -> null
        id                               = "44dc9cb8-4fac-4b7e-a3ce-2d512e785a9c"
        name                             = "environment_with_cd"
      - run_plan_on_pull_requests        = true -> null
      - ttl                              = "2025-02-23T15:47:00.000Z" -> null
        # (12 unchanged attributes hidden)
    }

```

**To Reproduce**
1. Apply the TF code above (use your own template-id) -> Now env and proj has the prPlan and Apply checked in env0 
2. Rerun apply drops the properties as shown above

**Expected behavior**
Not removing the inherited project-properties from the environment

**Provider Version**
1.24.5

TomerHeber:
…with a drift

resolves #1014 

### Solution

This was not a straightforward solution.
I ignore the fields if they're not set in the environment schema.
This required some manual fixing in the environment data source.

It's not a perfect solution. But I believe there's no better way to resolve it.

### DIFF
--- a/env0/data_environment.go
+++ b/env0/data_environment.go
@@ -150,6 +150,24 @@ func dataEnvironmentRead(ctx context.Context, d *schema.ResourceData, meta inter
 		return diag.FromErr(err)
 	}
 
+	// Set this explicitly because these are not set in setEnvironmentSchema due to possible project inheritance.
+
+	if environment.AutoDeployOnPathChangesOnly != nil {
+		d.Set("auto_deploy_on_path_changes_only", *environment.AutoDeployOnPathChangesOnly)
+	}
+
+	if environment.ContinuousDeployment != nil {
+		d.Set("deploy_on_push", *environment.ContinuousDeployment)
+	}
+
+	if environment.PullRequestPlanDeployments != nil {
+		d.Set("run_plan_on_pull_requests", *environment.PullRequestPlanDeployments)
+	}
+
+	if environment.LifespanEndAt != "" {
+		d.Set("ttl", environment.LifespanEndAt)
+	}
+
 	subEnvironments := []interface{}{}
 
 	if environment.LatestDeploymentLog.WorkflowFile != nil {
--- a/env0/resource_environment_test.go
+++ b/env0/resource_environment_test.go
@@ -2011,6 +2011,75 @@ func TestUnitEnvironmentResource(t *testing.T) {
 				mock.EXPECT().EnvironmentDestroy(environment.Id).Times(1)
 			})
 		})
+
+		t.Run("No drift with unspecified fields", func(t *testing.T) {
+			truthyValue := true
+			falseyValue := false
+
+			// Environment with fields set by API
+			environmentWithAPIFields := client.Environment{
+				Id:                          uuid.New().String(),
+				Name:                        "my-environment",
+				ProjectId:                   "project-id",
+				WorkspaceName:               "workspace-name",
+				LifespanEndAt:               "2023-12-13T10:00:00Z", // TTL is set by API but not in schema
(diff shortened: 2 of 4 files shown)


env0/data_environment.go
```diff
@@ -150,6 +150,24 @@ func dataEnvironmentRead(ctx context.Context, d *schema.ResourceData, meta inter
 		return diag.FromErr(err)
 	}
 
+	// Set this explicitly because these are not set in setEnvironmentSchema due to possible project inheritance.
```
yarivg, line 153: Can we have a test validating the issue was fixed? basically using the same code that is described in the issue, and validating 2nd apply doesn't have a drift

yarivg (commented):
All in all, seems good.
I'd like to have a test to ensure it's working & not getting broken over time

env0/data_environment.go
TomerHeber: Added acceptance test:
91687ecb46de61b82995c763267d20be3d5ce588

env0/resource_environment_test.go
```diff
@@ -2011,6 +2011,75 @@ func TestUnitEnvironmentResource(t *testing.T) {
 				mock.EXPECT().EnvironmentDestroy(environment.Id).Times(1)
 			})
 		})
+
+		t.Run("No drift with unspecified fields", func(t *testing.T) {
```
yarivg, line 2015: There are 2 cases:
1. Environment can inherit the fields from the project. So given a `env0_project` and `env0_project_policy` that is specifying one of these values, the environment should not be created with a drift. 
This is similar to what you did, but I couldn't see where the `env0_project` and `env0_project_policy` are defined.
Can we test that? (might require some change to existing test)
2. `env0_environment` is defined with these fields, but then upon changing some value on the `env0_environment` resource, we should detect a drift for the environment resource... 
Can we have a test for that?

yarivg: we spoke, we'll go with this test for now. It gives some coverage by simulating the API result for the env0_environment entity and checks it doesn't create it with the inherited fields in the state

yarivg (approved):
LGTM,
We can have better coverage by elaborating the test util with option to detect drifts.
But this is out of scope for this task.
Might worth opening a task for it for future use cases.

PR #1028, opened 2025-03-30
Feat: Make sub_environment_configuration not require the revision field

### LINKED ISSUE #1025: Make `sub_environment_configuration` not require a the `revision` field.
the resource `sub_environment_configuration`, for some reason - REQUIRES the user to input a `revision`.
which has nothing to do with the variable.

ideally remove it altogether, but we don't wanna break old configurations for users,
so at least make it not mandatory.

TomerHeber:
resolves #1025 

### Solution

Marked the field in the schema as deprecated and optional. Also added a note to remove it in one of the future versions.


Wassap124 (approved):
💪

PR #1033, opened 2025-04-05
Feat: add support for Self Hosted VCS in env0_environment_discovery_c…

### LINKED ISSUE #1032: Add support for Self Hosted VCS in env0_environment_discovery_configuration
Be able to define `env0_environment_discovery_configuration` using a self-hosted vcs is not possible via the provider.

The following template properties are missing
```
is_bitbucket_server (Boolean) true if this template uses bitbucket server repository
is_github_enterprise (Boolean) true if this template uses github enterprise repository
is_gitlab_enterprise (Boolean) true if this template uses gitlab enterprise repository 
```

Not sure if this is needed as we have the `gitlab_project_id` already.
`is_gitlab (Boolean) set to 'true' if the repository is Gitlab`

N/A

TomerHeber:
…onfiguration

resolves #1032 

### Solution

1. Updated the API structs.
2. Added the schemas to the resource.
3. Added an acceptance test to verify that the fields are set to true when needed. (Otherwise, they default to false).

### DIFF
--- a/env0/resource_environment_discovery_configuration.go
+++ b/env0/resource_environment_discovery_configuration.go
@@ -154,6 +154,24 @@ func resourceEnvironmentDiscoveryConfiguration() *schema.Resource {
 				Default:      false,
 				RequiredWith: []string{"token_id"},
 			},
+			"is_bitbucket_server": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "set to true if Bitbucket Server is used",
+				Default:     false,
+			},
+			"is_github_enterprise": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "set to true if GitHub Enterprise is used",
+				Default:     false,
+			},
+			"is_gitlab_enterprise": {
+				Type:        schema.TypeBool,
+				Optional:    true,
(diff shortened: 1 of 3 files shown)


env0/resource_environment_discovery_configuration.go
```diff
@@ -154,6 +154,24 @@ func resourceEnvironmentDiscoveryConfiguration() *schema.Resource {
 				Default:      false,
 				RequiredWith: []string{"token_id"},
 			},
+			"is_bitbucket_server": {
```
yarivg, line 157: Just wondering, I don't think it is a real problem, but once a user who has a env_discovery resource with previous provider version and then upgrade to the version containing this change, will they see a drift? (from not having this props as false and now having them)

yarivg (approved):
Looks good

env0/resource_environment_discovery_configuration.go
TomerHeber: it's possible.
but generally there's a default: "false". So I believe if they match it wouldn't cause a drift.
if it's set to true in the in the backend - then it will probably cause a drift that users will need to fix and set to 'true' in the schema. (Which is expected).

PR #1073, opened 2025-11-10
Chore: env0_environment_drift_detection: retry on transient 'Invalid environment id' + migrate to helper/retry

liranfarage89:
## Summary
Add retry/backoff when enabling drift detection immediately after environment creation to eliminate flakiness, and migrate from deprecated helper/resource RetryContext to helper/retry.

Example:
https://github.com/env0/env0/actions/runs/19232722131/job/54974828355

## Root cause
The scheduling service may not immediately recognize a newly-created environment. When the drift detection resource tries to enable scheduling right away, the API can return 400 Bad Request with 'Invalid environment id'. This was causing the integration test 040_drift_detection to fail intermittently.

## Changes
- Wrap EnvironmentUpdateDriftDetection in RetryContext (helper/retry) with a 2-minute timeout.
- Retry only on 400 Bad Request errors that contain 'Invalid environment id'. Other errors are treated as non-retryable.
- Replace deprecated helper/resource RetryContext with helper/retry equivalents (RetryContext, RetryableError, NonRetryableError) and update imports.

## Why this works
It bridges the eventual consistency gap between environment creation and its availability to the scheduling subsystem, stabilizing the test and real-world applies.

## Testing
- Re-run integration: tests/integration/040_drift_detection should consistently pass.
- Confirm no deprecation warnings related to RetryContext.

## Notes
- Scope is limited to env0_environment_drift_detection. The environment resource drift update path can be similarly extended if we observe comparable flakiness there.

### DIFF
--- a/env0/resource_drift_detection.go
+++ b/env0/resource_drift_detection.go
@@ -2,10 +2,15 @@ package env0
 
 import (
 	"context"
+	"fmt"
+	"strings"
+	"time"
 
 	"github.com/env0/terraform-provider-env0/client"
+	env0http "github.com/env0/terraform-provider-env0/client/http"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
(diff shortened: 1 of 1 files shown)


yarden-fishler-dev (approved):
LGTM

PR #1077, opened 2025-12-14
Fix: env0_api_key resource does not assign user key permissions

eranelbaz:
### Issue & Steps to Reproduce / Feature Request
[//]: <> (choose “resolves” “fixes” or “closes” and put link for the issue)
If you follow our docs the provider does create the API key, but without project permissions

### Solution
Fix the JSON serialization

### QA
Built locally and ran this snippet and saw we assign correct permissions to the API Key
```
terraform {
  required_providers {
    env0 = {
      version = "6.6.6"
      source  = "terraform.env0.com/local/env0"
    }
  }
}

provider "env0" {
  api_key         = ""
  api_secret      = ""
  organization_id = ""

}

resource "env0_api_key" "combined_key" {
  name              = "Should-have-project-permissions"
  organization_role = "User"

  project_permissions {
    project_id   = ""
    project_role = "Viewer"
  }
}
```

PR #1078, opened 2025-12-25
Fix: [ENG-836] project_permissions JSON field name in API key creation (ENG-836)

liranfarage89:
## Problem
Users reported that API keys created with  blocks were not receiving the correct project access, resulting in 403 errors. This affected multiple customers including Adaptavist and Allstate.

## Root Cause
The  struct in  was using an incorrect JSON field name:
- **Expected**:  (singular)
- **Was**:  (plural)

This mismatch prevented project permissions from being properly assigned during API key creation.

## Solution
- Change JSON tag from  to  in the  struct
- Add comprehensive tests to verify correct JSON serialization behavior
- Ensure both empty and non-empty project permissions are handled correctly

## Changes
- **File**:  - Fixed JSON field name in struct tag
- **File**:  - Added tests for JSON serialization verification

## Testing
- All existing tests continue to pass (429/429 specs)
- New tests verify correct JSON field name usage
- Tests cover both empty and populated project permissions scenarios

## Impact
This fix ensures that API keys with  configurations receive the proper project assignments, resolving the 403 authorization errors reported by users.

Fixes: ENG-836

### DIFF
--- a/client/api_key.go
+++ b/client/api_key.go
@@ -20,7 +20,7 @@ type ProjectPermission struct {
 
 type ApiKeyPermissions struct {
 	OrganizationRole   string              `json:"organizationRole"`
-	ProjectPermissions []ProjectPermission `json:"projectPermissions,omitempty"`
+	ProjectPermissions []ProjectPermission `json:"projectsPermissions,omitempty"`
 }
 
 type ApiKeyCreatePayload struct {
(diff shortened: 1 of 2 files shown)


liranfarage89:
nvm it's irrelevant... resolved by @eranelbaz already #1077

PR #1089, opened 2026-03-15
Feat: add agent pool and agent secret resources

yarden-fishler-dev:
## Summary
- Add `env0_agent_pool` resource with full CRUD, import by id/name, and self-hosted logs configuration
- Add `env0_agent_pool` data source with lookup by name or id (ExactlyOneOf)
- Add `env0_agent_secret` resource (create/delete only, secret value stored in state as sensitive, compound import format `agentId/secretId`)
- Add client methods for agent pool and secret CRUD APIs
- Deprecate `env0_agent_values` data source in favor of `env0_agent_pool`
- Add integration test `041_agent_pool`

## Test plan
- [x] Unit tests for agent pool resource (8 tests: CRUD, logs create/update/remove, import, drift, error cases)
- [x] Unit tests for agent pool data source (7 tests: by id, by name, with logs, error cases)
- [x] Unit tests for agent secret resource (6 tests: create with/without description, import, drift, error cases)
- [x] Client unit tests for all new API methods (Ginkgo)
- [x] Integration test `041_agent_pool` (create, update via second_run, data source lookups, agent secret)
- [x] `go build ./...` passes

### Manual e2e testing against dev API

**Agent pool with logs:**
- [x] Create agent pool with logs (account_id + region) — verified via API: name preserved after POST+PATCH, logs set correctly
- [x] Remove logs block — verified via API: `logs: null` sent, logs cleared
- [x] Add logs back — verified via API: logs restored
- [x] Destroy — verified via API: 404

**Secret rotation flow:**
- [x] Create empty agent pool — verified via API: no secrets, no logs
- [x] Add 1 secret — verified via API: 1 secret listed
- [x] Delete secret — verified via API: 0 secrets
- [x] Add 2 secrets — verified via API: 2 secrets listed
- [x] Destroy all (secrets + pool) — verified via API: agent 404


env0/resource_agent_pool.go
```diff
@@ -0,0 +1,226 @@
...
+			Name:        d.Get("name").(string),
+			Description: d.Get("description").(string),
+			Logs:        logs,
+		}
+
+		_, err := apiClient.AgentPoolUpdate(agentPool.Id, updatePayload)
+		if err != nil {
+			return diag.Errorf("agent pool created successfully but failed to set logs configuration: %v", err)
+		}
+	}
```
RLRabinowitz: This is a problem in terraform/opentofu. It's not good practice to have the creation/update be two actions that each can persist data. 

Technically, if the creation succeeds, and the update fails, then terraform/tofu will think the agent pool wasn't created, but it actually was created. 

There's no way of having terraform/tofu "play nice" here, we can just make sure that such a failure causes a "taint", which means that a failure in the update taints the resource, and lets terraform/tofu know it does not know what the status of the resource is. Then, on the next apply, it'll destroy the resource and attempt to recreate it

Ideally - the API should just allow for self-hosted log configuration at creation time

env0/resource_agent_pool.go
```diff
@@ -0,0 +1,226 @@
...
+	// Try by id first; fall back to name lookup.
+	agentPool, err := apiClient.AgentPool(d.Id())
+	if err != nil {
+		tflog.Info(ctx, "Agent pool not found by id, trying by name", map[string]any{"id": d.Id()})
+
+		agentPool, err = getAgentPoolByName(d.Id(), meta)
+		if err != nil {
+			return nil, fmt.Errorf("could not get agent pool: %w", err)
+		}
+	}
```
RLRabinowitz: Should check if ID exists, if it does - fetch by ID. If it doesn't - fetch by name. Should not default to "by name" if the fetch by ID fails

client/agent.go
```diff
@@ -27,3 +27,127 @@ func (client *ApiClient) AgentValues(id string) (string, error) {
...
+	Description string `json:"description,omitempty"`
+	CreatedAt   string `json:"createdAt,omitempty"`
+	CreatedBy   User   `json:"createdBy,omitempty"`
+}
+
+type AgentSecretCreatePayload struct {
+	Description string `json:"description,omitempty"`
+}
+
+func (client *ApiClient) AgentPools() ([]AgentPool, error) {
```
RLRabinowitz, line 84: This is using the exact same API as the old `Agents` does - Should probably add a comment there regarding deprecation

env0/data_agent_values.go
```diff
@@ -10,7 +10,8 @@ import (
 
 func dataAgentValues() *schema.Resource {
 	return &schema.Resource{
-		ReadContext: dataAgentValuesRead,
+		ReadContext:        dataAgentValuesRead,
+		DeprecationMessage: "use env0_agent_pool data source instead",
```
RLRabinowitz: Should probably mention from which agent version this is not needed / supported. Consider a link to docs (not sure if we've done this before in the provider repo)

client/agent.go
```diff
@@ -27,3 +27,127 @@ func (client *ApiClient) AgentValues(id string) (string, error) {
 
 	return result, nil
 }
+
+// Agent Pool types and methods (self-service agents)
+
+type AgentPoolSelfHostedLogs struct {
+	AccountId string `json:"accountId"`
+	Region    string `json:"region"`
```
chpl: should also have `ExternalId`

env0/resource_agent_pool.go
yarden-fishler-dev: Fixed in 269fec9 — the API now supports logs in the create payload, so we send a single request instead of POST then PATCH.

client/agent.go
yarden-fishler-dev: externalId can only be the org ID or remain unchanged, so it's not really user-configurable. Do you think we should still expose it in the schema?

env0/resource_agent_pool.go
yarden-fishler-dev: The current implementation tries by ID first, then falls back to name lookup. Unlike api_key which uses UUID parsing to decide, we can't do that here since old agents have non-UUID IDs. Should we narrow the fallback to only trigger on NotFoundError (so a 500 doesn't silently fall through to a name lookup)?

client/agent.go
yarden-fishler-dev: f83e97e

env0/data_agent_values.go
yarden-fishler-dev: 2ce8477

client/agent.go
chpl: But the api needs it, no?

env0/resource_agent_pool.go
RLRabinowitz: Yes please

RLRabinowitz (approved):
See the 2 unresolved comments before merging

client/agent.go
yarden-fishler-dev: It's optional — the type is `externalId?: string` and the validation in `validateExternalId` skips entirely when it's null/undefined (`if (logs?.dynamo?.selfHosted?.externalId == null) return`).

env0/resource_agent_pool.go
yarden-fishler-dev: 021f376

PR #1094, opened 2026-04-12
fix: deprecate is_required in env0_environment configuration block

### LINKED ISSUE #1065: Deprecate and eventually remove `is_required` in `env0_environment`'s `configuration`
Currently, in the `configuration` blocks of `env0_environment`, you can define `is_required = true`. This is mostly pretty meaningless, as the purpose of `is_required` is to make sure that a value is supplied in deploy time (which is basically the `configuration` block in this case)

So, removing the `configuration` block that has `is_required = true` is not going to work, and fail with `failed deploying environment: 400 Bad Request: Missing required environment variable(s)`. 

A workaround is to remove the `is_required` and apply, and only later remove the `configuration` block

`is_required` should be removed from the `configuration` block of environments. We need to first deprecate the field, and remove it later down the road in a minor/major version change

An alternative is an env0-wide change, to change the behaviour of required variables and make them removable. This approach has its own downfalls, though

GiliFaroEnv0:
## Summary
- Deprecates the `is_required` field in the `env0_environment` `configuration` block — it has no effect since the value is already supplied inline
- Stops sending `is_required` to the API, preventing `400 Bad Request: Missing required environment variable(s)` errors when removing configuration blocks
- Stops reading `is_required` from API responses to avoid perpetual state drift

Closes #1065

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./env0/ -run TestEnvironment` passes
- [x] Verify that removing a `configuration` block with `is_required = true` no longer returns a 400 error
- [x] Verify that existing configs with `is_required` show a deprecation warning on `terraform plan`


chpl (commented):
@GiliFaroEnv0 pls
- remove `docs/superpowers/*`
- fix linter errros

chpl (approved):

PR #1098, opened 2026-05-19
Fix: prevent perpetual drift in sub-environment configuration blocks

yarivg:

`sub_environment_configuration.configuration` blocks show perpetual drift after `terraform apply`. The plan keeps re-adding the same changes, and applies never reach a converged state.

#### Minimal repro

```hcl
resource "env0_environment" "test" {
  name        = "drift-repro"
  project_id  = var.project_id
  template_id = var.workflow_template_id

    alias = var.sub_env_alias

    configuration {
      name         = "DB_PASSWORD"
      value        = "super-secret"
      is_sensitive = true
    }
  }
}
```

`terraform plan` immediately after a successful `terraform apply`:

```
~ sub_environment_configuration {
    ~ configuration {
        name  = "DB_PASSWORD"
      ~ value = "***************" -> "super-secret"
      }
  }

Plan: 0 to add, 1 to change, 0 to destroy.
```

Running `apply` again produces the same diff on the next `plan`. State never converges.

### Solution

Refresh sub-environment configuration in schema order and keep the schema value for sensitive variables (the API returns a redacted value, which previously got written into state). Remote-only variables are still surfaced as drift, so the original UI-edit drift detection still works.

After the fix, `terraform plan` against the same state and same TF code:

```
No changes. Your infrastructure matches the configuration.
```

---

- [x] Built local provider, repro'd perpetual drift on remote `env0/env0 v1.31.0` against a workflow env with multiple configuration blocks including a sensitive one
- [x] Swapped to local fixed build against the same state — `terraform plan` after `apply` reports `No changes`
- [x] Unit tests cover ordering, sensitive-value preservation, remote-only drift, and missing-from-API drop

### DIFF
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -728,9 +728,8 @@ func resourceEnvironmentRead(ctx context.Context, d *schema.ResourceData, meta a
 	return nil
 }
 
-func setSubEnvironmentSchema(d *schema.ResourceData, apiClient client.ApiClientInterface) any {
+func setSubEnvironmentSchema(d *schema.ResourceData, apiClient client.ApiClientInterface) error {
 	iSubEnvironments, ok := d.GetOk("sub_environment_configuration")
-
 	if !ok {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


GiliFaroEnv0 (approved):
LGTM — nice, targeted fix. The schema-order alignment plus keeping the schema value for sensitive vars is exactly the right approach, and the `any → error` signature tightening is a welcome cleanup (the caller already treated it as `error`).

A few non-blocking suggestions for a follow-up:

**1. Sensitive flag source.** The merge keys off `remote.IsSensitive` only:
```go
if remote.IsSensitive != nil && *remote.IsSensitive {
    newVariable["value"] = variable["value"]
}
```
If a user just toggled `is_sensitive = true` in HCL but the API still reports `false` during a transition, the redacted value could still land in state. Edge-case-y, but worth either ORing with `variable["is_sensitive"]` or adding a line to the doc comment.

**2. Duplicate-name behavior is implicit.** If `stateVariables` contains the same `name` twice, only the first is matched and the second is silently dropped (and remote isn't re-appended either, since the second loop's `found` check matches any state entry with that name). Probably fine — worth a one-line comment so a future reader knows it's intentional.

**3. O(n·m) lookups.** Both passes are nested linear scans. For typical config counts this is irrelevant; a `map[string]int` keyed on name would make it O(n+m) and arguably simpler. Skip unless you expect large configs.

**4. Minor style.** Second loop's `remote := &remoteVariables[i]` doesn't need the pointer (only `remote.Name` is read) — `for _, remote := range remoteVariables` reads cleaner.

**5. Test coverage gaps.** Unit tests on `mergeSubEnvironmentConfiguration` are great. Two things not covered:
- The new empty-`id` branch in `setSubEnvironmentSchema`
- The `error` return type change

Consider extending `TestUnitEnvironmentWithSubEnvironment` with a sensitive variable so the regression can't silently come back via the read path.

**6. Document the trade-off.** Remote-side edits to sensitive variable *values* are now invisible by design (the PR description notes this). Worth surfacing in the `mergeSubEnvironmentConfiguration` doc comment too, so a future reader doesn't "fix" it and reintroduce the perpetual drift.

Approving — none of the above are blocking.

PR #1093, opened 2026-04-12
Feat: extend VCS drift suppression to all VCS fields

Yossi-kerner:
## Summary
- Extends `suppressVcsFieldDrift` to suppress all legacy VCS fields (`github_installation_id`, `bitbucket_client_key`, `token_id`, `is_azure_devops`, `is_gitlab`) when `vcs_connection_id` is set, and vice versa
- Introduces a `VcsFields` struct to pass all relevant field pointers, replacing the old pairwise `(int, string)` signature
- Updates all 4 resources that call the function: template, approval_policy, custom_flow, environment_discovery_configuration
- Adds a test verifying all legacy fields are suppressed when only `vcs_connection_id` is configured

Closes ENG-1378

## Test plan
- [x] Existing pairwise drift tests still pass
- [x] New test: `vcs_connection_id` suppresses all backend-populated legacy VCS fields

## QA
applied the following tofu code and then ran plan and saw no changes
```
# GitLab via vcs_connection_id
resource "env0_template" "gitlab_vcs" {
  name              = "qa-gitlab-vcs"
  type              = "opentofu"
  opentofu_version    = "latest"
  repository        = "https://gitlab.com/env0/gitlab-vcs-integration-tests.git"
  path              = "null-resource"
  vcs_connection_id = "f35766dc-1427-4843-9050-a1b7b1ff6e8d"
}

# GitLab via legacy fields
resource "env0_template" "gitlab_legacy" {
  name       = "qa-gitlab-legacy"
  type       = "opentofu"
  opentofu_version    = "latest"
  repository = "https://gitlab.com/env0/gitlab-vcs-integration-tests.git"
  path       = "null-resource"
  is_gitlab  = true
  token_id   = "cb38eade-ab5d-4fec-b2d7-bd2450203726"
}

# Bitbucket via vcs_connection_id
resource "env0_template" "bb_vcs" {
  name              = "qa-bb-vcs"
  type              = "opentofu"
  opentofu_version    = "latest"
  repository        = "https://env0@bitbucket.org/env0/templates.git"
  vcs_connection_id = "8e76dc74-6f91-4f8e-9531-33cc6ba30b90"
}

# Bitbucket via legacy field
resource "env0_template" "bb_legacy" {
  name                 = "qa-bb-legacy"
  type                 = "opentofu"
  opentofu_version    = "latest"
  repository           = "https://env0@bitbucket.org/env0/templates.git"
[...]

PR #1100, opened 2026-06-04
[ENG-1740] Fix: prevent false drift in env0_module vcs_connection_id

liranfarage89:

Fixes [ENG-1740](https://linear.app/env-zero/issue/ENG-1740/env0-module-shows-false-drift-vcs-connection-id-flagged-for-removal-on)

When an `env0_module` is configured with **legacy VCS fields only** (e.g. `token_id`, `is_gitlab`, `github_installation_id`) and no `vcs_connection_id`, the backend resolves and returns a populated `vcs_connection_id`. `resourceModuleRead` wrote that value straight into state, so on the next plan Terraform saw config (empty) ≠ state (populated) and proposed removing `vcs_connection_id` — a perpetual, spurious `UPDATE` on every plan.

Repro:
1. Define an `env0_module` using a legacy VCS field (e.g. `token_id`), no `vcs_connection_id` in config
2. `terraform apply` — succeeds
3. `terraform plan` with no config changes
4. Module is flagged for `UPDATE`, `vcs_connection_id` shown as removed

`env0_module` was the one resource missed by #1093, which added `suppressVcsFieldDrift` to `env0_template`, `env0_approval_policy`, `env0_custom_flow`, and `env0_environment_discovery_configuration`.

### Solution

Replaced the partial, one-direction handling in `resourceModuleRead` (which only zeroed `github_installation_id` when `vcs_connection_id` was set) with the shared `suppressVcsFieldDrift` helper, matching the four resources from #1093. When legacy fields are configured, the backend-returned `vcs_connection_id` is now zeroed before being written to state.

Note: unlike the other resources, `client.Module` uses **pointer** types (`*int`/`*string`) for `GithubInstallationId`, `VcsConnectionId`, and `BitbucketClientKey`. Those are passed to `VcsFields` directly (they're already pointers); the value-typed fields (`TokenId`, `IsAzureDevOps`, `IsGitlab`) are passed by address. The helper's existing `nil` guard makes nil module pointers safe — a nil pointer is already "zeroed" from the `omitempty` writer's perspective — so no helper change was needed.

- [x] I verified it works
- [x] I verified I didn't break anything

- Added a regression unit test `legacy VCS field ignores vcs_connection_id from backend to avoid drift`: config sets `token_id`, the mocked backend returns a populated `vcs_connection_id`, and state is asserted to contain **no** `vcs_connection_id`. Confirmed the test **fails on the pre-fix code** and passes with the fix.
- The existing `Success with vcs_connection_id` and `vcs_connection_id ignores github_installation_id from backend to avoid drift` tests still pass, confirming the opposite direction (the previous behavior) is preserved.
- Ran the full `go test ./env0/` suite (exit 0), `go vet ./env0/`, and `gofmt` — all clean. `resourceModuleRead` is the only caller touched; create/update/delete and import flows are unaffected.

### DIFF
--- a/env0/resource_module.go
+++ b/env0/resource_module.go
@@ -217,10 +217,15 @@ func resourceModuleRead(ctx context.Context, d *schema.ResourceData, meta any) d
 		return nil
 	}
 
-	_, vcsConnectionIdOk := d.GetOk("vcs_connection_id")
-	if vcsConnectionIdOk {
-		module.GithubInstallationId = nil
-	}
+	// Avoid drifts: the backend automatically populates VCS fields from one another.
+	suppressVcsFieldDrift("", VcsFields{
+		GithubInstallationId: module.GithubInstallationId,
+		VcsConnectionId:      module.VcsConnectionId,
+		BitbucketClientKey:   module.BitbucketClientKey,
+		TokenId:              &module.TokenId,
+		IsAzureDevOps:        &module.IsAzureDevOps,
+		IsGitlab:             &module.IsGitlab,
+	}, d)
 
 	if err := writeResourceData(module, d); err != nil {
 		return diag.Errorf("schema resource data serialization failed: %v", err)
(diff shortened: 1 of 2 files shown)


Yossi-kerner (approved):
LGTM

PR #1103, opened 2026-06-24
Fix: implement prevent_auto_deploy flag to allow configuration updates without triggering deployments

jackcloudquery:

`prevent_auto_deploy` is respected when an `env0_environment` is **created**, but is
**ignored on update**. Changing `configuration`, `revision`, `variable_sets`, `template_id`,
or `sub_environment_configuration` on an existing environment triggers a deployment (run) -
even though the environment is explicitly flagged not to auto-deploy. Teams that manage the
*shape* of their environments with Terraform but run deployments themselves (CI/CD, behind
review gates) get runs kicked off outside their pipeline including production.

#### Minimal repro

```hcl
resource "env0_environment" "this" {
  name                = "example"
  project_id          = env0_project.this.id
  template_id         = var.template_id
  prevent_auto_deploy = true

  configuration {
    name  = "FOO"
    value = "bar" # change to "baz", apply -> a deployment is triggered
  }
}
```

After changing `value` to `"baz"`, `terraform apply` queues a new deployment on the
environment (env0 → Activity) instead of just updating the variable.

Root cause: `shouldDeploy` decides whether to deploy on update purely from which fields
changed and never consults `prevent_auto_deploy`, so the update path always hits the deploy
endpoint.

### Solution

On update, when `prevent_auto_deploy` is set, the provider no longer queues a deployment.
Instead it persists what env0 can apply *without* a run, directly:

- **`configuration`** and **`sub_environment_configuration.*.configuration`** - written via the
  configuration endpoints (create / update / delete per variable).
- **`variable_sets`** - assigned / unassigned via the configuration-set assignment endpoints.
- **`revision` / `template_id`** - these only take effect as part of a deployment, so under
  `prevent_auto_deploy` they're kept declaratively in state and applied by the user's own
  deploy. The read no longer overwrites them from the latest deployment log, avoiding
  perpetual drift.

This mirrors what the env0 backend already does at **create** time (`createWithoutDeploy`):
on create the provider just sends `preventAutoDeploy` in the payload and the backend forks;
on update there's no equivalent endpoint, so the provider makes the same decision client-side.

Notes (by design):

- A `revision` change under `prevent_auto_deploy` is recorded in state but is a no-op until the
  next deployment; external changes to `revision`/`template_id` are not surfaced as drift
  (documented on the field).
- Variables are persisted with sequential per-variable calls (not atomically like the deploy
  endpoint); a mid-loop failure leaves a partial apply that Terraform reconciles next run -
  same behavior as the standalone `env0_configuration_variable` resource.

After the fix the repro updates `FOO` to `baz` with **no** deployment queued, and
`terraform plan` afterwards reports `No changes`.

---

### Manual E2E verification:

- [x] Built the provider locally, wired in via `dev_overrides`, ran against the env0 **dev** stack
- [x] Created an `env0_environment` with `prevent_auto_deploy = true` and a `configuration`
      variable (self-contained config: project + template + assignment + environment)
- [x] Changed the variable value **and** `revision`, then `terraform apply` → **no new
      deployment** on the environment (env0 → Activity) and the variable value is updated
      (env0 → Variables)
- [x] `terraform plan` after the apply → `No changes. Your infrastructure matches the
      configuration.` (no drift on the variable or revision)
- [x] Unit tests cover update / create / delete of variables under `prevent_auto_deploy`
      (asserting `ConfigurationVariableUpdate/Create/Delete` fire and `EnvironmentDeploy` does
      **not**), variable-set assign/unassign, sub-environment configuration, and
      revision/template_id drift suppression

> Re @liranfarage89 (WORKFLOW-scoped top-level config): **confirmed.** The env0 backend treats
> `WORKFLOW` as a first-class scope on the direct `/configuration` write
> (`assert-configuration-scopes.ts`) and returns it for the matching `workflowEnvironmentId`
> query (`find-by-scope.ts`) - the exact query the provider's read issues. Verified end-to-end via
> an API round-trip on the dev stack: a direct `scope=WORKFLOW` write returned HTTP 200 and was
> persisted (`scopeSortKey=WORKFLOW:<envId>`), then read back with `scope=WORKFLOW, scopeId=<envId>`.

Closes https://linear.app/env-zero/issue/APO-96/preventautodeploy-is-ignored-on-certain-environment-updates

### DIFF
--- a/docs/resources/environment.md
+++ b/docs/resources/environment.md
@@ -73,7 +73,7 @@ If true must specify one of the following - 'github_installation_id' if using Gi
 - `is_remote_backend` (Boolean) should use remote backend
 - `k8s_namespace` (String) kubernetes (or helm) namespace to be used. If modified deletes current environment and creates a new one
 - `output` (String) the deployment log output. Returns a json string. It can be either a map of key-value, or an array of (in case of Terragrunt run-all) of moduleName and a map of key-value. Note: if the deployment is still in progress returns 'null'
-- `prevent_auto_deploy` (Boolean) use this flag to prevent auto deploy on environment creation
+- `prevent_auto_deploy` (Boolean) use this flag to prevent the provider from triggering a deployment (run) when the environment is created or updated. On update, changes to 'configuration' and 'variable_sets' are still saved to env0 without a deployment, while changes to 'revision' and 'template_id' are only applied by your next deployment
 - `removal_strategy` (String) by default when removing an environment, it gets destroyed. Setting this value to 'mark_as_archived' will force the environment to be archived instead of tying to destroy it ('Mark as inactive' in the UI)
 - `revision` (String) the revision the environment is to be run against. Please note that changing this attribute will require environment redeploy
 - `run_plan_on_pull_requests` (Boolean) should run terraform plan on pull requests creations.
--- a/docs/resources/environment.md
+++ b/docs/resources/environment.md
@@ -73,7 +73,7 @@ If true must specify one of the following - 'github_installation_id' if using Gi
 - `is_remote_backend` (Boolean) should use remote backend
 - `k8s_namespace` (String) kubernetes (or helm) namespace to be used. If modified deletes current environment and creates a new one
 - `output` (String) the deployment log output. Returns a json string. It can be either a map of key-value, or an array of (in case of Terragrunt run-all) of moduleName and a map of key-value. Note: if the deployment is still in progress returns 'null'
-- `prevent_auto_deploy` (Boolean) use this flag to prevent the provider from triggering a deployment (run) when the environment is created or updated. On update, changes to 'configuration' and 'variable_sets' are still saved to env0 without a deployment, while changes to 'revision' and 'template_id' are only applied by your next deployment
+- `prevent_auto_deploy` (Boolean) use this flag to prevent the provider from triggering a deployment (run) when the environment is created or updated. On update, changes to 'configuration', 'sub_environment_configuration' variables and 'variable_sets' are still saved to env0 without a deployment, while changes to 'revision' and 'template_id' are only applied by your next deployment
 - `removal_strategy` (String) by default when removing an environment, it gets destroyed. Setting this value to 'mark_as_archived' will force the environment to be archived instead of tying to destroy it ('Mark as inactive' in the UI)
 - `revision` (String) the revision the environment is to be run against. Please note that changing this attribute will require environment redeploy
 - `run_plan_on_pull_requests` (Boolean) should run terraform plan on pull requests creations.
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -272,7 +272,7 @@ func resourceEnvironment() *schema.Resource {
 			},
 			"prevent_auto_deploy": {
 				Type:        schema.TypeBool,
-				Description: "use this flag to prevent auto deploy on environment creation",
+				Description: "use this flag to prevent the provider from triggering a deployment (run) when the environment is created or updated. On update, changes to 'configuration' and 'variable_sets' are still saved to env0 without a deployment, while changes to 'revision' and 'template_id' are only applied by your next deployment",
 				Optional:    true,
 			},
 			"terragrunt_working_directory": {
@@ -433,7 +433,8 @@ func setEnvironmentSchema(ctx context.Context, d *schema.ResourceData, environme
 	}
 
 	if !isTemplateless(d) {
-		if environment.LatestDeploymentLog.BlueprintId != "" {
+		preventAutoDeploy, _ := d.Get("prevent_auto_deploy").(bool)
+		if environment.LatestDeploymentLog.BlueprintId != "" && !preventAutoDeploy {
 			d.Set("template_id", environment.LatestDeploymentLog.BlueprintId)
 			d.Set("revision", environment.LatestDeploymentLog.BlueprintRevision)
 		}
@@ -844,7 +845,11 @@ func resourceEnvironmentUpdate(ctx context.Context, d *schema.ResourceData, meta
 	}
 
 	if shouldDeploy(d) {
-		if err := deploy(d, apiClient); err != nil {
+		if d.Get("prevent_auto_deploy").(bool) {
+			if diagErr := updateWithoutDeploy(d, apiClient); diagErr != nil {
+				return diagErr
+			}
+		} else if err := deploy(d, apiClient); err != nil {
 			return err
 		}
 	}
@@ -924,6 +929,21 @@ func updateDriftDetection(d *schema.ResourceData, apiClient client.ApiClientInte
 	return nil
 }
 
+func getEnvironmentConfigurationScope(d *schema.ResourceData) client.Scope {
+	if _, ok := d.GetOk("sub_environment_configuration"); ok {
+		return client.ScopeWorkflow
+	}
+
+	return client.ScopeEnvironment
+}
+
+func getSubEnvironmentConfigurationChanges(d *schema.ResourceData, index int, subEnvironmentId string, apiClient client.ApiClientInterface) (client.ConfigurationChanges, error) {
+	configuration := d.Get(fmt.Sprintf("sub_environment_configuration.%d.configuration", index)).([]any)
+	configurationChanges := stripIsRequired(getConfigurationVariablesFromSchema(configuration))
+
+	return getUpdateConfigurationVariables(configurationChanges, subEnvironmentId, client.ScopeEnvironment, apiClient)
+}
+
 func deploy(d *schema.ResourceData, apiClient client.ApiClientInterface) diag.Diagnostics {
 	deployPayload, err := getDeployPayload(d, apiClient, true)
 	if err != nil {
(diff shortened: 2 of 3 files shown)


env0/resource_environment.go
```diff
@@ -983,6 +1000,157 @@ func deploy(d *schema.ResourceData, apiClient client.ApiClientInterface) diag.Di
...
+
+		if change.Id != "" {
+			if _, err := apiClient.ConfigurationVariableUpdate(client.ConfigurationVariableUpdateParams{Id: change.Id, CommonParams: params}); err != nil {
+				return fmt.Errorf("could not update configuration variable %q: %w", change.Name, err)
+			}
+
+			continue
+		}
+
+		if _, err := apiClient.ConfigurationVariableCreate(params); err != nil {
```
liranfarage89, line 1139: Verify this works at `WORKFLOW` scope. For a workflow env, top-level config is applied here via a direct `ConfigurationVariableCreate`/`Update` with `scope=WORKFLOW`. The deploy path never creates WORKFLOW-scoped vars directly (it ships them in the deploy payload for the backend to apply), and no other provider path POSTs a WORKFLOW-scoped var to `/configuration`. Worth testing against a real workflow env that creating a new top-level var with `prevent_auto_deploy` actually persists.

env0/resource_environment.go
```diff
@@ -983,6 +1000,157 @@ func deploy(d *schema.ResourceData, apiClient client.ApiClientInterface) diag.Di
...
+	if len(changes.Unassign) > 0 {
+		if err := apiClient.UnassignConfigurationSets(client.ENVIRONMENT, d.Id(), changes.Unassign); err != nil {
+			return fmt.Errorf("could not unassign variable sets: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func applyConfigurationChangesWithoutDeploy(configurationChanges client.ConfigurationChanges, scope client.Scope, scopeId string, apiClient client.ApiClientInterface) error {
```
liranfarage89, line 1117: These run as sequential per-variable create/update/delete calls, so a mid-loop failure leaves the env half-applied and errors out (Terraform re-reconciles next apply). The deploy endpoint applies atomically. Consistent with the standalone config-var resource, so fine — just worth a comment noting the non-atomic behavior.

env0/resource_environment_test.go
```diff
@@ -2011,6 +2011,201 @@ func TestUnitEnvironmentResource(t *testing.T) {
 			})
 		})
 
+		t.Run("prevent_auto_deploy - update persists configuration without deploying", func(t *testing.T) {
```
liranfarage89, line 2014: Tests only cover the update branch of `applyConfigurationChangesWithoutDeploy`. The create (`change.Id == ""`) and delete (`change.ToDelete`) branches are uncovered. Add a step that adds a new var and drops an existing one under `prevent_auto_deploy`, asserting `ConfigurationVariableCreate`/`ConfigurationVariableDelete` fire and no deploy happens.

env0/resource_environment.go
```diff
@@ -433,7 +433,8 @@ func setEnvironmentSchema(ctx context.Context, d *schema.ResourceData, environme
 	}
 
 	if !isTemplateless(d) {
-		if environment.LatestDeploymentLog.BlueprintId != "" {
+		preventAutoDeploy, _ := d.Get("prevent_auto_deploy").(bool)
```
liranfarage89, line 443: nit: skipping the `template_id`/`revision` sync when `prevent_auto_deploy` is set is the right call to avoid drift, but note it also means genuine external changes to those fields won't be detected, and a `revision` change silently no-ops (kept in state, never deployed). By-design here and the docs cover it.

yarivg (commented):
@jackcloudquery how did you QA it? mind sharing the example to the PR desc

jackcloudquery:
> @jackcloudquery how did you QA it? mind sharing the example to the PR desc

Added to the PR description

docs/resources/environment.md
```diff
@@ -73,7 +73,7 @@ If true must specify one of the following - 'github_installation_id' if using Gi
 - `is_remote_backend` (Boolean) should use remote backend
 - `k8s_namespace` (String) kubernetes (or helm) namespace to be used. If modified deletes current environment and creates a new one
 - `output` (String) the deployment log output. Returns a json string. It can be either a map of key-value, or an array of (in case of Terragrunt run-all) of moduleName and a map of key-value. Note: if the deployment is still in progress returns 'null'
-- `prevent_auto_deploy` (Boolean) use this flag to prevent auto deploy on environment creation
+- `prevent_auto_deploy` (Boolean) use this flag to prevent the provider from triggering a deployment (run) when the environment is created or updated. On update, changes to 'configuration' and 'variable_sets' are still saved to env0 without a deployment, while changes to 'revision' and 'template_id' are only applied by your next deployment
 - `removal_strategy` (String) by default when removing an environment, it gets destroyed. Setting this value to 'mark_as_archived' will force the environment to be archived instead of tying to destroy it ('Mark as inactive' in the UI)
```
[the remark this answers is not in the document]
jackcloudquery: Updated, regenerated docs to reflect that sub-environment
configuration variables are persisted without a deployment too.

env0/resource_environment.go
```diff
@@ -272,7 +272,7 @@ func resourceEnvironment() *schema.Resource {
 			},
 			"prevent_auto_deploy": {
 				Type:        schema.TypeBool,
-				Description: "use this flag to prevent auto deploy on environment creation",
+				Description: "use this flag to prevent the provider from triggering a deployment (run) when the environment is created or updated. On update, changes to 'configuration' and 'variable_sets' are still saved to env0 without a deployment, while changes to 'revision' and 'template_id' are only applied by your next deployment",
 				Optional:    true,
```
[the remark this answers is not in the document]
jackcloudquery: Updated the prevent_auto_deploy description to include that
sub_environment_configuration.*.configuration is also persisted without a deployment.

env0/resource_environment.go
```diff
@@ -983,6 +1000,157 @@ func deploy(d *schema.ResourceData, apiClient client.ApiClientInterface) diag.Di
...
+		return fmt.Errorf("failed to extract sub environments from resource data: %w", err)
+	}
+
+	for i, subEnvironment := range subEnvironments {
+		if subEnvironment.Id == "" {
+			continue
+		}
+
+		configurationChanges, err := getSubEnvironmentConfigurationChanges(d, i, subEnvironment.Id, apiClient)
+		if err != nil {
```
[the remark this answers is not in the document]
jackcloudquery: Fixed: added a per-sub-environment guard so we only recompute/apply variable deltas when that
sub-env's own `configuration` block changed (HasChange on
sub_environment_configuration.<i>.configuration), so changes to workspace/approve_plan_automatically
no longer touch variables.

env0/resource_environment.go
jackcloudquery: Added a comment about that here https://github.com/env0/terraform-provider-env0/pull/1103/changes#diff-5dfeada10c7bdbcd4234860009b1a414cb4d99b87791eb4cb6ad6c3b7c3f703aR1113

env0/resource_environment.go
jackcloudquery: Confirmed the env0 backend supports it: [assert-configuration-scopes.ts](https://github.com/env0/env0/blob/0111bb7ceee8965abc98a2796ab52496099305f5/services/configuration-service/functions/variables/common/assert-configuration-scopes.ts) accepts WORKFLOW writes and [find-by-scope.ts](https://github.com/env0/env0/blob/77cb9c0d057bb74c2fa2aee3ec98c3cb264b29ca/services/configuration-service/functions/common/find-by-scope.ts) returns them for workflowEnvironmentId (the provider's read query). Then verified end-to-end with a live API round-trip on the dev stack (HTTP 200 write → read back at scope=WORKFLOW). [Documented](https://github.com/env0/terraform-provider-env0/pull/1103/changes#diff-5dfeada10c7bdbcd4234860009b1a414cb4d99b87791eb4cb6ad6c3b7c3f703aR1034) the support in a code comment.

env0/resource_environment_test.go
jackcloudquery: Added a [new test](https://github.com/env0/terraform-provider-env0/pull/1103/changes#diff-ac99e6aa060e0755de8e0203d6d0b2b28fcae669dd13a6144628b46a645d2b27R2120) that drops an existing var (delete branch) and adds a new var (create branch) under prevent_auto_deploy, asserting ConfigurationVariableCreate + ConfigurationVariableDelete fire and EnvironmentDeploy/ConfigurationVariableUpdate do not. Passes.

env0/resource_environment.go
jackcloudquery: Expanded the [comment](https://github.com/env0/terraform-provider-env0/pull/1103/changes#diff-5dfeada10c7bdbcd4234860009b1a414cb4d99b87791eb4cb6ad6c3b7c3f703aR436) to explicitly state the trade-off: external changes to revision/template_id aren't detected, and a revision change is kept in state but only takes effect on the next deploy. The field's docs already describe it.

liranfarage89 (approved):
LGTM. All comments addressed — verified the WORKFLOW-scope create against the backend + live round-trip, the per-sub-env guard, and the new create/delete test. CI green. 🙏

PR #1107, opened 2026-07-06
[APO-118] Feat: Add is_read_only support to env0_variable_set variables

dcelasun:
Wire the is_read_only attribute into the variable {} block of the env0_variable_set resource. The backend API and the client struct (ConfigurationVariable.IsReadOnly) already supported it; only the resource schema and its config<->state mapping were missing.

- Add is_read_only (Boolean, Optional, default false) to the variable schema
- Map it in getVariableFromSchema (config -> API) and getSchemaFromVariables (API -> state)
- Cover it in the unit test and example, regenerate docs

**Manual verification**

- `go build .` and `go test -race -v ./...` are clean.

Fixes https://linear.app/env-zero/issue/APO-118/add-is-read-only-support-to-env0-variable-set-variables

### DIFF
--- a/docs/resources/variable_set.md
+++ b/docs/resources/variable_set.md
@@ -59,6 +59,14 @@ resource "env0_variable_set" "organization_scope_example" {
     type            = "terraform"
     format          = "dropdown"
   }
+
+  variable {
+    name         = "n6"
+    value        = "v6"
+    type         = "terraform"
+    format       = "text"
+    is_read_only = true
+  }
 }
 
 resource "env0_variable_set" "project_scope_example" {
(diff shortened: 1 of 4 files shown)


liranfarage89 (approved):
✅ LGTM — small, focused, and well-tested. Validated the cross-repo claims against the backend and docs before approving:

- **Casing is correct and load-bearing.** The client serializes `isReadonly` (lowercase "o"), which matches the backend exactly — there is no `isReadOnly` anywhere in `env0/env0`. And since the ConfigurationSet request schemas use `additionalProperties: false`, a mis-cased field would 400 rather than silently no-op, so this had to be exact. It is.
- **The ConfigurationSet create/update path honors it.** Both set endpoints funnel variables through the same `createOrUpdate` flow that persists `isReadonly`, so it works for variable-set variables specifically, not just standalone variables.
- **Semantics match.** "Cannot be overridden in a lower scope" lines up with both the backend merge logic and the public docs (verbatim in the admin guide).

PR #1109, opened 2026-07-07
Chore: parallelize unit tests to cut CI time (~13m → ~8m)

liranfarage89:
## Why

CI takes ~13 min on every PR. `unit-tests` and `integration-tests` run in parallel, so wall-clock = the slower job. **Unit Tests was the critical path**, dominated by the `Go Test` step.

The "unit tests" use a **fully mocked** API client (`env0/provider_test.go`) — they don't hit a backend, but they spawn the real `terraform` binary ~1,200× (620 test cases × ~2 apply steps). Two things made that slow:

1. **`TF_LOG=DEBUG`** floods every one of those ~1,200 invocations with debug logging — pure waste on green runs.
2. **No top-level parallelism.** 619/620 `runUnitTest` calls are wrapped in `t.Run(...)`, and `runUnitTest` uses `resource.ParallelTest`, which only parallelizes *subtests within one function*. The 135 top-level `TestUnit*` functions ran **sequentially**, so concurrency reset per function instead of running sustained.

## Changes

- **`ci.yml`:** `TF_LOG=DEBUG go test -v ./...` → `TF_LOG_PROVIDER=info go test -race -timeout 20m -parallel 16 ./...`
  - keeps `-race` (from #1108) — the PR's own CI now proves the parallelized tests are race-clean;
  - `-parallel 16` since terraform-per-step is IO/process-bound;
  - `TF_LOG_PROVIDER=info` (per review) surfaces provider logs on *failing* tests at ~no cost — measured 228.8s vs 222.6s, within noise, and zero extra output on green runs since `-v` is off;
  - dropped `TF_LOG=DEBUG`/`-v` (huge per-invocation overhead).
- **`provider_test.go`:** moved the shared `os.Setenv` calls into a `TestMain` so they run once — a per-call data race under parallel execution.
- **102 top-level test functions:** added `t.Parallel()`. Functions that call `runUnitTest` directly on the top-level `t` are already parallel via `ParallelTest`, so they're **skipped** (a second `t.Parallel()` would panic). Done with a Go AST rewriter, not a blind sed, to get that distinction right.
- **integration job:** added a Go module/build cache.

## Coverage

Unchanged — no test removed, same assertions, just parallel. Full suite passes; `golangci-lint` v2.9.0 (matches CI) reports 0 issues.

## Results (real CI, run 28859062268)

| Job / step | Before (current main, incl. `-race` from #1108) | This PR | Δ |
|---|---|---|---|
| **Go Test** | 13m12s | **10m54s** | **~18% faster** |
| Unit Tests (job) | ~13.5m | **12m04s** | |
| Integration Tests (job) | ~8.5m | 8m48s | (unchanged; real backend floor) |

**Note on the ceiling:** #1108 added `-race`, which turns these tests from IO/process-bound into **CPU-bound**. On the 4-vCPU runner that caps how much `-parallel` can help — locally, *without* `-race`, the same change runs 406s → 221s (~46% faster), but with `-race` on 4 cores the realizable win is ~18%. Further reductions would need a larger runner (more cores/RAM directly speed up `-race`) or moving `-race` off the PR critical path (e.g. nightly) — deliberately out of scope here. The integration job's ~8.5m (real env0 environment deployments) remains the wall-clock floor.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,9 +61,10 @@ jobs:
           echo "Using Terraform at: $TF_PATH"
           echo "Terraform version: $TERRAFORM_VERSION"
           # -race keeps the data-race detector (APO-185); -parallel oversubscribes the
-          # runner since terraform-per-step is IO/process-bound. TF_LOG=DEBUG/-v omitted
-          # on purpose (huge per-invocation overhead) — re-run locally with them to debug.
-          go test -race -timeout 20m -parallel 16 ./...
+          # runner since terraform-per-step is IO/process-bound. TF_LOG_PROVIDER=info
+          # surfaces provider logs on failing tests at ~no cost (measured); full
+          # TF_LOG=DEBUG/-v omitted on purpose (huge per-invocation overhead).
+          TF_LOG_PROVIDER=info go test -race -timeout 20m -parallel 16 ./...
         env:
           TF_ACC: true
           TF_ACC_TERRAFORM_PATH: ${{ env.TF_PATH }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,8 +60,9 @@ jobs:
         run: |
           echo "Using Terraform at: $TF_PATH"
           echo "Terraform version: $TERRAFORM_VERSION"
-          # Run tests with detailed logging and increased timeout
-          TF_LOG=DEBUG go test -timeout 20m -v ./...
+          # Tests run terraform per step; -parallel oversubscribes the runner since steps are IO/process-bound.
+          # TF_LOG=DEBUG omitted on purpose (huge per-invocation overhead); re-run locally with it to debug a failure.
+          go test -timeout 20m -parallel 16 ./...
         env:
           TF_ACC: true
           TF_ACC_TERRAFORM_PATH: ${{ env.TF_PATH }}
@@ -78,5 +79,13 @@ jobs:
         run: apk add opentofu
       - name: Checkout code
         uses: actions/checkout@v4
+      - name: Cache Go modules and build
+        uses: actions/cache@v4
+        with:
+          path: |
+            /go/pkg/mod
+            /root/.cache/go-build
+          key: integ-go-${{ hashFiles('go.sum') }}
+          restore-keys: integ-go-
       - name: Run Harness tests
         run: go run tests/harness.go
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,9 @@ jobs:
   unit-tests:
     name: Unit Tests
     timeout-minutes: 15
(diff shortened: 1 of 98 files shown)


.github/workflows/ci.yml
```diff
@@ -60,8 +60,10 @@ jobs:
         run: |
           echo "Using Terraform at: $TF_PATH"
           echo "Terraform version: $TERRAFORM_VERSION"
-          # Run tests with the race detector, detailed logging and increased timeout
-          TF_LOG=DEBUG go test -race -timeout 20m -v ./...
+          # -race keeps the data-race detector (APO-185); -parallel oversubscribes the
+          # runner since terraform-per-step is IO/process-bound. TF_LOG=DEBUG/-v omitted
+          # on purpose (huge per-invocation overhead) — re-run locally with them to debug.
+          go test -race -timeout 20m -parallel 16 ./...
```
chpl: Integration tests are using [TF_LOG_PROVIDER=info](https://github.com/env0/terraform-provider-env0/blob/13aa5481f3c70710f76b58ac9d6b26a8dd214430/tests/harness.go#L203) - not sure if it will affect the UT, can you try, pls.

chpl (approved):

.github/workflows/ci.yml
liranfarage89: Good call — added `TF_LOG_PROVIDER=info` to the Go Test step (matching the harness).

Measured both ways locally (`-parallel 16`, full suite):
- without it: **228.8s**
- with `TF_LOG_PROVIDER=info`: **222.6s** (within noise)

No runtime impact, and since `-v` is off the output on green runs is identical (provider logs are buffered/discarded), while a *failing* test still surfaces its provider-level info logs. Best of both.

chpl:
@liranfarage89 maybe we also should use larger runners (both CI and integration tests)

liranfarage89:
@chpl done — moved `unit-tests` to an 8-vCPU Blacksmith runner (`blacksmith-8vcpu-ubuntu-2404`). It made a big difference because Go Test is CPU-bound under `-race`, so the extra cores + RAM unlocked the parallelism the 4-vCPU runner was capping:

| Step | 4-vCPU | 8-vCPU | Δ |
|---|---|---|---|
| **Go Test** | 10m54s | **3m25s** | **~69% faster** |
| Unit Tests (job) | 12m04s | **4m18s** | |
| Integration | ~8.8m | ~9.1m | unchanged |

I left **integration on the standard runner** on purpose — it's backend-bound, not runner-bound (all 41 tests start at once, ~36 finish within 3 min, and the ~9m tail is a few tests waiting on real env0 environment deployments while the runner sits idle). A larger runner there would cost more for ~zero gain; the real lever is backend/deployment speed.

Net: unit tests went from the critical path to a non-issue, and overall CI is now ~13m → ~9m, floored by the integration job. Merged 🎉

PR #1117, opened 2026-07-22
[APO-291] Fix: Resolve env0_template by name via server-side filter

yarivg:

[APO-291](https://linear.app/env-zero/issue/APO-291) — concurrent Terraform plans return partial HTTP 500s from the templates API.

The `env0_template` data source resolves a template by name via `GET /blueprints?organizationId=…`, which returns the entire org template list and matches client-side. Under many concurrent plans these full-list reads overload the API and surface as partial 500s.

### Solution

- `client/template.go` — new `TemplatesByName(name)` calling `GET /blueprints?organizationId=…&name=…`, so the API returns only the matching template(s).
- `env0/data_template.go` — `getTemplateByName` uses `TemplatesByName`; the `!IsDeleted` filter and duplicate-name guard are unchanged.
- `client/api_client.go` / `client/api_client_mock.go` — interface method + mock entry.

**Compatibility:** backward compatible. Against an API without the `name` filter, the param is ignored and the full list is returned, so the existing client-side match still resolves correctly. The plural `env0_templates` data source is unchanged.

---

- [x] How did I verify this works

  Ran the provider against an API with the filter available: `data "env0_template" { name = "…" }` resolved the correct template id, and the outgoing request was `GET /blueprints?organizationId=…&name=…` returning only the matching template.

- [x] How did I verify I didn't break anything

  Resolution by `id` is untouched; `env0_templates` still calls `Templates()`; duplicate-name and deleted-template handling in `getTemplateByName` is unchanged; behavior is identical against an API that ignores the new param.

### DIFF
--- a/client/api_client.go
+++ b/client/api_client.go
@@ -35,6 +35,7 @@ type ApiClientInterface interface {
 	ModuleTestingProject() (*ModuleTestingProject, error)
 	Template(id string) (Template, error)
 	Templates() ([]Template, error)
+	TemplatesByName(name string) ([]Template, error)
 	TemplateCreate(payload TemplateCreatePayload) (Template, error)
 	TemplateUpdate(id string, payload TemplateCreatePayload) (Template, error)
 	TemplateDelete(id string) error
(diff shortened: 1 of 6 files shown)


liranfarage89 (approved):
Looks great

PR #1118, opened 2026-07-23
Chore: summarize integration test failures with environment links

yaronya:

When an integration test fails in CI, the harness halts on the first failure and the actual error is buried in thousands of interleaved provider log lines. For environment-related failures (e.g. `failed to wait for environment destroy to complete, deployment status is: INTERNAL_FAILURE`) there is no indication of *which* environment the test ran against, making failures hard to investigate.

### Solution

Improve the integration test harness (`tests/harness.go`) output — no provider code changes:

- **Run all tests to completion** instead of aborting on the first failure (which also killed other in-flight tests and skipped their cleanup destroys), then exit non-zero if any test failed
- **Print a pass/fail summary** at the end of the run, with only the terraform error diagnostics (`╷ Error ╵` blocks) extracted per failed test, filtering out the verbose provider logs
- **Link failed tests to their env0 environments**: before cleanup, capture the `env0_environment` resources still in the failed test's terraform state (`tofu show -json`, including child modules) and print each environment's name, id, and a direct link to it in the env0 UI, derived from the configured `ENV0_API_ENDPOINT`
- **Write the summary as markdown to `GITHUB_STEP_SUMMARY`** so it renders on the workflow run's summary page (pass/fail table + per-failure environment links and error excerpt)
- Fix output capture: include stderr in returned command errors, and capture process output directly instead of via unflushed buffered writers whose tail could be lost

### Output before / after

**Before** — the run ends mid-flight on the first failure, with no summary and no hint of which environment failed:

```
2026-07-23T09:08:32 [ERROR] provider.terraform-provider-env0: Response contains error diagnostic: ... diagnostic_summary="failed to wait for environment destroy to complete, deployment status is: INTERNAL_FAILURE" ...
╷
│ Error: failed to wait for environment destroy to complete, deployment status is: INTERNAL_FAILURE
│
╵
2026/07/23 09:08:32 error running tofu [destroy -auto-approve -var second_run=0] in 014_environment error: exit status 1
2026/07/23 09:08:32 Halting due to test '014_environment' failure: ...
```

**After** — the full run completes and ends with a summary (the verbose per-test log stream above it is unchanged):

```
==================== Integration tests summary ====================
PASS  003_full_environment
PASS  009_project
FAIL  014_environment
-------------------- failure details: 014_environment --------------------
environment 'my-env-14' (id: 4f2a1c9e-...) https://app.env0.com/p/<project-id>/environments/4f2a1c9e-...
╷
│ Error: failed to wait for environment destroy to complete, deployment status is: INTERNAL_FAILURE
│
╵
==================== 24/25 tests passed ====================
```

And the same summary is rendered on the workflow run's **Summary** page via `GITHUB_STEP_SUMMARY`:

> ## Integration tests summary
>
> | Test | Result |
> |---|---|
> | 003_full_environment | ✅ Pass |
> | 014_environment | ❌ Fail |
>
> ### Failures
>
> #### ❌ 014_environment
>
> - environment 'my-env-14' (id: 4f2a1c9e-...) https://app.env0.com/p/&lt;project-id&gt;/environments/4f2a1c9e-...
>
> ```
> ╷
> │ Error: failed to wait for environment destroy to complete, deployment status is: INTERNAL_FAILURE
> │
> ╵
> ```

### Validation

Validated end-to-end with a stubbed `tofu` binary: verified the console summary, environment link construction, state parsing (environments only, child modules included), the `GITHUB_STEP_SUMMARY` markdown, and exit codes for passing/failing/`DESTROY_ONLY` runs.

### DIFF
--- a/tests/harness.go
+++ b/tests/harness.go
@@ -1,11 +1,11 @@
 package main
 
 import (
-	"bufio"
 	"bytes"
 	"encoding/json"
 	"fmt"
 	"log"
+	"net/url"
 	"os"
 	"os/exec"
 	"path"
@@ -18,6 +18,13 @@ import (
 const TESTS_FOLDER = "tests/integration"
 
 const initMaxAttempts = 3
+
+type testResult struct {
+	name         string
+	passed       bool
+	err          error
+	environments []string
+}
 
 func main() {
 	if err := compileProvider(); err != nil {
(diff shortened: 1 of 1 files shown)


liranfarage89:
@yaronya what about adding some eye candies to the PR description? i.e before & after

liranfarage89 (commented):
Nice cleanup. The `bufio` writers were never flushed, so this also quietly fixes a real bug: up to the last 4KB of every command's output could get dropped, including the JSON that `output -json` parses. Running all tests to completion instead of `Fatalf`-ing on the first failure is the right call too, since the old path also skipped the other in-flight tests' cleanup destroys.

Left three comments below. One is worth verifying before merge (the dev UI-host mapping). Everything compiles, vets, and gofmt's clean.

tests/harness.go
```diff
@@ -138,13 +314,97 @@ func runTest(testName string, destroy bool) (bool, error) {
...
+
+	parsed, err := url.Parse(apiEndpoint)
+	if err != nil || parsed.Host == "" {
+		return ""
+	}
+
+	var host string
+
+	switch {
+	case strings.HasPrefix(parsed.Host, "api-"):
```
liranfarage89, line 399: 🟠 **Major** — CI runs against the `@ dev` org, so `ENV0_API_ENDPOINT` isn't `api.env0.com` and this `api-` branch is the one that actually fires. Stripping `api-` from something like `api-dev.env0.com` gives `dev.env0.com`. Is that the real UI host, or should it be `app-dev`/`console-dev`? Worth checking against the dev endpoint before merge, otherwise every link in the summary points nowhere. The prod `api.`→`app.` case looks right.

tests/harness.go
```diff
@@ -37,24 +44,185 @@ func main() {
...
+			inDiagnostic = true
+
+			highlights = append(highlights, line)
+		case strings.HasPrefix(trimmed, "╵"):
+			inDiagnostic = false
+
+			highlights = append(highlights, line)
+		case inDiagnostic:
+			highlights = append(highlights, line)
+		case strings.Contains(trimmed, "Error") && !strings.Contains(trimmed, "[INFO]"):
```
liranfarage89, line 149: 🟡 **Minor** — this only drops `[INFO]` lines, so `[WARN]`/`[DEBUG]`/`[ERROR]` provider logs and any info line with the word "Error" in it still land in the excerpt. The 100-line cap keeps it from blowing up, but skipping every `[LEVEL]`-prefixed provider line would keep the output tighter. Non-blocking.

tests/harness.go
```diff
@@ -138,13 +314,97 @@ func runTest(testName string, destroy bool) (bool, error) {
...
 		}
 	}
 
 	log.Println("Successfully finished running test ", testName)
 
-	return true, nil
+	return true, nil, nil
+}
+
+type stateModule struct {
```
liranfarage89, line 326: 🟢 **Nit** — inline anonymous struct is fine here. If the link ever needs the environment status or `template_id`, this is where you'd add it. No change needed now.

liranfarage89 (approved):

tests/harness.go
yaronya: Verified against the dev endpoint — the mapping is correct. Our dev API hosts are `api-<org>.<suffix>` and the app for the same org lives at `<org>.<suffix>` (same suffix, no `app` segment), so `TrimPrefix("api-")` lands exactly on the UI host. Prod is the `api.`→`app.` case. Confirmed the generated links resolve; just not pasting the concrete hosts here.

tests/harness.go
yaronya: Intentional — I only want to strip `[INFO]`. `[WARN]`/`[ERROR]` provider lines are frequently the actual cause of a failure (a provider-side error the terraform diagnostic block does not echo), so blanket-dropping every `[LEVEL]`-prefixed line risks filtering out the one line that explains the failure. The `[INFO]` carve-out exists only because info lines are pure noise that happen to contain the word "Error". Diagnostic blocks (╷…╵) are captured in full regardless of level, and the 100-line cap bounds the size — so I would rather over-include than hide the smoking gun.

PR #1121, opened 2026-08-02
[ENG-2115] Chore: Document workflow-level env0_configuration_variable example

odagadi:
[ENG-2115](https://linear.app/env-zero/issue/ENG-2115/docs-add-workflow-level-no-sub-environment-alias-example-for-env0) | [Artifacts](https://cloud.humanlayer.com/tasks/019fc1b9-c6ed-72d6-9251-365e55984e50/artifacts) | [Task](https://cloud.humanlayer.com/deep/tasks/019fc1b9-c6ed-72d6-9251-365e55984e50)

## What problems was I solving

The provider docs for `env0_configuration_variable` only ever showed the per-`sub_environment_alias` example. Setting `template_id` to the workflow template and leaving `sub_environment_alias` out already creates a workflow-level variable that applies to every sub environment, exactly what the UI's "entire workflow" tab produces. That behavior works today, it's just completely undocumented.

The result is customers concluding the capability doesn't exist. Bain hit this directly (split out of [PRD-746](https://linear.app/env-zero/issue/PRD-746)), and the same pattern showed up at Allstate, Visa, Pismo and Cars Commerce.

After this ships, someone reading the resource docs can see both scopes side by side and pick the right one without asking support. Success looks like: no more "is workflow-level variables supported in TF?" tickets from these accounts, and no follow-up doc requests on the same resource.

## What user-facing changes did I ship

- [examples/resources/env0_configuration_variable/resource.tf](https://github.com/env0/terraform-provider-env0/pull/1121/files#diff-add62ed9831a3971ae8b466bddfb098c65d39e1ff230141191808ead9f7b0f0e) - added a `workflow_example` resource showing the no-alias, workflow-scoped form, plus a comment on the existing `sub_environment_example` clarifying it's the sub environment scope
- [docs/resources/configuration_variable.md](https://github.com/env0/terraform-provider-env0/pull/1121/files#diff-eba6c64032ee8bd19d4344808b3df1a77459f3ec61b608733003454d54db186f) - the generated doc page, regenerated so the rendered example block matches

## How I implemented it

Docs only, no provider code touched.

### Example source
- [examples/resources/env0_configuration_variable/resource.tf](https://github.com/env0/terraform-provider-env0/pull/1121/files#diff-add62ed9831a3971ae8b466bddfb098c65d39e1ff230141191808ead9f7b0f0e) - dropped the new `env0_configuration_variable.workflow_example` block right above the existing sub environment one, so the two scopes read as a pair. The new block sets `name`, `value`, `template_id` and `description` and deliberately omits `sub_environment_alias`, since the omission *is* the thing being documented. Two leading comments carry the actual explanation: one on the workflow block ("set template_id to the workflow template and omit sub_environment_alias, applied to every sub environment"), one on the existing sub environment block ("set both, applied only to the matching sub environment"). The second comment is the small but important half, before this the reader had no signal that `sub_environment_alias` was optional at all.

### Generated docs
- [docs/resources/configuration_variable.md](https://github.com/env0/terraform-provider-env0/pull/1121/files#diff-eba6c64032ee8bd19d4344808b3df1a77459f3ec61b608733003454d54db186f) - mirrors the example file exactly. This page is produced by tfplugindocs from the `examples/` directory, so it's regenerated output rather than a hand edit. Worth confirming on review that CI's docs-generation step comes back clean and doesn't want to rewrite it.

## Deviations from the plan

No plan file found in the task directory, only the ticket. Nothing to compare against.

### Manual Testing
- [ ] Read the rendered example on the PR's Files changed tab and check the two blocks read as a clear pair of scopes
- [ ] Confirm `docs/resources/configuration_variable.md` matches `examples/resources/env0_configuration_variable/resource.tf` byte for byte in the example section
- [ ] Sanity check the claim itself against the API: apply an `env0_configuration_variable` with a workflow `template_id` and no `sub_environment_alias`, and confirm it shows up under the "entire workflow" tab in the UI
- [ ] Check the docs-generation CI job passes without producing a diff

### Automated Tests
```bash
git fetch origin eng-2115-docs-add-workflow-level-no-sub_environment_alias-example-for
git checkout eng-2115-docs-add-workflow-level-no-sub_environment_alias-example-for
go generate ./...
git diff --exit-code docs/
```

No unit or acceptance tests were added, there's no code path here to cover.

## Description for the changelog

Docs: document the workflow-level (no `sub_environment_alias`) form of `env0_configuration_variable`.


examples/resources/env0_configuration_variable/resource.tf
```diff
@@ -20,6 +20,17 @@ resource "env0_configuration_variable" "json_variable" {
   format = "JSON"
 }
 
+# Workflow scope: set template_id to the workflow template and omit sub_environment_alias.
```
liranfarage89: nit: "Workflow scope" is ambiguous, there are two. This writes `BLUEPRINT` on the workflow template (UI labels it "Template"). The separate `WORKFLOW` scope is keyed by the workflow *environment* id and comes from `env0_environment` + `sub_environment_configuration`, not this resource. Anyone who's used the environment page's "Entire Workflow" tab will read this as that one.

```suggestion
# Workflow template scope: set template_id to the workflow template and omit sub_environment_alias.
```

examples/resources/env0_configuration_variable/resource.tf
```diff
@@ -20,6 +20,17 @@ resource "env0_configuration_variable" "json_variable" {
...
+# The variable is applied to every sub environment of the workflow.
+resource "env0_configuration_variable" "workflow_example" {
+  name        = "WORKFLOW_VARIABLE"
+  value       = "workflow value"
+  template_id = "example-workflow-template-id"
+  description = "Variable for the entire workflow scope"
+}
+
+# Sub environment scope: set both template_id and sub_environment_alias.
+# The variable is applied only to the matching sub environment.
```
liranfarage89, line 33: Two scopes side by side, so the next question is which wins. `find-by-scope.ts` orders `SUB_ENVIRONMENT_BLUEPRINT` ahead of `BLUEPRINT`/`workflowBlueprintId` under "first items will override later items" — same-named sub env var beats the workflow-level one. Worth one more comment line here.

liranfarage89 (commented):
Traced the claim and it holds up: `template_id` with no alias lands on `client.ScopeTemplate` = `"BLUEPRINT"` ([resource_configuration_variable.go#L153](https://github.com/env0/terraform-provider-env0/blob/main/env0/resource_configuration_variable.go#L153), [client/configuration_variable.go#L12](https://github.com/env0/terraform-provider-env0/blob/main/client/configuration_variable.go#L12)), and env0/env0's `find-by-scope.ts` resolves `BLUEPRINT` against `workflowBlueprintId` when assembling a sub environment's config. Same scope the template page's "Entire Workflow" tab writes. So this documents real behavior.

Also: generated docs are in sync, you can tick that checkbox. The Docs job auto-commits when `generate-docs.sh` produces a diff, and the branch still has just your one commit.

Left two inline nits. One thing above nit level:

**🟠 The example fixes the wrong surface.** [`env0/resource_configuration_variable.go#L73`](https://github.com/env0/terraform-provider-env0/blob/main/env0/resource_configuration_variable.go#L73) still says `sub_environment_alias` = "create the variable for sub templates of a workflow. Requires template_id to be set.", and `template_id` says "create the variable under this template, not globally". Neither hints that omitting the alias scopes to the whole workflow. The Schema block is what people actually scroll to on the registry page, and it regenerates for free:

```go
Description: "create the variable for a specific sub environment of a workflow. Requires template_id to be set. Omit to apply the variable to every sub environment of the workflow template.",
```

**🟢 nit:** [`env0/data_configuration_variable.go#L70`](https://github.com/env0/terraform-provider-env0/blob/main/env0/data_configuration_variable.go#L70) has the same gap. If you're touching the resource description anyway, do both.

**🟢 nit:** Unit Tests is red — cancelled at the 24h timeout on 2026-08-02 (run `30741302754`). Nothing behind it on a docs-only change, but re-run before merge.

LGTM once the schema description is updated. Happy to call that a follow-up if you'd rather keep this examples-only.

liranfarage89 (approved):

liranfarage89 (approved):

PR #1123, opened 2026-08-16
[APO-443] Feat: Add tags attribute to env0_environment

liranfarage89:

Resolves [APO-443](https://linear.app/env-zero/issue/APO-443/terraform-provider-tags-attribute-on-env0-environment) — the Terraform provider slice of the [environment-level tags tech design](https://github.com/env0/env0/pull/21795) (§6). Customers who provision environments programmatically (ZoomInfo) need environment tags manageable as code.

Before this change there was no way to set or read environment tags from Terraform.

### Solution

`tags` map attribute on the `env0_environment` resource (Optional) and the `env0_environment` data source (Computed).

- **Create** — tags go out in the `POST /environments` body.
- **Update** — `PUT /environments/{id}/tags` with an RFC 7396 merge-patch. Keys dropped from the configuration are sent as `null` so they are removed; everything else is replaced wholesale. The endpoint is only called when `tags` actually changed.
- **Read** — tags come back on the environment payload and are written to state, so UI-side edits show as drift.

**Multi-value shape (the decision the ticket left to this PR):** the API models a tag as `key -> []values`. Terraform's `TypeMap` cannot hold lists, so values are comma-joined (`tags = { team = "eng,payments" }`) rather than modelled as a set of `{key, values}` blocks. The tag charset forbids commas, so the join is lossless, and it is the same convention the CloudQuery plugin uses for the `Map(String,String)` platform column. It also keeps the attribute idiomatic (`tags = { ... }`).

Validation (charset, 1–128 char keys, 1–256 char values, ≤ 50 pairs) is left to the API rather than duplicated in the provider.

**Sequencing:** the API landed on env0/env0 master on Aug 12, so the bors harness (`tf-provider-tests.yml`, running against `api-bors.dev.env0.com`) should have it. Worth confirming the integration run is green before merging.

- [x] I verified my changes work
- [x] I verified I didn't break anything

**That it works:**
- New `TestUnitEnvironmentTags` covers the full round trip: create with a multi-value tag (`team = "eng,payments"`) plus a single-value tag, then an update that changes one key, adds one and drops one — asserting the exact `EnvironmentUpdateTags` patch (`{"team":["eng"],"cost-center":["123"],"owner":null}`) and the comma-joined state values.
- New `EnvironmentUpdateTags` client spec asserts the `PUT /environments/{id}/tags` path and payload, including a `nil` value marshalling to `null`.
- `TestEnvironmentDataSource` now asserts `tags.team` / `tags.owner` come back joined from the API's arrays.
- Integration test `012_environment` sets tags on the `example` environment and changes them on the second run (add `cost-center`, remove `owner`, keep `team`), with a new `environment_tags_team` output read through the `env0_environment` data source — so create, merge-patch update and data-source read all run against a real backend.
- Verified the API contract against env0/env0 master: `EnvironmentTags = { [key: string]: string[] }`, `UpdateTags.Request.Body = { [key: string]: string[] | null }`, `tags` accepted on `CreateEnvironmentRequest` and validated in `functions/commons/create/create-environment.ts`.

**That I didn't break anything:**
- The reflection helpers (`readResourceData` / `writeResourceData`) have no `reflect.Map` case, so both `Tags` fields are tagged `tfschema:"-"` and handled explicitly — no other struct field changed behaviour.
- `getCreatePayload` only sets `Tags` when non-empty, so existing exact-payload `EnvironmentCreate` expectations (e.g. `TestUnitEnvironmentIsRequiredDeprecated`) still match.
- Checked every `setEnvironmentSchema` caller — the resource and the singular data source, both of which now declare `tags`. The plural `env0_environments` data source doesn't use it and is untouched.
- Full suites green: `go test ./env0/...` (195s), `go test ./client/...`, plus `gofmt -l .`, `go vet ./...`, `golangci-lint run ./...` (0 issues) and `terraform fmt -check`.

Docs were hand-edited to match what `tfplugindocs` emits (it can't run from a git worktree — it derives the provider name from the directory), and the docs workflow regenerates them on the PR anyway.

### DIFF
--- a/client/api_client.go
+++ b/client/api_client.go
@@ -74,6 +74,7 @@ type ApiClientInterface interface {
 	EnvironmentUpdate(id string, payload EnvironmentUpdate) (Environment, error)
 	EnvironmentDeploy(id string, payload DeployRequest) (EnvironmentDeployResponse, error)
 	EnvironmentUpdateTTL(id string, payload TTL) (Environment, error)
+	EnvironmentUpdateTags(id string, payload EnvironmentTags) (Environment, error)
 	EnvironmentMove(id string, projectId string) error
 	EnvironmentScheduling(environmentId string) (EnvironmentScheduling, error)
 	EnvironmentSchedulingUpdate(environmentId string, payload EnvironmentScheduling) (EnvironmentScheduling, error)
--- a/client/api_client_mock.go
+++ b/client/api_client_mock.go
@@ -1145,6 +1145,21 @@ func (mr *MockApiClientInterfaceMockRecorder) EnvironmentUpdateTTL(arg0, arg1 an
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "EnvironmentUpdateTTL", reflect.TypeOf((*MockApiClientInterface)(nil).EnvironmentUpdateTTL), arg0, arg1)
 }
 
+// EnvironmentUpdateTags mocks base method.
+func (m *MockApiClientInterface) EnvironmentUpdateTags(arg0 string, arg1 EnvironmentTags) (Environment, error) {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "EnvironmentUpdateTags", arg0, arg1)
+	ret0, _ := ret[0].(Environment)
+	ret1, _ := ret[1].(error)
+	return ret0, ret1
+}
+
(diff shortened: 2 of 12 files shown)


alonnoga (approved):
Reviewed the create, update, read, and merge-patch behavior against the backend contract and test coverage. No findings.

PR #1125, opened 2026-08-17
[APO-551] Fix: reject tag values padded around the comma separator

liranfarage89:
Closes APO-551. Found during the APO-495 Terraform-provider QA.

`env0_environment` `tags` values are comma-joined lists of a key's values. A space is legal *inside* a tag value, so `team = "eng, payments"` was stored verbatim as `["eng", " payments"]` — the apply succeeded, the re-plan was empty, and no error ever appeared, but the environment silently stopped matching a filter on `payments`, in both the env0 API and CloudQuery.

Trimming on the write path alone would swap one bug for another: the server returns the trimmed values, read writes them back to state, and config never matches state again. So this rejects the input at plan time instead — along with empty entries (`"a,,b"`), which until now failed as a raw 400 at apply time. A space *inside* a value stays legal.

Also adds the missing `tags` example to the resource docs, which had none.

## Tags created outside terraform

Padding is legal server-side, so a tag created in the UI or via the API can already have it. Worth being explicit about what this validator does and does not change there, since it is the one behaviour reviewers are most likely to ask about:

- **Import still works.** `terraform import` does not run config validation, so the environment imports and the padded value lands in state verbatim.
- **Config cannot restate the padding.** Writing `owner = " alice"` to match the imported state is rejected. This is deliberate: that value is invisible to `?tags=owner=alice`, so allowing it would keep the silent-filter-miss bug alive for single values.
- **The escape hatch is one apply.** Drop the padding in config and apply — the tag normalizes server-side and the re-plan is empty. Verified end to end.
- The error message says so, rather than talking about comma-joining when the value holds no comma.

The tradeoff: a user who genuinely wants a padded tag value cannot express it through this provider. That seems clearly right — a leading space in a tag value is not something anyone means, and the failure mode of allowing it is silent.

## Manual QA

Built from this branch and run with real Terraform against the dev stage.

Config authoring:

| `tags` in config | Before | After |
| --- | --- | --- |
| `team = "eng,payments"` | applies | applies (unchanged) |
| `note = "platform eng"` | applies | applies (space inside a value still legal) |
| `team = "eng, payments"` | applies, stores `" payments"` | rejected at plan |
| `team = "eng ,payments"` | applies, stores `"eng "` | rejected at plan |
| `owner = " alice"` | applies, stores `" alice"` | rejected at plan |
| `team = ""` | 400 at apply | rejected at plan |
| `team = "eng,,payments"` | 400 at apply | rejected at plan |

Import of an environment whose tags were set out-of-band to `{"squad":["eng"," payments"],"owner":[" alice"]}`:

| Step | Result |
| --- | --- |
| `terraform import` | succeeds; state holds `squad = "eng, payments"`, `owner = " alice"` |
| plan with config matching state | rejected, with the message pointing at normalization |
| plan with padding dropped | `~ "owner" = " alice" -> "alice"`, `~ "squad" = "eng, payments" -> "eng,payments"` |
| apply, then re-plan | converges in one apply; server holds `{"owner":["alice"],"squad":["eng","payments"]}`; re-plan empty |

Confirmed the original impact on the dev stage before fixing: with `["eng", " payments"]` stored, `GET /environments?tags=squad=payments` returned no match while `tags=squad=%20payments` matched.

Full suite green (`go test ./...`). Docs regenerated with `tfplugindocs`.

### DIFF
--- a/docs/resources/environment.md
+++ b/docs/resources/environment.md
@@ -27,6 +27,19 @@ resource "env0_environment" "example" {
   template_id = data.env0_template.example.id
 }
 
+resource "env0_environment" "example_with_tags" {
+  name        = "environment with tags"
+  project_id  = data.env0_project.default_project.id
+  template_id = data.env0_template.example.id
+
+  tags = {
+    owner         = "alice"
+    "cost-center" = "1234"
+    # a key may hold several values - join them with a comma and no space
+    team = "eng,payments"
+  }
+}
+
 resource "env0_environment" "example_with_hcl_configuration" {
   name        = "environment with hcl"
   project_id  = data.env0_project.default_project.id
@@ -80,7 +93,7 @@ If true must specify one of the following - 'github_installation_id' if using Gi
 If true must specify one of the following - 'github_installation_id' if using GitHub, 'gitlab_project_id' and 'token_id' if using GitLab, or 'bitbucket_client_key' if using BitBucket.
 - `sub_environment_configuration` (Block List) the subenvironments for a workflow environment. Template type must be 'workflow'. Must match the configuration as defined in 'env0.workflow.yml' (see [below for nested schema](#nestedblock--sub_environment_configuration))
 - `tags` (Map of String) the environment's tags. Keys and values may contain letters, digits, spaces and the characters _ . : / + - @ (up to 50 key-value pairs).
-A key may hold several values - join them with a comma (for example: "eng,payments").
+A key may hold several values - join them with a comma and no space (for example: "eng,payments").
 - `template_id` (String) the template id the environment is to be created from.
 Important note: the template must first be assigned to the same project as the environment (project_id). Use 'env0_template_project_assignment' to assign the template to the project. In addition, be sure to leverage 'depends_on' if applicable. Please note that changing this attribute will require environment redeploy
 - `terragrunt_working_directory` (String) The working directory path to be used by a Terragrunt template. If left empty '/' is used. Note: modifying this field destroys the current environment and creates a new one
--- a/env0/resource_environment.go
+++ b/env0/resource_environment.go
@@ -240,11 +240,12 @@ func resourceEnvironment() *schema.Resource {
 			},
 			"tags": {
 				Type:        schema.TypeMap,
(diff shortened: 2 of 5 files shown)


Yossi-kerner (approved):
Approach looks right to me — rejecting at plan instead of trimming on write avoids the permanent config-vs-state diff, and the import/out-of-band section answers the question I was going to ask. Verified locally: full suite green, `terraform fmt -check -recursive examples/` and `gofmt` clean. (The red "Unit Tests" check is a 429 downloading `actions/setup-go`, not a real failure — worth a re-run.)

One thing worth a look, non-blocking:

**Padded tag *keys* are still accepted** — `env0/validators.go` only inspects values, and `getTagsFromSchema` (`env0/resource_environment.go:546`) sends keys verbatim. Since the attribute description says keys may contain spaces, `tags = { " owner" = "alice" }` creates the key `" owner"` server-side, `?tags=owner=alice` misses it, and `tagsToSchema` returns the padded key unchanged so the state round-trips with no drift — the same silent failure this PR is fixing for values. Assuming the API preserves key padding the way you verified it preserves value padding, a `strings.TrimSpace(key) != key` check is a couple of lines. Fine as a follow-up if it's outside APO-551.

liranfarage89:
Good catch — fixed in 2b46d5d. Keys now go through the same rule (`" owner"`, `"owner "` rejected; a space *inside* a key stays legal), with three test cases added. Suite green.

Checked the backend assumption you named, and padded keys turn out to be worse than padded values. Nothing normalizes on write: `TAG_CHARSET` includes a space and the key check is length-only (`environment-service/functions/commons/tags.ts`). But the filter parser trims each whole `key=value` segment (`common/lambda/src/commons/parse-comma-separated-param.ts`), so a key stored as `" owner"` cannot be matched by *any* filter — `?tags=%20owner=alice` looks up `owner`. It also surfaces as a separate key in `GET /environments/tags` and consumes one of the 50 pairs. Unfilterable rather than just easy to miss, so it belongs in APO-551 rather than a follow-up.

The one-apply escape hatch holds for keys too: `updateTags` already sends keys dropped from config as `null` for the merge-patch, so the padded key is removed and the trimmed one added in the same apply.

Unit Tests is re-running on this push — agreed the earlier red was the `actions/setup-go` 429.

PR #1127, opened 2026-08-18
[ATL-305] Fix: back off properly on 429 responses

alonnoga:
Linear: [ATL-305](https://linear.app/env-zero/issue/ATL-305/change-provider-retry-behavior-on-429-responses)

## Problem

WAF rate limits blocked tens of thousands of provider requests (57k on `/configuration` alone in 7 days). The provider already has a client-side rate limiter, but its 429 handling made a blocked window worse instead of letting it drain:

1. **Retries bypassed the rate limiter.** `Wait()` ran in `HttpClient.request()`, i.e. once per logical request. Resty's retry loop re-executes the request without going through it, so retries were sent on top of the allowed rate — exactly when the server was already pushing back.
2. **A 429 was retried like a 5xx** (1s base, 30s cap). A limit measured over a minute-sized window doesn't clear in a second, so early attempts were spent on requests that were certain to be blocked.
3. **`Retry-After` was ignored.**
4. **Only the blocked request backed off.** With Terraform's default parallelism of 10, the siblings kept firing into the closed window.

## Changes

- **Rate limiting moved into the resty middleware** (`OnBeforeRequest`), so every *attempt* — retries included — waits for a slot.
- **A 429 pauses the whole client.** New `RateLimiter.Pause(d)`: on 429 the limiter holds back every request for `Retry-After`, or 5s when the header is absent. A pause doesn't consume the window budget and is never shortened by a later, smaller pause.
- **429 gets its own retry ladder**: 5s base, jittered over `[wait/2, wait)`, capped at the same 30s max wait as before; `Retry-After` is honored (plus jitter) when present, and a `Retry-After` beyond the cap is still waited out in full by the limiter pause. 5xx and networking failures keep the previous 1s..30s bounds unchanged.
- **The pause wake-up is spread.** A pause is one shared deadline, so waiters would otherwise all resume in the same instant; `nextAvailable` adds a bounded random spread.
- Rate limits themselves are unchanged — still 950 requests/minute.

## Tests

- `ParseRetryAfter`: seconds, HTTP-date, missing/unparsable/past values.
- Limiter `Pause`: blocks, preserves window budget, extends but never shortens, `Wait` blocks and respects context cancellation.
- HTTP client: retried attempts count against the limit; a 429 holds back requests that hadn't been sent yet, until `Retry-After` expires.
- Retry ladder: 429 is retried through the full ladder, always waits longer than a 5xx on the same attempt, honors `Retry-After`, and is jittered.

Each new test was checked to fail without its corresponding change.

## Manual QA (env0 dev, org `TF-provider-integration-tests`)

Provider built from this branch and driven by the real integration harness (`go run tests/harness.go`) against `api-dev.dev.env0.com`.

**Happy flow — unchanged.** `002_project`, `003_configuration_variable`, `009_team`, `020_api_key` → **4/4 passed** (apply → second apply → outputs → destroy). Moving the limiter into the resty middleware does not break auth, pacing or any CRUD path.

**429 flow, with `Retry-After: 2`.** A local proxy in front of the real API answered 429 on every 4th request; `002_project` ran through it: **154 requests, 38 of them 429 → test passed** (~2 min). Proxy timeline shows the wait is the header value, and that the pause applies to the whole client, not just the blocked request:

```
23.216s 429  GET /organizations
25.397s pass GET /organizations                  <- retried after 2.18s (Retry-After + jitter)
25.645s 429  POST /projects
27.969s pass POST /projects                       <- everything held 2.32s
29.631s 429  GET /projects/7ee7e45a...
31.797s pass GET /projects/7ee7e45a...
31.971s 429  POST /projects
34.059s pass PUT  /costs/project/.../budget       <- unrelated request, same pause
```

**429 flow, no `Retry-After` header** (the WAF case). Same setup, header omitted, `009_team`: 30 requests, 3 × 429 → **test passed**, and each 429 held every request for the 5s default plus the wake-up spread:

```
27.142s 429  GET /organizations
32.165s pass GET /organizations                   <- +5.02s
38.225s 429  GET /teams/organizations/...
43.696s pass GET /teams/organizations/...         <- +5.47s
45.846s 429  GET /organizations
51.202s pass GET /organizations                   <- +5.36s
```

All three runs ended with a successful `tofu destroy`; no leftover resources. The injecting proxy was throwaway QA scaffolding and is not part of this PR.

Both 429 scenarios were re-run after the review fixes (shared 30s cap, pause wake-up spread); numbers above are from the re-run.

### DIFF
--- a/client/http/client.go
+++ b/client/http/client.go
@@ -11,11 +11,12 @@ import (
 	"github.com/go-resty/resty/v2"
 )
 
-// How long the rate limiter holds back every request after a 429 that carries no Retry-After
-// header. The limit is enforced server side (API gateway and WAF), so the whole client has to
-// idle: retrying only the blocked request while its siblings keep firing is what turns a single
-// 429 into thousands of blocked requests.
-const defaultRateLimitPause = 5 * time.Second
+// DefaultRateLimitPause is how long the rate limiter holds back every request after a 429 that
+// carries no Retry-After header. The limit is enforced server side (API gateway and WAF), so the
+// whole client has to idle: retrying only the blocked request while its siblings keep firing is
+// what turns a single 429 into thousands of blocked requests. The provider's 429 backoff starts
+// from the same value, so the two can't drift apart.
+const DefaultRateLimitPause = 5 * time.Second
 
 type HttpClientInterface interface {
 	Get(path string, params map[string]string, response any) error
@@ -60,6 +61,9 @@ func NewHttpClient(config HttpClientConfig) (*HttpClient, error) {
 // applyRateLimiter wires the limiter into the resty middleware rather than into HttpClient's
 // request builder, so that retried attempts count against the limit as well. Otherwise a run
 // that starts getting 429s or 5xx answers sends the retries on top of the allowed rate.
+//
+// The hooks are appended to the resty client, so it must not be handed to NewHttpClient twice:
+// a second call would register a second Wait and halve the effective rate.
 func applyRateLimiter(client *resty.Client, rateLimiter ratelimiter.RateLimiter) {
 	client.OnBeforeRequest(func(c *resty.Client, r *resty.Request) error {
 		return rateLimiter.Wait(r.Context())
@@ -72,7 +76,7 @@ func applyRateLimiter(client *resty.Client, rateLimiter ratelimiter.RateLimiter)
 
 		pause, ok := ParseRetryAfter(r.Header().Get("Retry-After"))
 		if !ok {
-			pause = defaultRateLimitPause
+			pause = DefaultRateLimitPause
 		}
 
 		rateLimiter.Pause(pause)
--- a/client/http/client.go
+++ b/client/http/client.go
@@ -3,12 +3,19 @@ package http
 //go:generate mockgen -destination=client_mock.go -package=http . HttpClientInterface
 
 import (
-	"context"
+	"net/http"
 	"reflect"
+	"time"
 
 	"github.com/env0/terraform-provider-env0/client/http/ratelimiter"
 	"github.com/go-resty/resty/v2"
 )
+
+// How long the rate limiter holds back every request after a 429 that carries no Retry-After
+// header. The limit is enforced server side (API gateway and WAF), so the whole client has to
+// idle: retrying only the blocked request while its siblings keep firing is what turns a single
+// 429 into thousands of blocked requests.
+const defaultRateLimitPause = 5 * time.Second
 
 type HttpClientInterface interface {
 	Get(path string, params map[string]string, response any) error
@@ -19,11 +26,10 @@ type HttpClientInterface interface {
 }
 
 type HttpClient struct {
-	ApiKey      string
-	ApiSecret   string
-	Endpoint    string
-	client      *resty.Client
-	rateLimiter *ratelimiter.RateLimiter
+	ApiKey    string
+	ApiSecret string
+	Endpoint  string
+	client    *resty.Client
 }
 
 type HttpClientConfig struct {
@@ -36,26 +42,46 @@ type HttpClientConfig struct {
 }
 
 func NewHttpClient(config HttpClientConfig) (*HttpClient, error) {
+	restClient := config.RestClient.SetBaseURL(config.ApiEndpoint).SetHeader("User-Agent", config.UserAgent)
+
+	if config.RateLimiter != nil {
+		applyRateLimiter(restClient, config.RateLimiter)
+	}
+
 	httpClient := &HttpClient{
-		ApiKey:      config.ApiKey,
-		ApiSecret:   config.ApiSecret,
-		client:      config.RestClient.SetBaseURL(config.ApiEndpoint).SetHeader("User-Agent", config.UserAgent),
-		rateLimiter: &config.RateLimiter,
+		ApiKey:    config.ApiKey,
+		ApiSecret: config.ApiSecret,
+		client:    restClient,
 	}
 
 	return httpClient, nil
 }
 
-func (client *HttpClient) request() *resty.Request {
-	if *client.rateLimiter != nil {
-		ctx := context.Background()
+// applyRateLimiter wires the limiter into the resty middleware rather than into HttpClient's
+// request builder, so that retried attempts count against the limit as well. Otherwise a run
+// that starts getting 429s or 5xx answers sends the retries on top of the allowed rate.
+func applyRateLimiter(client *resty.Client, rateLimiter ratelimiter.RateLimiter) {
+	client.OnBeforeRequest(func(c *resty.Client, r *resty.Request) error {
+		return rateLimiter.Wait(r.Context())
+	})
 
-		err := (*client.rateLimiter).Wait(ctx)
-		if err != nil {
-			return client.client.R().SetError(err)
+	client.OnAfterResponse(func(c *resty.Client, r *resty.Response) error {
+		if r.StatusCode() != http.StatusTooManyRequests {
+			return nil
 		}
-	}
 
+		pause, ok := ParseRetryAfter(r.Header().Get("Retry-After"))
+		if !ok {
+			pause = defaultRateLimitPause
+		}
+
+		rateLimiter.Pause(pause)
+
+		return nil
+	})
+}
+
+func (client *HttpClient) request() *resty.Request {
 	return client.client.R().SetBasicAuth(client.ApiKey, client.ApiSecret)
 }
 
--- a/client/http/rate_limiter_test.go
+++ b/client/http/rate_limiter_test.go
@@ -82,9 +82,13 @@ var _ = Describe("SlidingWindow Rate Limiter", func() {
 
 			httpClient = createClient(2, 200*time.Millisecond)
 
-			var response string
+			done := make(chan struct{})
 
 			go func() {
+				defer close(done)
+
+				var response string
+
 				_ = httpClient.Get(path, nil, &response)
 			}()
 
@@ -94,7 +98,7 @@ var _ = Describe("SlidingWindow Rate Limiter", func() {
 			callCount := httpmock.GetCallCountInfo()
 			Expect(callCount["GET "+BaseUrl+path]).To(Equal(2))
 
-			time.Sleep(300 * time.Millisecond)
+			Eventually(done, 3*time.Second).Should(BeClosed())
 
 			callCount = httpmock.GetCallCountInfo()
 			Expect(callCount["GET "+BaseUrl+path]).To(Equal(4))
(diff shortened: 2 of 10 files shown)


env0/provider.go
```diff
@@ -200,8 +263,12 @@ func createRestyClient(ctx context.Context) *resty.Client {
 	subCtx := tflog.NewSubsystem(ctx, "env0_api_client")
 
 	return resty.New().SetRetryCount(retryCount).
-		SetRetryWaitTime(time.Second).
-		SetRetryMaxWaitTime(time.Second * 30).
+		SetRetryWaitTime(retryWaitTime).
+		// The upper bound of both ladders: retryWaitTimeFor caps the non-rate-limit one itself.
+		SetRetryMaxWaitTime(rateLimitMaxWaitTime).
```
liranfarage89: Networking failures silently went from a 30s cap to 60s. When `resp == nil` resty never calls your `RetryAfter` func — `sleepDuration` returns `jitterBackoff(min, max, attempt)` first ([retry.go:169](https://github.com/go-resty/resty/blob/v2.17.2/retry.go#L169)) with `max` = this value. So `retryMaxWaitTime` only applies when there's a response, and the PR body's "networking failures keep the previous 1s..30s bounds" isn't true.

Simplest fix is to accept 60s for everything and delete `retryMaxWaitTime` plus the non-429 branch of `retryWaitTimeFor`.

client/http/ratelimiter/sliding_window_limiter.go
```diff
@@ -101,12 +124,17 @@ func (l *SlidingWindowLimiter) nextAvailable() time.Duration {
...
+
+	if len(l.requests) >= l.maxRequests {
+		// Wait for the oldest request to expire
+		oldest := l.requests[0]
+		delay = oldest.Add(l.window).Sub(now)
 	}
 
-	// Wait for the oldest request to expire
-	oldest := l.requests[0]
+	if pause := l.pausedUntil.Sub(now); pause > delay {
```
liranfarage89, line 144: `pausedUntil` is a single shared deadline, so every waiter wakes at the same instant — I measured 80µs spread across 10 goroutines on a 300ms pause. In the no-`Retry-After` case the pause (5s) outlasts the first 429 backoff (`[2.5s, 5s)`), so `Wait` re-synchronizes exactly the requests the jitter was meant to spread.

Bounded by TF parallelism so it's not what got us blocked, but it does defeat the stated goal. Consider a small random spread here.

env0/provider.go
```diff
@@ -183,13 +184,75 @@ func Provider(version string) plugin.ProviderFunc {
...
+		// keep them from coming back as a single burst the moment it expires.
+		return retryAfter + jitter(retryAfter/4)
+	}
+
+	return jitterBackoff(rateLimitWaitTime, rateLimitMaxWaitTime, exponent)
+}
+
+// jitterBackoff doubles base once per previous attempt, up to maxWait, and randomizes the result
+// over [wait/2, wait) so requests that failed together don't retry in lockstep.
+func jitterBackoff(base, maxWait time.Duration, exponent int) time.Duration {
```
liranfarage89, line 231: nit: the first 5xx retry loses its jitter. This returns `[0.5s, 1s)` for exponent 0, then resty clamps `if result < min { result = min }` with `min` = `retryWaitTime` = 1s, so it's always exactly 1s. `TestRetryWaitTimeFor` asserts on the unclamped value so it can't catch it.

env0/provider.go
```diff
@@ -183,13 +184,75 @@ func Provider(version string) plugin.ProviderFunc {
 }
 
 const (
+	// Same value as net/http's StatusTooManyRequests, spelled out because importing net/http
+	// here would shadow the provider's own http client package.
+	tooManyRequests = 429
```
liranfarage89: nit: `nethttp "net/http"` is one line and gives you `nethttp.StatusTooManyRequests`. The constant plus its shadowing comment costs more than the alias would.

client/http/client.go
```diff
@@ -3,13 +3,20 @@ package http
...
 
 	"github.com/env0/terraform-provider-env0/client/http/ratelimiter"
 	"github.com/go-resty/resty/v2"
 )
 
+// How long the rate limiter holds back every request after a 429 that carries no Retry-After
+// header. The limit is enforced server side (API gateway and WAF), so the whole client has to
+// idle: retrying only the blocked request while its siblings keep firing is what turns a single
+// 429 into thousands of blocked requests.
+const defaultRateLimitPause = 5 * time.Second
```
liranfarage89: nit: this and `rateLimitWaitTime` in `provider.go` are the same 5s policy in two packages, free to drift apart.

client/http/client.go
```diff
@@ -36,26 +42,46 @@ type HttpClientConfig struct {
...
 	return httpClient, nil
 }
 
-func (client *HttpClient) request() *resty.Request {
-	if *client.rateLimiter != nil {
-		ctx := context.Background()
+// applyRateLimiter wires the limiter into the resty middleware rather than into HttpClient's
+// request builder, so that retried attempts count against the limit as well. Otherwise a run
+// that starts getting 429s or 5xx answers sends the retries on top of the allowed rate.
+func applyRateLimiter(client *resty.Client, rateLimiter ratelimiter.RateLimiter) {
```
liranfarage89, line 67: nit: this appends to the caller's resty client, so a second `NewHttpClient` on the same client double-registers `Wait` and halves throughput. One call site today — worth a line in the doc comment.

client/http/rate_limiter_test.go
```diff
@@ -63,6 +64,78 @@ var _ = Describe("SlidingWindow Rate Limiter", func() {
...
+					res.Header.Set("Retry-After", "1")
+
+					return res, nil
+				})
+
+			httpClient = createClient(10, time.Minute)
+
+			var response string
+
+			Expect(httpClient.Get(path, nil, &response)).ToNot(BeNil())
```
liranfarage89: nit: `HaveOccurred()` reads better than `ToNot(BeNil())` for an error.

liranfarage89 (approved):
LGTM, approving. Diagnosis holds up — I checked resty v2.17.2 and `executeBefore` does run `udBeforeRequest` on every attempt (`client.go:1204`), and `afterResponse` runs per attempt too, so both halves of the fix land where you intended. Pausing the shared limiter is the right lever for a server-side limit. Nice side effect: dropping the `rateLimiter` field also kills the old `SetError(err)` bug, where a `Wait` failure set the error *result struct* and the request went out anyway.

Ran the new tests 5x under `-race` plus `-cpu=1` — stable. `golangci-lint`: 0 issues.

Left a few comments, only the first is worth changing before merge.

Two questions:

1. `r.Context()` is always `context.Background()` — nothing in the repo calls `SetContext` (grepped, zero hits). So the `Wait` cancellation path and the "respect context cancellation while paused" spec cover something production can't reach, and a paused client can't be interrupted. Same as before the PR, just worth knowing.
2. Worst case per request roughly doubles: the old 5xx ladder summed to ~2.3 min expected, the 429 ladder (5s→60s over 10 retries) is ~5.5 min, plus pause time, with nothing shown in the TF output meanwhile. Still want `retryCount = 10` on the 429 path?

env0/provider.go
alonnoga: Confirmed and fixed in dc8108b. `sleepDuration` returns `jitterBackoff(min, max, attempt)` before it ever looks at `RetryAfter` when `resp == nil`, so the client-level max was the only thing governing networking failures — the PR body claim was wrong.

Went the other way from your suggestion though: instead of accepting 60s everywhere, both ladders now share `retryMaxWaitTime` (30s), so 5xx and networking retries are byte-for-byte what they were before this branch. The 60s cap was only there so a large `Retry-After` could be waited out inside resty, and that is already covered better by the limiter pause, which holds the *whole client* for the full header value with no cap. So nothing is lost, and the blast radius of the PR shrinks. Kept the two branches since the 429 base (5s) still differs from the 5xx base (1s).

client/http/ratelimiter/sliding_window_limiter.go
alonnoga: Fair, and fixed in dc8108b: `nextAvailable` now adds a bounded random spread to the pause wait — 25% of the remaining pause, capped at 1s — so waiters do not resume on one shared edge. New spec drives 10 goroutines through a 200ms pause and asserts the wake-ups are not simultaneous.

Re-ran the proxy QA after the change; both scenarios still pass, and the no-header gaps now read 5.02s / 5.47s / 5.36s instead of a flat 5.002s.

env0/provider.go
alonnoga: True, and deliberate: `jitterBackoff(1s, 30s, 0)` returns `[0.5s, 1s)` and resty clamps it up to `RetryWaitTime`, so the first non-429 wait is exactly 1s. That is also exactly what resty's own default backoff did before this function existed, so the alternative here is changing 5xx timing, which this PR is trying not to do. Added a comment on `retryWaitTimeFor` recording the clamp so the next reader does not think the jitter is live.

env0/provider.go
alonnoga: Done in dc8108b — `nethttp "net/http"`, constant and comment deleted.

client/http/client.go
alonnoga: Fixed in dc8108b: exported as `http.DefaultRateLimitPause`, and `provider.go` derives `rateLimitWaitTime` from it, so the pause and the 429 backoff base cannot drift.

client/http/client.go
alonnoga: Documented in dc8108b: the doc comment now says the hooks are appended and the same resty client must not be handed to `NewHttpClient` twice, since a second `Wait` would halve the rate.

client/http/rate_limiter_test.go
alonnoga: Fixed in dc8108b.

alonnoga:
Thanks — all seven addressed in dc8108b (details in each thread). Answering the two questions:

**1. `r.Context()` is always `context.Background()`.** Correct, and worth stating in the PR rather than leaving implied: nothing in the repo calls `SetContext`, so the `Wait` cancellation path is unreachable in production and a paused client cannot be interrupted. That is unchanged by this PR — the old `HttpClient.request()` passed `context.Background()` explicitly. Threading Terraform's per-operation context through `client.ApiClient` into every call is a real change with a wide diff (every method signature), so I left it out of a fix meant to be backportable. The waits are all bounded (pause ≤ `Retry-After`, window ≤ 1 min, ladder ≤ retryCount × 30s), so a hang is not reachable either — it is a responsiveness gap, not a liveness one. Happy to open a follow-up ticket for the context plumbing; the spec stays as a guard for when that lands.

**2. `retryCount = 10` on the 429 path.** Keeping it, and the number is smaller than your estimate now that both ladders share the 30s cap: expected total sleep for a request that gets 429 on all 10 retries is ~3.1 min (3.75 + 7.5 + 15 + 22.5 × 7), against ~2.3 min for the old 5xx ladder — roughly 1.3×, not 2×. Plus pause time, which is bounded by whatever the server asks for.

The trade is a slow apply versus a failed apply, and a failed apply is the expensive one here: it drops Terraform mid-graph and the retry is a whole new run that re-reads everything, i.e. more load on the limit that just rejected us. WAF blocks are also bursty (57k blocked requests in 7 days concentrated in a couple of IPs), so the tail attempts are exactly the ones that get through. On visibility you are right that the TF output says nothing — `TF_LOG_PROVIDER=warn` now logs attempt number and `Retry-After` per 429, which is the same channel the empty-list and 404 retries already use.

PR #1129, opened 2026-08-30
[APO-684] Fix: Clear the module opentofu version when module tests are disabled

yarivg:

Linear: [APO-684](https://linear.app/env-zero/issue/APO-684/env0-module-opentofu-version-never-clears-after-disabling-module)

`ModuleUpdatePayload.OpentofuVersion` was tagged `omitempty`, so an empty version is stripped from the PATCH body. The backend merges the request over the existing module, so a missing key leaves the old version in place, and `disableModuleTesting` never clears it.

The apply succeeds and writes `""` to state. The next refresh reads `1.6.0` back, so the module drifts on every plan.

Reproduce: create a module with `module_test_enabled = true` and `opentofu_version = "1.6.0"`, then disable tests, drop the version, apply, and plan again.

### Solution

Dropped `omitempty` from `ModuleUpdatePayload.OpentofuVersion`, so an empty version is sent and the backend overwrite clears it. The API accepts it: `BlueprintApi.OpentofuVersion` is a plain string with no enum or `minLength`, and `assertTofuVersion` is skipped on a falsy value.

`ModuleCreatePayload` keeps `omitempty`. There is nothing to clear on create.

Tests cover both layers: the client test asserts `opentofuVersion` survives JSON marshalling when empty, and the resource test asserts the update payload carries an empty version when module tests are disabled.

---

- [x] Reproduced the bug against the dev stage (`api-dev.dev.env0.com`), driving a provider binary built from the parent commit through a filesystem mirror. Created a module with tests enabled and `opentofu_version = "1.6.0"`, then applied with tests off and the version dropped: state showed `""`, `GET /modules/:id` still returned `"1.6.0"`, and every re-plan showed `- opentofu_version = "1.6.0" -> null`.

- [x] Pointed the same stuck module at a binary built from this branch. One apply cleared it: the server dropped `opentofuVersion` and the re-plan returned `No changes`. Users already drifted recover without touching state.

- [x] Ran a fresh cycle on this branch: create with `1.6.0` (server `"1.6.0"`), disable tests (server key gone, plan clean), re-enable with `1.7.0` (server `"1.7.0"`, plan clean). Setting a version still works, so the fix does not trade one direction for the other.

- [x] Destroyed both modules and confirmed through the API that no QA modules and no module-testing environments were left in the org.

Ran with `terraform`, not `tofu`, which is not installed locally. The provider binary under test is the same either way.

### DIFF
--- a/client/module.go
+++ b/client/module.go
@@ -77,6 +77,8 @@ type ModuleCreatePayloadWith struct {
 	OrganizationId string `json:"organizationId"`
 }
 
+// OpentofuVersion is deliberately not 'omitempty': an empty version must be sent so that the backend clears a
+// previously set version - for example when module tests are disabled.
 type ModuleUpdatePayload struct {
 	ModuleName            string         `json:"moduleName,omitempty"`
 	ModuleProvider        string         `json:"moduleProvider,omitempty"`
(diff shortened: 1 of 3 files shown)


Yossi-kerner (approved):
LGTM

PR #1131, opened 2026-09-03
[APO-717] Feature: expose enable_post_merge_pr_comments on env0_organization_policy

yarivg:

Linear: [APO-717](https://linear.app/env-zero/issue/APO-717)

`enablePostMergePrComments` is a new organization policy on the env0 API: after a PR is merged, env0 comments on it for each environment whose deployment is waiting for approval or failed. The provider does not expose it, so an organization that manages its policies in Terraform cannot turn it on, and a plan would drift against it.

### Solution

Adds `enable_post_merge_pr_comments` (Boolean, optional, defaults to `false`) to `env0_organization_policy`, plus the matching fields on the client `Organization` and `OrganizationPolicyUpdatePayload`. Docs regenerated with `./generate-docs.sh`.

The API rejects enabling it together with the organization's Disable PR Comments policy (that one is not exposed by the provider), so the description says so and the error surfaces from the API.

Covered by the existing `TestUnitOrganizationPolicyResource` create/update/delete flow, extended with the new attribute:

```
go test ./env0/ -run TestUnitOrganizationPolicyResource -count=1   # ok
```

### DIFF
--- a/client/organization.go
+++ b/client/organization.go
@@ -15,6 +15,7 @@ type Organization struct {
 	EnableOidc                          bool    `json:"enableOidc"`
 	EnforcePrCommenterPermissions       bool    `json:"enforcePrCommenterPermissions"`
 	AllowMergeableBypassForPrApply      bool    `json:"allowMergeableBypassForPrApply"`
+	EnablePostMergePrComments           bool    `json:"enablePostMergePrComments"`
 	Description                         string  `json:"description"`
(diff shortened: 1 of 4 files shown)


yarivg:
Closing unmerged. Direction changed on the env0 side: the feature this policy gated is being implemented through the merge commit's check instead of a post-merge PR comment, so `enablePostMergePrComments` is being removed from the API rather than kept. Nothing for the provider to expose.

PR #1133, opened 2026-09-15
[APO-750] Chore: Document that a failed env0_environment update leaves a partial update

dcelasun:

Fixes https://linear.app/env-zero/issue/APO-750/document-that-a-failed-env0-environment-update-leaves-a-partial-update

`resourceEnvironmentUpdate` runs up to seven env0 API calls in a fixed order, and each one returns on error, so a failure part-way through leaves the earlier calls applied and the later ones not; the resource had no description saying so.

### Solution

Added a top-level `Description` to `resourceEnvironment` that states the fixed call order with no rollback, and that the apply returns once the deployment is queued, then regenerated `docs/resources/environment.md`.

---

- [x] Regenerated the docs with `tfplugindocs generate --provider-name terraform-provider-env0` and confirmed `docs/resources/environment.md` is the only doc file changed, with both notes in the frontmatter and as two paragraphs under the resource heading.

- [x] Ran `go build ./...` and `go vet ./env0/`.

No runtime behaviour changes, so there is nothing to exercise against a stage.


env0/resource_environment.go
```diff
@@ -163,6 +163,9 @@ func resourceEnvironment() *schema.Resource {
 	}
 
 	return &schema.Resource{
+		Description: "Note: updating an environment issues several env0 API calls in a fixed order - project move, template update, environment update, TTL, drift detection, tags, and finally the deployment - and there is no rollback. If one of them fails, the changes before it are already applied and the ones after it are not; the next plan re-diffs whatever is left and can be applied again.\n\n" +
```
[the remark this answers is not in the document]
dcelasun: Correct, fixed in 00d1d26. The last step now reads "and last a deployment (or, with 'prevent_auto_deploy', direct writes of 'configuration', 'sub_environment_configuration' and 'variable_sets')", and the note says each step runs only when its own fields changed.

env0/resource_environment.go
```diff
@@ -163,6 +163,9 @@ func resourceEnvironment() *schema.Resource {
...
+			"Note: the provider does not wait for the deployment it triggers. The apply returns once the deployment is queued, so the state holds the new 'revision' and 'configuration' while env0 may still be running the previous ones, or while that deployment fails. The next refresh reads back what is actually deployed and the plan re-diffs it.",
```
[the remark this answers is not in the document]
dcelasun: Correct, and this was the more useful catch. `setEnvironmentSchema` sets `revision` and `template_id` from `LatestDeploymentLog` without reading its `Status`, and `latestDeploymentLog` is the newest deployment whatever its state (the `output` description already admits it returns `null` while a deployment is in progress). So a refresh reads back the revision that was attempted, not the one that is deployed.

Fixed in 00d1d26. The note no longer promises a re-diff; it now says a refresh does not correct this, that a queued, running or failed deployment reads back the revision it attempted, that the next plan is quiet, and to check the deployment in env0.

The status blindness in `setEnvironmentSchema` is a real bug rather than a docs gap. Out of scope for this docs-only PR, filing it separately.

env0/resource_environment.go
```diff
@@ -163,6 +163,9 @@ func resourceEnvironment() *schema.Resource {
 	}
 
 	return &schema.Resource{
+		Description: "Note: updating an environment issues several env0 API calls in a fixed order - project move, template update, environment update, TTL, drift detection, tags, and last a deployment (or, with 'prevent_auto_deploy', direct writes of 'configuration', 'sub_environment_configuration' and 'variable_sets'). Each step runs only when its own fields changed, and there is no rollback. If a step fails, the steps before it are already applied and the ones after it are not; the next plan re-diffs the rest and can be applied again.\n\n" +
```
yarivg: This string is also the Terraform Registry search summary, so the warnings become the resource's first impression. Keep one neutral sentence here and move the detail to an 'Update behavior' section in the doc body.

env0/resource_environment.go
```diff
@@ -163,6 +163,9 @@ func resourceEnvironment() *schema.Resource {
...
+			"Note: the provider does not wait for the deployment it triggers. The apply returns once the deployment is queued, so the state holds the new 'revision' and 'configuration' while env0 may still be running the previous ones. A refresh does not correct this: 'revision' and 'template_id' are read from the latest deployment log whatever its status, so a queued, running or failed deployment reads back the revision it attempted and the next plan is quiet. Check the deployment in env0 to confirm it succeeded.",
```
yarivg: 'Check the deployment in env0' reads as a permanent manual step. Open a follow-up for a wait_for_deployment flag (mirroring wait_for_destroy) and reference it here as the mitigation.

yarivg (approved):
Verified the claims against the code and they are accurate: the fixed update order with early returns is in resourceEnvironmentUpdate, and Read pulls revision/template_id from LatestDeploymentLog regardless of deployment status, so the quiet-drift note is real.

Approving with two non-blocking suggestions (inline): soften the Registry-facing placement of the warnings, and turn the "check the deployment in env0" manual step into a wait_for_deployment follow-up (same pattern as the existing wait_for_destroy flag) so the second note eventually becomes unnecessary.

env0/resource_environment.go
yarivg: Softer framing, same facts: lead with how updates work, keep the recovery story, drop 'no rollback' phrasing. Regenerate docs/resources/environment.md after.

```suggestion
		Description: "Updating an environment applies changes as a sequence of independent steps - project move, template update, environment update, TTL, drift detection, tags, and last a deployment (or, with 'prevent_auto_deploy', direct writes of 'configuration', 'sub_environment_configuration' and 'variable_sets'). Each step runs only when its own fields changed. If a step fails, earlier steps stay applied and the next apply completes the remaining ones.\n\n" +
```

env0/resource_environment.go
yarivg: Same idea here: frame as async behavior instead of a refresh defect. Regenerate docs/resources/environment.md after.

```suggestion
			"Deployments triggered by an update are asynchronous: the apply returns once the deployment is queued and env0 runs it to completion. Until then the state reflects the requested 'revision' and 'configuration', and a queued or failed deployment is not surfaced as drift on the next plan, so check the deployment in env0 to confirm it succeeded.",
```

Follow-up in progress: an opt-in `wait_for_deployment` flag (same pattern as `wait_for_destroy`) that makes the apply wait for the deployment to reach SUCCESS and fails it otherwise. Once that lands, this note can point at the flag instead of the manual check.

env0/resource_environment.go
dcelasun: Agreed, done in 4a63408. `Description` is now one sentence, and the detail moved to an `## Update behavior` section in a new `templates/resources/environment.md.tmpl`, same shape as the existing `templates/data-sources/vcs_connection.md.tmpl`.

The template reproduces the default layout otherwise: I diffed the regenerated page and the Example Usage, Schema and Import sections are byte-identical, and the `# generated by` marker is kept. Trade-off worth knowing: this page no longer tracks the tfplugindocs default layout, so future default changes will not reach it. The repo already accepts that for `vcs_connection`.

env0/resource_environment.go
dcelasun: Fair on the wording, so I dropped the imperative in 4a63408. The section now ends with a statement of the gap instead of a chore: "There is no equivalent of `wait_for_destroy` for deployments, so a deployment's outcome is not reflected in the apply result."

On opening the follow-up, though: that ticket already exists and was cancelled. APO-751 ("Add `wait_for_deployment` so create and update wait for the env0 deployment") was cancelled on 2026-09-14, and the tech design's Status section records the decision as "`wait_for_deployment` is not being added", along with APO-753 and APO-754. That was the day before this PR opened.

So I did not file a new one, and I would rather not point shipped docs at a cancelled flag as the mitigation - users would read it as something on the way. If you want APO-751 revived I will link it here and reword the section around it. Your call.

yarivg: I will check it, I think we should include such option for users, as an opt-in

yarivg (approved):
Suggested softer wording but LGTM

env0/resource_environment.go
dcelasun: Taken in d245c97, with one word changed.

Heads up that the text moved in 4a63408 (your other comment): it now lives in the `## Update behavior` section of `templates/resources/environment.md.tmpl`, not in the `Description` string, so I applied the equivalent edit there rather than committing the suggestion.

The one change: `the next apply completes the remaining ones` reads as a guarantee, and it collides with the second paragraph. It holds when the failing step is the PUT or the TTL, because a refresh reads those back from env0 and the plan re-diffs them. It does not hold when the failing step is the deployment, which is the case the next paragraph says goes quiet. Used `the next apply can complete the rest` so the two paragraphs agree.

Current text:

> Updating an environment applies changes as a sequence of independent steps: project move, template update, environment update, TTL, drift detection, tags, and last a deployment (or, with `prevent_auto_deploy`, direct writes of `configuration`, `sub_environment_configuration` and `variable_sets`). Each step runs only when its own fields changed. If a step fails, earlier steps stay applied and the next apply can complete the rest.

Docs regenerated.

env0/resource_environment.go
dcelasun: Kept your async framing, which is better than what I had, but changed three things. Applied in d245c97, in `templates/resources/environment.md.tmpl` (the text moved out of `Description` in 4a63408).

1. `env0 runs it to completion` is not true for a failed deployment - it runs to a terminal state, which may be FAILURE. As written it reads as a success guarantee, which is the one reassurance this paragraph exists to withhold.
2. `Until then` implies the state self-corrects once the deployment finishes. It does not: `setEnvironmentSchema` copies `revision` and `template_id` from `latestDeploymentLog` without checking `Status`, so the attempted revision stays in state afterwards either way. The next clause in the same sentence (not surfaced as drift) says the opposite.
3. `check the deployment in env0 to confirm it succeeded` is the imperative you asked me to drop in the thread above, which I removed in 4a63408. Left it out here.

I also kept the mechanism (read back from the latest deployment log whatever its status), since that is why the plan stays quiet - without it the claim looks arbitrary.

Current text:

> Deployments triggered by an update are asynchronous: the apply returns once the deployment is queued, and env0 runs it in the background. The state holds the requested `revision` and `configuration` from the moment the apply returns, whether or not the deployment later succeeds. A failed or still-running deployment is not surfaced as drift, because `revision` and `template_id` are read back from the latest deployment log whatever its status. There is no equivalent of `wait_for_destroy` for deployments, so the apply result does not reflect the deployment's outcome.

Happy to take your wording if you disagree on any of the three - say which and I will change it.

PR #1137, opened 2026-09-16
[APO-752] Fix: Report a destroy that is waiting for approval instead of silently succeeding

yarivg:

Resolves [APO-752](https://linear.app/env-zero/issue/APO-752/report-a-destroy-that-is-waiting-for-approval-instead-of-silently).

Destroying an environment with `approve_plan_automatically = false`: without `wait_for_destroy`, Terraform reported success and dropped the resource from state while the destroy sat in `WAITING_FOR_USER`. With `wait_for_destroy = true`, the poller warned every 10 seconds until the timeout, and the error never said an approval was the blocker.

### Solution

- `wait_for_destroy = true`: the wait keeps pending so an approval in env0 completes the destroy. The approval link is logged once as soon as the deployment starts waiting; on the `delete` timeout, the error names the deployment and links to its approval page.
- Without it: destroy always warns it was triggered but not verified, with a conditional approval hint and link (the approval requirement may be policy-inherited, so it is always included).
- Rerunning a destroy after the timeout error reuses the deployment already waiting for approval instead of queuing a duplicate, and skips the destroy entirely when the environment is already `INACTIVE`.
- New `ui_links.go` maps `api_endpoint` to the web UI host per stage; `ApiEndpoint()` added to the API client.
- Documented that `approve_plan_automatically = false` also gates destroys; regenerated docs.

---

On the dev org, locally built provider, environment requiring approval:

- [x] Destroy with `wait_for_destroy = true` and `timeouts { delete = "3m" }`, no approval: stayed pending, errored after 3m with the working approval link; resource stayed in state.
- [x] Same destroy, approved in env0 mid-wait: completed successfully and cleared the state.
- [x] Destroy without `wait_for_destroy`: warning named the deployment with the approval link, state cleared; deployment parked at `WAITING_FOR_USER` until approved.
- [x] With `TF_LOG=WARN`: the approval link is logged once when the destroy starts waiting, with terse "still waiting" lines after.
- [x] Rerunning the destroy while a destroy deployment was parked: no duplicate deployment was created, the parked one was reused and completed after approval (the old code queued a duplicate that sat `QUEUED` until timeout).

<details>
<summary>CLI output</summary>

Pending wait, approval link logged, approved mid-wait, destroy completes:

```
env0_environment.qa: Still destroying... [id=3dc4d393..., 00m50s elapsed]
[WARN] provider.terraform-provider-env0: deployment is waiting for approval in env0,
  approve it at https://dev.dev.env0.com/p/d56130bb.../environments/3dc4d393.../deployments/9bef9066...?organizationId=63c723d4... to proceed
[WARN] provider.terraform-provider-env0: deployment is still waiting for approval in env0
env0_environment.qa: Destruction complete after 1m42s
Destroy complete! Resources: 1 destroyed.
```

No approval until the `delete` timeout (resource stays in state):

```
env0_environment.qa: Still destroying... [id=9d56846c..., 02m50s elapsed]
Error: destroy deployment 'b22d4884...' is waiting for approval in env0,
  approve it at https://dev.dev.env0.com/p/d56130bb.../environments/9d56846c.../deployments/b22d4884...?organizationId=63c723d4... and run the destroy again
```

Without `wait_for_destroy` (resource removed from state):

```
env0_environment.qa: Destruction complete after 2s
Warning: destroy deployment 'f2190656...' was triggered but not verified
The provider did not wait for the destroy to finish ('wait_for_destroy' is false), so its
result is not verified. ... The destroy will only run once it is approved at
https://dev.dev.env0.com/p/d56130bb.../environments/e68570d8.../deployments/f2190656...?organizationId=63c723d4...
Destroy complete! Resources: 1 destroyed.
```

</details>


dcelasun:
@yarivg — read this next to APO-752. Two things worth settling before merge, plus two smaller code points.

**1. This implements the opposite of the ticket's main AC, and the PR body doesn't say so.** APO-752 asks that `wait_for_destroy = true` error immediately on `WAITING_FOR_USER` ("errors within one poll interval and the resource stays in state"). This waits to the full `delete` timeout instead. I think your call is the better one — failing fast makes `wait_for_destroy` unusable for approval-gated orgs, which breaks the `workspace`-rename replacement flow the flag exists for — but it's a spec change. Call it out and get the ticket updated rather than landing it quietly.

**2. The unconditional warning I'd push back on.** The ticket asks for one deployment-log read after the POST, warning only on `WAITING_FOR_USER`. As written, every `terraform destroy` warns, including the clean ones — 30 environments, 30 warnings, every teardown. "The provider cannot always tell whether an approval gates the destroy" isn't quite right: one read tells you exactly. The real obstacle is the race with `QUEUED`, and a short poll handles that.

**3. The reuse/skip pre-check is gated on the wrong signal** (`env0/resource_environment.go:1345`). `requiresApproval` comes from `GetOkExists("approve_plan_automatically")`, so it is false whenever the attribute is absent from config — including when approval is inherited from a project policy. Those users still get the duplicate queued destroy this PR fixes. Drop the gate: the branch already fetches `Environment(d.Id())`, and `LatestDeploymentLog.Status == WAITING_FOR_USER` is authoritative on its own. Cost is one extra GET per destroy.

**4. `err` is discarded when the last status was `WAITING_FOR_USER`** (`env0/resource_environment.go:1385`). `waitForDeployment` keeps the previous `status` when a poll fails, and returns it on `context.Canceled` too — so a transient API error or a Ctrl-C mid-wait both surface as "waiting for approval… run the destroy again", with the real cause swallowed. Gate on the error as well as the status.

Nothing blocking in `ui_links.go` — inferring the UI host from the API host is a guess, but the fallback covers it.

env0/resource_environment.go
```diff
@@ -1338,13 +1341,51 @@ func resourceEnvironmentDelete(ctx context.Context, d *schema.ResourceData, meta
 		return diag.Errorf("could not delete environment: %v", err)
 	}
 
+	// Delete runs against prior state, so the approval requirement is read from there.
+	//nolint:staticcheck // https://github.com/hashicorp/terraform-plugin-sdk/issues/817
+	approveAutomatically, approveAutomaticallySet := d.GetOkExists("approve_plan_automatically")
+	requiresApproval := approveAutomaticallySet && !approveAutomatically.(bool)
```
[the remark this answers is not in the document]
dcelasun: The stated failure mode isn't accurate: `approvalHint()` is called unconditionally, so the no-wait warning always carries the approval link regardless of `requiresApproval` — no link is lost, and the yes/no branch in the flowchart doesn't exist in the code.

The gate is wrong for a different reason, though: it suppresses the parked-deployment reuse and the INACTIVE skip for policy-inherited approvals. Details in the main thread.

env0/resource_environment_test.go
```diff
@@ -1076,6 +1076,114 @@ func TestUnitEnvironmentResource(t *testing.T) {
...
+					gomock.InOrder(
+						mock.EXPECT().Template(environment.LatestDeploymentLog.BlueprintId).Times(1).Return(template, nil),
+						mock.EXPECT().EnvironmentCreate(environmentCreateWithApproval).Times(1).Return(environmentWithApproval, nil),
+						mock.EXPECT().Environment(environment.Id).Times(1).Return(environmentWithApproval, nil),
+						mock.EXPECT().ConfigurationVariablesByScope(client.ScopeEnvironment, environment.Id).Times(1).Return(client.ConfigurationChanges{}, nil),
+						mock.EXPECT().ConfigurationSetsAssignments("ENVIRONMENT", environment.Id).Times(1).Return(nil, nil),
+						mock.EXPECT().EnvironmentDestroy(environment.Id).Times(1).Return(destroyResponse, nil),
+					)
+				})
+			})
```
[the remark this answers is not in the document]
dcelasun: The coverage exists, just not in the two tests you looked at: `TestUnitResourceEnvironmentDeleteWarning` calls `resourceEnvironmentDelete` directly and asserts severity, summary, detail and the full approval URL.

Agreed that these two acceptance-harness tests' names overpromise — they verify the API call sequence, not the diagnostic. Renaming them is the fix, not adding assertions the harness can't make.

yarivg:
@dcelasun thanks, good catches on 3 and 4, both fixed: the pre-check now always does the Environment GET (dropped the `approve_plan_automatically` gate, so policy-inherited approvals get the reuse/skip too), and the wait now returns a typed timeout error, so "approve it and run the destroy again" only shows on an actual timeout. A failed status read or a Ctrl-C surfaces as-is. Pushing shortly.

env0/resource_environment.go
```diff
@@ -224,7 +228,7 @@ func resourceEnvironment() *schema.Resource {
 			},
 			"approve_plan_automatically": {
 				Type:        schema.TypeBool,
-				Description: "should deployments require manual approvals",
+				Description: "should deployments be approved automatically. When set to false, any deployment that modifies resources (apply, targeted apply, destroy) waits for manual approval in the env0 UI before it runs",
```
liranfarage89, line 231: The description isn't accurate: approval doesn't gate the deployment "before it runs". env0 always runs the plan first and gates the apply — the docs are consistent on this ("the next deployment pauses after the plan step"), and the setting is literally "Approve plan automatically to skip the manual approval step".

```suggestion
				Description: "should deployments be approved automatically. When set to false, any deployment that modifies resources (apply or destroy) waits for manual approval in the env0 UI after the plan, before the apply runs",
```

Dropped "targeted apply" — targeted deploys aren't a separate deployment kind. Also worth knowing: Ansible destroys have no approval flow unless an approval policy exists.

env0/resource_environment.go
```diff
@@ -1327,30 +1331,93 @@ func resourceEnvironmentDelete(ctx context.Context, d *schema.ResourceData, meta
...
 	}
 
-	res, err := apiClient.EnvironmentDestroy(d.Id())
-	if err != nil {
-		if frerr, ok := err.(*http.FailedResponseError); ok && frerr.BadRequest() {
-			tflog.Warn(ctx, "Could not delete environment. Already deleted?", map[string]any{"id": d.Id(), "error": frerr.Error()})
+	var deploymentId string
+
+	// A previous destroy may already be parked waiting for approval. Reuse it instead of queuing a
+	// duplicate destroy behind it, and skip the destroy when the environment is already inactive. The
```
liranfarage89, line 1337: This block already fetches the environment, so why the state proxy rather than `environment.RequiresApproval`? That field is the authoritative request-time signal — `destroy.ts:136-138` does `destroyParams.requiresApproval ??= environmentRequiresApproval`, with no policy term.

Reading it instead closes the gate hole @dcelasun raised *and* lets the no-wait warning fire only when approval is actually expected, instead of on every teardown. One extra GET, one fewer `//nolint:staticcheck`.

env0/resource_environment.go
```diff
@@ -1327,30 +1331,93 @@ func resourceEnvironmentDelete(ctx context.Context, d *schema.ResourceData, meta
...
+	// check cannot be gated on 'approve_plan_automatically' in the state: the approval requirement may
+	// be inherited from a policy.
+	if environment, err := apiClient.Environment(d.Id()); err == nil {
+		if environment.Status == environmentStatusInactive {
+			return nil
+		}
+
+		if environment.LatestDeploymentLog.Type == "destroy" && environment.LatestDeploymentLog.Status == deploymentWaitingForUserStatus {
```
liranfarage89, line 1345: Dropping the resource from state with no log line at all — the sibling "Already deleted?" branch warns. Worth a `tflog.Info` for whoever debugs this later.

FYI this won't fire for a workflow environment with a still-ACTIVE sub-environment: those stay ACTIVE after a successful destroy.

liranfarage89 (commented):
Verified the cross-repo claims against `env0/env0@master` and `env0/env-zero-docs@main`. Build, vet and the full unit suite are green locally too.

**Things done well**

- The host mapping in `ui_links.go` matches the backend's own derivation exactly (`common/lambda/src/commons/get-api-domain.ts` + `get-app-domain.ts`), and there's no EU/regional host to break it. The route `/p/:projectId/environments/:environmentId/deployments/:deploymentLogId` and `?organizationId=` are both real, and that page is where the Approve button lives.
- `LatestDeploymentLog.Type == "destroy"` is right. `destroy` is the only destroy-flavored log type; TTL/scheduled/CLI destroys all funnel through the same writer and the flavor lives in `triggerName`.
- Both new behaviors fix real backend behavior rather than hypotheticals. POSTing destroy on an INACTIVE environment returns 200 and creates a fresh deployment (no 400), so the old "BadRequest means already deleted" assumption never covered it. And a second destroy against a parked one creates a duplicate `QUEUED` row that can never start: `WAITING_FOR_USER` is a blocking status, the queue drainer bails on it, and dedup only collapses consecutive `deploy` rows. Reruns piled up stuck rows unbounded.

**Already raised by @dcelasun, agreed** — the APO-752 spec deviation, the unconditional warning, the wrong gate, and the discarded `err`. The last two both fall out of the `RequiresApproval` comment below. On the two acceptance tests at `resource_environment_test.go:1118` and `:1149`: agreed the harness can't assert diagnostics, so renaming is the fix; `TestUnitResourceEnvironmentDeleteWarning` already covers the diagnostic properly.

**Verdict:** approve once the `approve_plan_automatically` wording is corrected and the gate moves to `environment.RequiresApproval`. Everything else is non-blocking.

env0/resource_environment.go
```diff
@@ -1327,30 +1331,93 @@ func resourceEnvironmentDelete(ctx context.Context, d *schema.ResourceData, meta
...
+			deploymentId = environment.LatestDeploymentLog.Id
+			tflog.Info(ctx, "reusing the destroy deployment that is already waiting for approval", map[string]any{"deploymentId": deploymentId})
+		}
+	}
```
liranfarage89, line 1349: `latestDeploymentLog` is only written at the IN_PROGRESS transition, so a destroy still `QUEUED` behind another run isn't visible here and the rerun still duplicates. Doesn't affect the flow you QA'd (a parked destroy has passed IN_PROGRESS), but it is the residual case — worth a line in the comment above so the next reader doesn't assume it's covered.

env0/resource_environment.go
```diff
@@ -1327,30 +1331,93 @@ func resourceEnvironmentDelete(ctx context.Context, d *schema.ResourceData, meta
...
+			return diag.Errorf("could not delete environment: %v", err)
+		}
+
+		if res == nil {
 			return nil
 		}
 
-		return diag.Errorf("could not delete environment: %v", err)
+		deploymentId = res.Id
+	}
```
liranfarage89, line 1368: `EnvironmentDestroy` never returns `(nil, nil)`, so this is only reachable from the ~23 existing mocks that stub it without a `.Return(...)`. Side effect: almost every destroy test now returns here and never reaches the new warning path.

Fine as a shim, but say so in a comment, or give those mocks a real `destroyResponse`.

env0/resource_environment.go
```diff
@@ -1327,30 +1331,93 @@ func resourceEnvironmentDelete(ctx context.Context, d *schema.ResourceData, meta
...
+
+	approvalHint := func() string {
+		if url := deploymentUrl(apiClient, d.Get("project_id").(string), d.Id(), deploymentId); url != "" {
+			return "at " + url
+		}
+
+		return "in the env0 UI"
```
liranfarage89, line 1375: The closure doesn't buy laziness: `OrganizationId()` is memoized and the provider primes it at configure time, and the no-wait branch below calls it unconditionally anyway. A plain string would do.

env0/resource_environment.go
```diff
@@ -1327,30 +1331,93 @@ func resourceEnvironmentDelete(ctx context.Context, d *schema.ResourceData, meta
...
 		}
+
+		return nil
 	}
 
-	return nil
+	// The provider cannot always tell whether an approval gates the destroy (the requirement may be
+	// inherited from a policy), so the approval hint is phrased conditionally and always included.
+	detail := fmt.Sprintf("The provider did not wait for the destroy to finish ('wait_for_destroy' is false), so its result is not verified. The environment was removed from the Terraform state and is no longer managed by Terraform. If the environment requires approval, the destroy will only run once it is approved %s.", approvalHint())
+
```
liranfarage89, line 1396: "inherited from a policy" isn't quite right — `requiresApprovalDefault` is a creation-time seed (`create-environment.ts:47`) and policy updates don't back-propagate.

What genuinely can't be predicted is decided agent-side after the plan: an OPA approval policy returning `pending`, or `ENV0_REQUIRES_APPROVAL=true` from a custom flow. Worth naming those two, since they're the actual justification for a conditional hint.

env0/resource_environment.go
```diff
@@ -1327,30 +1331,93 @@ func resourceEnvironmentDelete(ctx context.Context, d *schema.ResourceData, meta
...
+// deploymentTimeoutError marks the wait timing out, so callers can tell "the deployment is still waiting"
+// apart from a failed status read or a cancelled context.
+type deploymentTimeoutError struct {
+	deploymentType string
+	status         string
+}
+
+func (e *deploymentTimeoutError) Error() string {
+	return fmt.Sprintf("timeout! last '%s' deployment status was '%s'", e.deploymentType, e.status)
```
liranfarage89, line 1412: nit, non-blocking: `returnOnApproval` has no `true` caller outside tests. Since you're touching the signature anyway, dropping it would leave 5 params instead of 7.

env0/ui_links.go
```diff
@@ -0,0 +1,47 @@
...
+	"net/url"
+	"strings"
+
+	"github.com/env0/terraform-provider-env0/client"
+)
+
+// uiBaseUrl maps an env0 API endpoint to the matching web UI base URL. The prod API serves the UI
+// from app.env0.com; any other env0 endpoint of the form api-<host> serves it from <host> on the
+// same scheme. An unrecognized endpoint (e.g. self hosted) returns an empty string.
+func uiBaseUrl(apiEndpoint string) string {
```
liranfarage89, line 14: nit: `environmentUiUrl` in `tests/harness.go:385` (added in #1118) does the same api→ui derivation with a looser rule (`api.<anything>` → `app.<anything>`). `tests` is `package main` so you can't import this — but two implementations that disagree is a trap, so maybe a pointer comment either way.

env0/ui_links.go
```diff
@@ -0,0 +1,47 @@
...
+	parsed, err := url.Parse(strings.TrimSuffix(apiEndpoint, "/"))
```
liranfarage89, line 15: nit: the `TrimSuffix` is a no-op here — only `parsed.Host` is used, and a trailing slash lands in `Path`.

env0/resource_environment.go
yarivg: Fixed in b83b606, took your wording.

env0/resource_environment.go
yarivg: Fixed in b83b606. The state gate was already dropped in ab2ce49 (the pre-check always runs and keys the reuse/skip on the latest deployment log); now the no-wait warning is keyed on `environment.RequiresApproval` too, so a clean teardown stays quiet.

env0/resource_environment.go
yarivg: Fixed in b83b606, added a `tflog.Info`. Thanks for the workflow-environment pointer.

env0/resource_environment.go
yarivg: Added the QUEUED window to the comment in b83b606.

env0/resource_environment.go
yarivg: Documented the shim in b83b606.

env0/resource_environment.go
yarivg: Kept the closure, but the laziness is now real: with the warning keyed on `RequiresApproval` (your comment above), the clean no-wait path returns before ever building the link, and the wait path only resolves it on the first WAITING_FOR_USER. A plain string would call `ApiEndpoint`/`OrganizationId` on every destroy, which also means stubbing them in every destroy test mock.

env0/resource_environment.go
yarivg: Rewrote in b83b606: the comment now names the agent-side cases (an OPA approval policy returning 'pending', ENV0_REQUIRES_APPROVAL from a custom flow) as what cannot be predicted; those stay silently unverified, as before the warning existed.

env0/resource_environment.go
yarivg: Dropped in b83b606. unparam then flagged `deploymentType` (only ever receives "destroy" now), so that went too. 5 params.

env0/ui_links.go
yarivg: Added the pointer comment in b83b606.

env0/ui_links.go
yarivg: Removed in b83b606.

yarivg:
@liranfarage89 thanks for the deep verification, all addressed in b83b606: description wording fixed, the no-wait warning is now keyed on `environment.RequiresApproval` (clean teardowns stay quiet, agent-side OPA/custom-flow approvals stay silent as before), and the two no-wait acceptance tests were renamed since the harness cannot assert the diagnostics (covered by `TestUnitResourceEnvironmentDeleteWarning`). Also dropped `returnOnApproval` and `deploymentType` from `waitForDeployment`. Replied inline on the rest.

env0/resource_environment.go
```diff
@@ -1327,30 +1331,110 @@ func resourceEnvironmentDelete(ctx context.Context, d *schema.ResourceData, meta
...
+			tflog.Info(ctx, "the environment is already inactive, skipping the destroy", map[string]any{"id": d.Id()})
 
 			return nil
 		}
 
-		return diag.Errorf("could not delete environment: %v", err)
+		requiresApproval = environment.RequiresApproval != nil && *environment.RequiresApproval
+
+		if environment.LatestDeploymentLog.Type == "destroy" && environment.LatestDeploymentLog.Status == deploymentWaitingForUserStatus {
+			deploymentId = environment.LatestDeploymentLog.Id
```
liranfarage89: This is the one place where you have proof the destroy is parked, but `requiresApproval` still comes only from `environment.RequiresApproval` a few lines up. So the exact case the comment at :1405 says stays silent (an OPA policy returning `pending`, `ENV0_REQUIRES_APPROVAL`) goes silent here too, even though we just observed the deployment waiting.

```suggestion
			deploymentId = environment.LatestDeploymentLog.Id
			requiresApproval = true
```

env0/resource_environment.go
```diff
@@ -1327,30 +1331,110 @@ func resourceEnvironmentDelete(ctx context.Context, d *schema.ResourceData, meta
...
+
+	requiresApproval := false
+
+	// A previous destroy may already be parked waiting for approval. Reuse it instead of queuing a
+	// duplicate destroy behind it, and skip the destroy when the environment is already inactive. The
+	// check cannot rely on 'approve_plan_automatically' in the state: 'requiresApproval' on the
+	// environment is the authoritative request-time signal. Note that a destroy still QUEUED behind
+	// another run is not visible in the latest deployment log (it is only written on the transition to
+	// IN_PROGRESS), so a rerun in that window still queues a duplicate.
+	if environment, err := apiClient.Environment(d.Id()); err == nil {
```
liranfarage89: nit: when this errors, `requiresApproval` stays false and an approval-gated teardown warns nothing. Degrading to "no warning" is probably right, but worth half a sentence in the comment above so it reads as deliberate rather than accidental.

env0/resource_environment.go
```diff
@@ -1327,30 +1331,110 @@ func resourceEnvironmentDelete(ctx context.Context, d *schema.ResourceData, meta
...
+	return fmt.Sprintf("timeout! last 'destroy' deployment status was '%s'", e.status)
+}
+
+// waitForDeployment polls a destroy deployment until it reaches a terminal status, the timeout elapses or ctx is
+// cancelled. It returns the last observed deployment status. A deployment in WAITING_FOR_USER keeps the
+// wait pending, so an approval in env0 completes it. approvalHint is resolved lazily on the first
+// WAITING_FOR_USER and logged, so the user gets the approval link as soon as the deployment starts
+// waiting rather than only when the wait times out, and a wait that never hits the approval gate never
+// builds the link.
+func waitForDeployment(ctx context.Context, apiClient client.ApiClientInterface, deploymentId string, timeout time.Duration, approvalHint func() string) (string, error) {
```
liranfarage89: nit, take it or leave it: this is destroy-only now — the error strings hardcode `'destroy'`. `waitForDestroy` would say so in the name.

liranfarage89 (approved):
LGTM — all ten of my comments addressed, left three small ones below, none blocking.

First, a correction on my own review: I verified those line numbers against a stale local checkout and posted them against `ab2ce496`, which you had already pushed. Eight of the ten anchors landed 4-16 lines off the code they were about. You clearly read the substance rather than the anchors, but that was my mistake, not yours — sorry for the noise.

Gating the warning on `environment.RequiresApproval` was the right call and closes @dcelasun's unconditional-warning point in the same move. Good push-back on the lazy `approvalHint` too: with `if !requiresApproval { return nil }` the clean path really does skip building the link now — it didn't before, which is why I raised it.

Also checked: the schema description and `docs/resources/environment.md` match byte-for-byte, and build/vet plus the environment and ui-link suites pass locally. Heads up that CI has only reported CodeQL/Analyze on this head — Unit Tests, Integration Tests and Generate Docs haven't run yet.

env0/resource_environment.go
yarivg: Good catch, fixed in 44ce943: the reuse branch now sets `requiresApproval = true`, a parked destroy is proof enough. Added a unit test for the RequiresApproval-unset case.

env0/resource_environment.go
yarivg: Resolved by the merge with main: since APO-758 the delete starts with the environment GET and a failure aborts the delete (or drops the resource on driftDetected), so there is no silent-degradation path left, `requiresApproval` is always computed from a successful read.

env0/resource_environment.go
yarivg: Also resolved by the merge: APO-751's `awaitDeployment` calls it with "deploy" and `returnOnApproval` again, so it is back to being generic (`deploymentType` param restored).

PR #1138, opened 2026-09-16
[APO-756] Fix: Validate an environment move at plan time

dcelasun:

Resolves [APO-756](https://linear.app/env-zero/issue/APO-756/validate-an-environment-move-at-plan-time-and-say-in-the-docs-what).

`resourceEnvironmentUpdate` starts by calling `EnvironmentMove` on any `project_id` change, but the plan shows `project_id: "a" -> "b"` as an ordinary in-place field change. The move changes which project-level credentials and project-scoped variables apply to the environment, and re-keys its outputs, drift causes and cost records to the target project. The description said only "project id of the environment".

None of the move's preconditions are checked, so the server returns them as a 400 after the user has already approved the plan.

### Solution

`resourceEnvironmentCustomizeDiff` gains `validateMoveIsAllowed`. On an actual `project_id` change it reads the environment once and the template once, and fails the plan when:

- the environment is already in the target project (the server 400s on a no-op move);
- the environment's status is `IN_PROGRESS` or `ABORTING`;
- the template is not assigned to the target project — the error names `env0_template_project_assignment`.

`validateTemplateProjectAssignment` now takes a `GetOk`-only interface, so create and plan share one rule. Locking, policies, permissions, deployment limits, cross-organization targets, sub-environment moves and project-scoped templates stay with the server, whose message is surfaced unchanged.

The `project_id` description now says what a change does, and `docs/resources/environment.md` is regenerated.

Two checks the ticket asked for are not here, because they do not hold against the code:

- **Project-level templates.** `GET /blueprints/{id}` always returns the sentinel `projectId: "ROOT"`, so `Template.ProjectId` is not the project-level marker; the server keys its check off `creationScope`/`creationScopeId` (absent from `client.Template`) and off `latestDeploymentLog.blueprintId` rather than `template_id`. Left to the server, whose message already lists the blocking templates by name.
- **Same-project move from the diff alone.** `CustomizeDiff` runs only when `project_id` differs from refreshed state, so state can never say the environment is already there. The check is implemented against the freshly read environment instead, and fires under `terraform plan -refresh=false`.

### Manual QA

Built from this branch against dev, with two projects, a template assigned only to the origin, and one environment:

| Case | Result |
| --- | --- |
| `project_id` changed to a project the template is not assigned to | plan fails: `template is not assigned to project '<id>': assign it with 'env0_template_project_assignment'`, no `POST /environments/{id}/move` issued |
| template assigned in an earlier apply, then the same change | plans as `~ project_id`, applies; `GET /environments/{id}` confirms the new `projectId` |
| `project_id` changed while a deployment is running | plan fails: `cannot move an environment while a deployment is in progress (status 'DEPLOY_IN_PROGRESS')` |
| environment moved back via the API, then `terraform plan -refresh=false` | plan fails: `the environment is already in project '<id>'` |

All fixtures destroyed afterwards.

Note for reviewers: adding the `env0_template_project_assignment` and changing `project_id` in the **same** apply now fails at plan, because `CustomizeDiff` runs before the assignment is created. The server does not require the assignment for a move at all, so that combination works today and will not after this PR — this is the deliberate cost of moving the provider's own create-time rule to plan time, and the `project_id` description says the assignment must be applied first. Same shape as the `force_destroy` ordering APO-755 documented.


env0/resource_environment.go
```diff
@@ -444,6 +448,48 @@ func validateReplaceIsAllowed(d *schema.ResourceDiff) error {
...
+	}
+
+	templateId := d.Get("template_id").(string)
+
+	template, err := apiClient.Template(templateId)
+	if err != nil {
+		return fmt.Errorf("could not get template '%s': %w", templateId, err)
+	}
+
+	return validateTemplateProjectAssignment(d, &template)
```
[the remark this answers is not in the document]
dcelasun: Rejecting this one: the premise doesn't hold.

The move never fails over template assignment. `services/environment-service/functions/move-environment.ts` gates the move on exactly two things — `assertCanMoveEnvironment` (cross-organization target, target equal to the current project, status `IN_PROGRESS`/`ABORTING`, locked, direct sub-environment move, Environment Outputs policy mismatch, deployment limit) and `assertProjectScopedTemplatesAllowMove`. Neither checks whether a template is assigned to the target project. That was re-verified against the backend in APO-749: "No template-assignment requirement." So there is no apply-time 400 here for the plan to have missed — the assignment rule is the provider's own, enforced at create today and at move after this PR.

`assertProjectScopedTemplatesAllowMove` is the one that reads `latestDeploymentLog.blueprintId`, but it blocks on `blueprint.creationScope === 'PROJECT'` with a `creationScopeId` outside the target project's hierarchy — not on assignment. That check is deliberately left to the server (see the PR description): `GET /blueprints/{id}` always returns the sentinel `projectId: "ROOT"`, and `client.Template` carries neither `creationScope` nor `creationScopeId`, so the provider cannot evaluate it today. Validating the current blueprint's *assignment* would not approximate it.

On validating the new `template_id` rather than the current one: that is the intended end state. After the apply the environment is in the target project running the new template, and that pairing is what the rule is about. The window where the moved environment still points at the old template is closed by the redeploy later in the same update, and the server does not police it either — `deploy-flow.ts` only asserts the blueprint belongs to the organization.

env0/resource_environment.go
```diff
@@ -444,6 +448,48 @@ func validateReplaceIsAllowed(d *schema.ResourceDiff) error {
...
+	}
+
+	targetProjectId, ok := d.GetOk("project_id")
+	// Empty also means the target is not known yet at plan time.
+	if !ok {
+		return nil
+	}
+
+	// A templateless environment owns a single-use template, so no assignment applies to it.
+	if _, ok := d.GetOk("template_id"); !ok {
```
yarivg: Template-less environments do have a generated template, but its ID is stored in `without_template_settings.0.id`, not `template_id`. So this return also skips the status and same-project checks. A template-less environment in `DEPLOY_IN_PROGRESS` will still pass plan and fail at apply. Can we run those checks before this return and skip only the template-assignment check?

yarivg (approved):
LGTM, see comment about vcs environments

env0/resource_environment.go
dcelasun: Good catch, you're right — fixed in 03dff86.

The environment read and the same-project and in-progress checks now run before the `template_id` guard, and only the template-assignment check is skipped for a templateless environment. The comment says where the single-use template id actually lives.

Added a subtest in `TestUnitEnvironmentWithoutTemplateResource`: a templateless environment in `DEPLOY_IN_PROGRESS` whose `project_id` changes now fails at plan. It fails against the old ordering and passes with the fix, and its mock asserts the only `Template()` calls are the ones the read path already makes — no assignment lookup.

PR #638, opened 2023-04-20
Fix: do not fail on 'failed destroyed' environments when removing a project

TomerHeber:
### Issue & Steps to Reproduce / Feature Request
fixes #637 

### Solution

Removed use-case and test.

PR #1141, opened 2026-09-18
[APO-761] Chore: Replace the project's hardcoded destroy wait with a timeouts block

jackcloudquery:

Resolves [APO-761](https://linear.app/env-zero/issue/APO-761/replace-the-projects-hardcoded-destroy-wait-with-a-timeouts-block).

`wait` on `env0_project` polls against a hardcoded 10 minute cap nobody can change, and it ignores `ctx.Done()`: cancelling `terraform destroy` mid-wait keeps the goroutine polling until that cap elapses. It also polls under `force_destroy`, where the assert it is polling returns nil on the first call.

### Solution

Replaced the goroutine-and-channel loop with `waitForProjectEnvironmentsToBeArchived`: a plain loop that asserts first, then selects on the ticker, the timeout timer and `ctx.Done()`, so cancellation stops the poll immediately. Added a `timeouts` block to `env0_project` with `delete` defaulting to 10 minutes - the same shape #1136 gave `env0_environment` - and Delete passes `d.Timeout(schema.TimeoutDelete)` down. A timeout wraps the last `ActiveEnvironmentError`, so the error names the environment that was still blocking instead of only saying the wait expired. The wait is skipped under `force_destroy`, and a successful wait no longer re-asserts, saving one `ProjectEnvironments` call per destroy. The `retry` field of `ActiveEnvironmentError` has been dead since #638 and is gone. Docs regenerated.

Acceptance tests cover the timeout and the happy path. `TestUnitWaitForProjectEnvironmentsToBeArchived` covers the poller directly: a cancelled context, a timeout naming the environment, an error that is not an `ActiveEnvironmentError` returned immediately, and `force_destroy` skipping the wait.

Ran the locally built provider under real `terraform` (1.15.6) against a **local stub of the env0 API**, not a deployed stage - there are no env0 credentials on this machine. The stub always reports one unarchived `ACTIVE` environment, so the wait can never clear. Everything under test here is provider-side: the SDK's timeout plumbing, the context deadline and SIGINT handling.

- [x] `terraform destroy` with `wait = true` and `timeouts { delete = "10s" }` gave up after 10.3s with `timeout! found an active environment stuck-environment - its infrastructure may still exist`, instead of waiting out the old hardcoded 10 minutes.
- [x] SIGINT 4s into a wait with `delete = "5m"` ended the destroy at 4.07s with `context canceled`.
- [x] The same SIGINT against a provider built from `main` was ignored - still printing `Still destroying... 01m00s elapsed` 65s later, on its way to the hardcoded cap.
- [x] `main` rejects the new config outright: `Blocks of type "timeouts" are not expected here`.
- [ ] **Not** verified against a deployed stage. Worth a second pair of eyes on a real project with a stuck environment before merge.

### Screenshots

`terraform destroy` with `wait = true` and `timeouts { delete = "10s" }` - bounded at ~10s, and the error names the environment that was blocking:

terraform destroy bounded by a 10s delete timeout

SIGINT 4s into a wait with `timeouts { delete = "5m" }` - the destroy ends at 4.07s instead of running out the 5 minutes:

SIGINT during the destroy wait ends it immediately


dcelasun:
Nits only — nothing blocking, the change matches APO-761 point for point and mirrors the `env0_environment` poller from #1136 cleanly.

**Naming:** `defaultProjectDestroyTimeout` is camelCase sitting next to `PROJECT_DESTROY_WAIT_INTERVAL`, which stays SCREAMING_SNAKE. Worth renaming the interval in the same pass so the file is internally consistent.

**`min(timeout, 10*time.Second)` under `TF_ACC` silently overrides an explicit `timeouts` block in test config.** Whoever writes the next acceptance test and sets `delete = "30m"` gets 10s with no hint why. It matches `resource_environment.go:1520` and the comment does say so, so I'd leave it — just flagging the footgun for the next person.

**The `ctx.Done()` comment understates itself.** It says the SDK deadline "usually fires before the timer" — it always does: `helper/schema/resource.go:898` wraps Delete in `context.WithTimeout(ctx, d.Timeout(TimeoutDelete))`, so the same deadline is on `ctx` and starts fractionally earlier. That makes the explicit `timer` unreachable outside `TestUnitWaitForProjectEnvironmentsToBeArchived`, which calls the function directly with `context.Background()`. Fine to keep for parity with #1136 — just worth not mistaking it for a real second bound.

dcelasun (approved):
Approved with nits

jackcloudquery:
Thanks - both actionable nits are in 4969e40.

**Naming:** renamed `PROJECT_DESTROY_WAIT_INTERVAL` to `projectDestroyWaitInterval`. Both references were local to this file, so nothing else moved.

**`ctx.Done()` comment:** you are right that it is always, not usually, and I had verified `resource.go:898` while writing this and still hedged the comment. It now says the SDK deadline is the bound that fires under a real apply, and that the timer only bounds direct callers passing a context without a deadline - so nobody reads it as a second real bound. Kept the timer for parity with #1136 and because `TestUnitWaitForProjectEnvironmentsToBeArchived` relies on it.

**`TF_ACC` clamp:** left as-is per your call. Agreed it is a footgun - it did bite me while writing the acceptance test, which is why `Test wait gives up at the delete timeout` uses `delete = "500ms"`: below the clamped 1s poll interval so the wait polls exactly once, rather than racing the timer against the ticker. Worth a follow-up to make the clamp respect an explicit `timeouts` block in both resources, if you think it is worth a ticket.

